PR #5, opened 2026-07-20
Website publish workflows

planger:
## Summary

Adds GitHub Actions workflows to publish the website (`website/`) to GitHub Pages
and deploy per-PR previews, adapting the four-workflow pattern used by the Theia
website repos. The key difference from Theia: the sources live in this repo but
must publish to the separate `enclave-website` / `enclave-website-previews` repos,
so production uses a cross-repo branch push + legacy Pages (the same mechanism
Theia already uses for previews) rather than an in-repo `deploy-pages`.

## What's included

- **`website/build.sh`** — shared assembly: copies the static marketing site to
  `public/`, builds the Docusaurus docs with an injected `DOCS_BASE_URL`, and
  places them at `public/docs/`.
- **`website-publish.yml`** — on push to `main` (`website/**`), builds with
  `DOCS_BASE_URL=/enclave/docs/` and pushes `public/` to `enclave-website:gh-pages`.
- **`website-preview-build.yml`** — on PR (no secrets, untrusted context), builds
  with the per-PR base path and uploads `site` + `pr-metadata` artifacts.
- **`website-preview-deploy.yml`** — `workflow_run` (trusted) pushes the preview to
  `enclave-website-previews:previews` and comments the URL on the PR.
- **`website-preview-remove.yml`** — on PR close, removes the preview.

All third-party actions are SHA-pinned; license headers pass; the build is verified
end-to-end locally (marketing site + docs with the correct `/enclave/docs/` base URL).

## ⚠️ Requires org configuration before it works

Draft because the workflows depend on org config that is not yet in place:

**Otterdog changes (separate PR against `eclipse-enclave/.eclipsefdn`):**
1. `enclave-website`: `gh_pages_build_type: "legacy"` + `gh_pages_source_branch: "gh-pages"` + `gh_pages_source_path: "/"`.
2. `enclave` repo: declare secrets `DEPLOY_TOKEN` and `DEPLOY_PREVIEW_TOKEN`.
3. Optional cleanup of Theia-leftover env/branch policies on `enclave-website`.

**Eclipse Foundation help desk:**
- Two tokens stored in the EF `pass` store: `bots/ecd.enclave/github.com/publish-token`
  (→ `DEPLOY_TOKEN`, push to `enclave-website`) and
  `bots/ecd.enclave/github.com/preview-token`
  (→ `DEPLOY_PREVIEW_TOKEN`, push to `enclave-website-previews`).

Production is served at `eclipse.dev/enclave` via the EF reverse proxy (base path
`/enclave/`); previews at
`eclipse-enclave.github.io/enclave-website-previews/pr-previews/pr-<N>/`.


planger:
Otterdog changes: https://github.com/eclipse-enclave/.eclipsefdn/pull/3

.github/workflows/website-preview-deploy.yml
```diff
@@ -0,0 +1,70 @@
...
+
+name: Deploy website preview
+
+on:
+  workflow_run:
+    workflows: ["Build website preview"]
+    types: [completed]
+
+permissions:
+  pull-requests: write
```
xai, line 18: Not 100% sure, but it might be that we need `actions: read` here for the artifact download across runs?

.github/workflows/website-preview-build.yml
```diff
@@ -0,0 +1,52 @@
...
+        with:
+          node-version: "22.x"
+      - name: Build site
+        env:
+          DOCS_BASE_URL: /enclave-website-previews/pr-previews/pr-${{ github.event.number }}/docs/
+        run: ./website/build.sh
+      - name: Save PR metadata
+        run: |
+          mkdir -p ./pr-metadata
+          echo "${{ github.event.number }}" > ./pr-metadata/pr-number
```
xai, line 42: Not sure what happens if someone makes a PR from a fork and injects something here in this step (by editing the build workflow in the fork's PR, see https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/) that will then be written into `./pr-metadata/pr-number`. This could probably be exploited to overwrite other existing PR previews or attempt to do other harmful things using path traversal.

Instead, I think we could resolve the PR number in the (trusted) deploy workflow by deriving it from Github's `workflow_run` payload.

.github/workflows/website-preview-deploy.yml
tortmayr: Read action should no be needed here.
See https://github.com/eclipse-theia/theia-website/blob/master/.github/workflows/preview-deploy.yml

.github/workflows/website-preview-build.yml
planger: Right, good comment!
Re your suggestion: Not sure if the `workflow_run` payload is available from fork PRs.

How about the "simple" approach to at least minimize the risk by guarding the deploy workflow's "Read PR number" step to `[[ "$PR_NUMBER"  =~ ^[0-9]+$ ]] || exit 1`?

tortmayr: I'm fairly certain that workflows that have been modified in forks needed manual approval anyways before running.
If that is not the case we  have far higher risks then the PR number manipulation.
i.e. an attacker could modify the workflow arbitary, e.g granting itself write access.

We can quickly verify this after merging.

PR #6, opened 2026-07-20
chore: polish docs and remove an unused port helper

sdirix:
Use the camelCase spec.yaml field names (apiKey, securestorageDirEnv, postStart.openIDE) in extension-facing docs. The snake_case forms are internal model names, not what extension authors write.

Move the network apply note from the README footer into the networking guide and fix a stale example URL in the docusaurus config comment.

Remove the unused IsUnspecifiedHostIP helper. The loopback clamp it guarded no longer exists, so the helper and its stale attacker-scenario comment are dead code.


tortmayr (approved):
Thanks! LGTM 👍🏼

PR #7, opened 2026-07-20
feat(ports): accept Docker's "0" sentinel for auto-assigned host ports

sdirix:
-p rejected a host port of "0", so the standard Docker form 127.0.0.1:0:5391 for letting the daemon pick a free host port was reported as an invalid port format. Publishing the same container port across many concurrent sessions required hand-assigning a unique host port to each one.

ParsePortSpec now accepts "0" in the host-port position. A bare "0" and a "0" container port stay invalid. The sentinel reaches Docker unchanged and the assigned port is discoverable via enclave ps --json. ResolvePorts logs auto-assigned bindings as "<auto>" and the invalid format hint mentions the new form. Docs and tests updated.

### DIFF
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -125,7 +125,7 @@ Mutation commands (`add-domain`, `remove-domain`, `set-mode`) apply the new poli
 | `--backend <backend>` | Isolation backend: `docker` (default) or experimental `qemu` |
 | `--name <name>` | Named persistent session |
 | `--background` | Detached background session |
-| `-p <port>` | Publish a container port to the host (container → host, e.g. `-p 3002`) |
+| `-p <port>` | Publish a container port to the host (container → host, e.g. `-p 3002`). A host port of `0` (e.g. `-p 0:3000`) lets the daemon pick a free host port (Docker only); read it back with `enclave ps --json`. |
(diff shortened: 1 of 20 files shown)


xai (approved):
Works great and looks good to me! Two small follow-ups:

- A profile with a declared openUrl could print http://localhost:0; none of the bundled profiles currently hits this.
- The help and docs do not mention that auto-assigned ports are Docker-only. Experimental QEMU rejects host port 0, which is absolutely fine.

Both are not blocking for this PR.

sdirix:
Thanks for your review. I adjusted the commit a bit to properly log the port on startup. The only difficulty is (and the url logging had the same issue) is that the terminal is already in `raw` mode by then. Please have a look whether you like these changes.

Logs look like this now:
```
info: Network restricted to allowlisted domains via DNS gateway
warn: Docker is running without userns-remap or rootless mode; container root maps to host root. See docs/security/host-hardening.md.
info: Building gateway image for claude.
ok: Gateway image built
info: Port forwarding: 127.0.0.1:32798 -> 2000
```

xai:
Looks good, and even without the changes it would have been absolutely sufficient! Thanks @sdirix!

PR #12, opened 2026-07-21
feat: launch Theia in one command and allow Open VSX marketplace access

ndoschek:
#### What it does

Makes IDE tool profiles (`theia`, `theia-next`) usable in one command. These profiles run `sleep infinity` and only launch the IDE on the detached path, so a foreground `enclave --tool theia` hung with no IDE, and the useful flow required remembering `--background`.

- Force the detached path for the `run` action of profiles that declare `post_start.open_ide`, so a bare `enclave --tool theia` starts the container and opens the IDE in one step.
- Leave `shell`/`exec` sessions and explicit `--background` runs untouched.
- Reword the foreground IDE-hook warning and update docs (`cli-reference` + both tool READMEs).

It also fixes Open VSX marketplace access behind the network gateway. The Theia and Theia Next allowlists blocked all DNS by default and never allowlisted Open VSX, so the Extensions view failed with `Error fetching extensions. fetch failed`.

- Add an `openvsx.conf` allowlist fragment for `open-vsx.org` (registry/API, also covering subdomains such as `ai.open-vsx.org`) and `openvsx.eclipsecontent.org` (the Eclipse content CDN that serves `.vsix` downloads), included from both tool allowlists.

#### How to test

1. `make build && make test && make lint`
2. `./bin/enclave --tool theia` returns immediately (detached) and opens the IDE attached to the container. Confirm with `./bin/enclave ps`.
3. `./bin/enclave shell --tool theia` still gives a foreground container shell (no IDE).
4. In the IDE, open the Extensions view and search the marketplace. Extensions list and install instead of failing with `fetch failed`.

#### Follow-ups

- Attaching a second IDE while one is already running opens an empty, unattached window (the `--attach-container` argument is dropped). This is a Theia desktop (Electron single-instance) issue, not an enclave one. I'm already looking into it on the Theia side and will track it separately in the Theia repository.
- Consider whether the Open VSX domains belong in the shared base allowlist rather than only the Theia tools.

#### Breaking changes

#### Review checklist

### DIFF
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -52,9 +52,14 @@ With no command, starts a new agent session (`run` is implicit).
 
 These are host-side attach commands, not tool selections: they launch the
 host-installed IDE against an already-running container of any tool. With
-exactly one running container, the name may be omitted. To start a new
-session container that auto-launches the IDE on top, use the corresponding
-tool profile instead (`enclave --tool theia`); see [Tools](tools.md).
+exactly one running container, the name may be omitted.
+
+To start a new session container that auto-launches the IDE on top, use the
+corresponding tool profile instead (`enclave --tool theia`); see
+[Tools](tools.md). The IDE tool profiles run detached automatically: a bare
+`enclave --tool theia` starts the container in the background and opens the IDE
+in one step, printing the container name so you can reattach later with
+`enclave theia <container>`.
 
 ### Network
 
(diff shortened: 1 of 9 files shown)


xai (approved):
Thanks, works great!

ndoschek:
Thank you for the review Olaf! :tada: 
Can you please go ahead and merge when convenient, since I am not a committer on the project, thanks!

PR #15, opened 2026-07-21
feat(runtime): let enabled features ship agent skills

sdirix:
#### What it does

A feature extension can ship a skills/ directory. When the feature is enabled, its skills are composed into the tool's skills mount between tool-extension skills and shared global/project skills. Features that are not enabled contribute nothing.

Composition happens host-side because the container mounts the skills directory read-only. Both composition paths are covered, the standalone skills mount and the generated config source used with host passthrough and config overrides.

Feature skills overlay as trusted extension content and skip the portable-skill validation applied to shared skills.

#### How to test

I built a feature which ships skills in my fork: https://github.com/sdirix/enclave/tree/feat/diffity-feature

I tested it with Claude Code and it works as expected

#### Breaking changes

#### Review checklist


xai (changes requested):
Only very minor remarks:
- Slight documentation gap in docs/configuration.md, docs/ARCHITECTURE.md, and AGENTS.md, all of which still describe the skill precedence before this patch.
- The fallback skills-generated/ directory is shared by concurrent sessions. Sessions with different feature selections can probably overwrite or interfere with each other’s mounted skills, right?

xai (approved):

PR #18, opened 2026-07-22
Fix rolling release publishing

xai:
## Summary

- Publish rolling releases automatically for non-documentation pushes to `main`, while preserving manual dry runs and publishing.
- Replace tildes in artifact names before generating `checksums.txt` so checksum entries match the published assets.


tortmayr (approved):
LGTM! 👍🏼

PR #20, opened 2026-07-24
feat: let features declare auto-assigned host ports

sdirix:
#### What it does

Feature mixins can now declare top-level `ports` like tools. Entries with `publish: true` are bound only for sessions that enable the feature, through the same resolution seam as a user-supplied -p. User mappings and tool-declared ports win when the container port is already mapped.

A new `hostAllocation` field (allowed values `fixed` which is the current default, and the new `auto`) on port entries (tools and features) publishes as `0:<container>` when `auto` is chosen, letting the daemon assign a free host port so concurrent sessions do not contend.

#### How to test

Add the new attribute to an existing feature and check whether ports are assigned as expected.

It's used in my [custom features in my fork](https://github.com/sdirix/enclave) if you want to try it in practice without setting something up.

#### Breaking changes

#### Review checklist


xai (approved):
Looks good and works well!

I tested it with a minimal extension that opens an HTTP server using Python in its entrypoint and these settings in the extension's spec.yaml:
```yaml
ports:
  - container: 7654
    hostAuto: true
    publish: true
    label: Port Probe
    openUrl: "http://localhost:{host_port}"
```
Auto-assignment and reported URL worked!

sdirix:
Is `hostAuto` a good name or should we name it differently?

sdirix:
And maybe this should not be a feature setting but some enclave setting or commandline parameter? We could also make it automatic, i.e. we try to bind to that port on the host and if it's not free, then we use a free one?

xai:
> Is `hostAuto` a good name or should we name it differently?

I think `hostAuto` is understandable, so I'm happy with that name. Alternatively, we could do something like `hostAllocation` as key and `auto` as value, keeping us the option open to add values like `fixed` or `fallback` or so in the future. (Edit: `hostPort` as key wouldn't be good for the second scenario because it tempts users to put numbers as value then)

> And maybe this should not be a feature setting but some enclave setting or commandline parameter? We could also make it automatic, i.e. we try to bind to that port on the host and if it's not free, then we use a free one?

We still have enclave's `-p`, so that already covers the generic case, but with `0:` it will of course never try to bind the suggested port even if it is free on the host. Not sure what the best UX here would be.

sdirix:
I liked the `hostAllocation` suggestion so this PR is now adapted ;)

xai:
Needs a rebase, looks good otherwise!

xai (approved):

PR #21, opened 2026-07-24
Embed runtime assets in release binary

xai:
## Summary

To start Enclave, users currently need the `enclave` executable and additional runtime assets, such as build context, extensions, docs, and gateway proxy sources.
While the debian package or a `make install` take care of providing these artifacts, the standalone binary produced by our CI workflow does not.
Instead of offering a tarball and leaving the asset installation to the user, we now produce a self-contained `enclave` binary that contains its runtime assets.

This patch does the following:
- Embed the runtime asset tree () in the `enclave` binary. A binary with no asset tree adjacent to its
  executable (downloaded release binary, `make install`, `go install`)
  extracts them atomically into a content-keyed cache directory
  (`~/.cache/enclave/assets/<hash>/`), making it a complete installation.
- Asset discovery drops the previous data-root tier; deb installs and in-tree
  dev builds resolve exactly as before. `make install` now installs only
  the binary and conservatively removes known files staged by older source
  installs
- The rolling release publishes the self-contained `enclave-linux-amd64`
  and smoke-tests extraction (concurrent first runs, exec bits) in CI.

## Test plan

- [ ] `./scripts/smoke-standalone-binary.sh ./enclave` (also runs in CI)
- [ ] `make build && ./bin/enclave` uses live worktree assets and does not touch assets in cache
- [ ]  The debian package continues to use its installed worktree assets is `/usr/share/enclave/` and does not touch assets in cache
- [ ] A copied binary outside the checkout extracts to the cache and runs

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@
 #
 # SPDX-License-Identifier: MIT
 
-.PHONY: build cross-build install install-assets uninstall clean clean-lint-cache test deb lint lint-changed lint-report fmt vet generate lint-tools check-go completions check-license-headers
+.PHONY: build cross-build install install-binary clean-legacy-assets uninstall clean clean-lint-cache test deb lint lint-changed lint-report fmt vet generate lint-tools check-go completions check-license-headers
 
 BIN_DIR ?= ./bin
 BINARY ?= enclave
@@ -19,7 +19,6 @@ LINT_GO_DIRS := cmd internal extensions/tools
 LINT_TMP_DIR := ./.tmp
 LINT_GO_CACHE := $(LINT_TMP_DIR)/go-build
 LINT_GOLANGCI_CACHE := $(LINT_TMP_DIR)/golangci-lint-cache
-GATEWAY_PROXY_SOURCE_PATHS_FILE := $(CURDIR)/internal/gateway/gateway_proxy_build_inputs.txt
 BASE_REF ?=
 
 check-go:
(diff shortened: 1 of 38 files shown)


internal/config/embedded_assets_test.go
```diff
@@ -0,0 +1,268 @@
...
+	"enclave/internal/appassets"
+)
+
+func TestDiscoverAppRootDoesNotUseLegacyDataRoot(t *testing.T) {
+	home := t.TempDir()
+	t.Setenv("HOME", home)
+	t.Setenv("XDG_DATA_HOME", filepath.Join(home, "data"))
+	t.Setenv("XDG_CACHE_HOME", filepath.Join(home, "cache"))
+	t.Setenv("ENCLAVE_HOME", "")
+	legacyRoot := hostDataRoot(home)
```
tortmayr: With tier 3 gone, this test is the only remaining caller of `hostDataRoot` ([def on main](https://github.com/eclipse-enclave/enclave/blob/2d0134f99bcacddc2da88b54022e559f85b57243/internal/config/host_roots.go#L56-L61)). It's effectively dead production code kept alive by the regression test.

internal/config/embedded_assets.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	if !filepath.IsAbs(cacheRoot) {
+		return "", fmt.Errorf("asset cache root is not absolute: %s", cacheRoot)
+	}
+
+	assetsDir := filepath.Join(cacheRoot, embeddedAssetStoreDir)
+	root := filepath.Join(assetsDir, key)
+	if _, err := os.Lstat(root); err == nil {
+		if err := validatePublishedAppRoot(root); err != nil {
+			return "", fmt.Errorf("embedded asset cache %s is invalid: %w; remove cache directory %q and retry", root, err, root)
```
tortmayr: A cache entry that fails validation makes every subsequent run fail with this message until the user manually removes the directory. Given the atomic temp+rename publish this should never happen in practice, but a poisoned entry (or one left by a future format change) hard-blocks the binary. Might be worth extracting to a fresh entry instead of failing.

cmd/enclave/main.go
```diff
@@ -10,6 +10,7 @@ package main
 import (
 	"os"
 
+	_ "enclave"
```
tortmayr, line 13: This blank import bakes the full embedded tree into every `./cmd/enclave` build, so the deb binary now carries the assets in the binary and on disk under `/usr/share/enclave`. Harmless, but it doubles that storage. If the deb size matters, a build tag could drop the embed for the packaged build.

tortmayr (changes requested):
In general code looks good to me. I have a few remarks.
Executed the testplan and everything seems to work as expected.

internal/config/embedded_assets.go
```diff
@@ -0,0 +1,178 @@
...
+)
+
+const embeddedAssetStoreDir = "assets"
+
+func extractRegisteredAppAssets() (string, error) {
+	files, key, err := appassets.Embedded()
+	if err != nil {
+		return "", err
+	}
+	home, err := os.UserHomeDir()
```
sdirix: This resolves home directly instead of going through ResolveHostHome

internal/config/embedded_assets.go
```diff
@@ -0,0 +1,178 @@
...
+		}
+		return root, nil
+	} else if !os.IsNotExist(err) {
+		return "", fmt.Errorf("inspect embedded asset cache %s: %w", root, err)
+	}
+
+	if err := os.MkdirAll(assetsDir, 0o755); err != nil { // #nosec G301 -- extracted build contexts must be traversable.
+		return "", fmt.Errorf("create embedded asset cache %s: %w", assetsDir, err)
+	}
+	tempRoot, err := os.MkdirTemp(assetsDir, ".extract-"+key+"-")
```
sdirix: A process killed mid-extraction leaves `.extract` dir in cache which is not reclaimed

cmd/enclave/main.go
xai: You're right, we shouldn't make the deb and also its included binary larger than needed.

xai:
Thanks for the review, I addressed all three points.

internal/appassets/assets.go
```diff
@@ -0,0 +1,147 @@
...
+// Embedded returns the registered asset filesystem and its deterministic key.
+func Embedded() (fs.FS, string, error) {
+	embedded.Lock()
+	files := embedded.files
+	embedded.Unlock()
+	if files == nil {
+		return nil, "", fmt.Errorf("embedded assets are unavailable")
+	}
+
+	embedded.once.Do(func() {
```
tortmayr, line 53: For standalone binaries this hashes the entire embedded tree on every invocation (~2.7ms here) plus a full stat-walk in `validatePublishedAppRoot` to derive and verify the cache key. Not blocking, but a build-time key would keep it off the hot path.

Makefile
```diff
@@ -38,64 +37,43 @@ completions:
...
 install: build completions
+ifeq ($(filter Linux Darwin,$(UNAME_S)),)
+	$(error make install supports Linux and macOS hosts only; on Windows use WSL2)
+endif
 	$(MAKE) install-assets INSTALL_BINARY_LABEL="$(BINARY)"
 
 INSTALL_BINARY_LABEL ?= $(BINARY)
 
-install-assets:
+install-assets: clean-legacy-assets
```
tortmayr: `install-assets` no longer installs any assets, it just stages the binary/completions and runs legacy cleanup. The name is now misleading

internal/appassets/assets.go
xai: I need to look into this in more detail

tortmayr (approved):
LGTM! 👍🏼

PR #24, opened 2026-07-28
feat: vnc feature serving a contained GUI

sdirix:
#### What it does

Opt-in `vnc` mixin: a virtual X display (Xvnc) running a fullscreened Chromium, served over VNC (RFB). The raw RFB port is published on the host loopback so any VNC client can attach directly. Find the port with `enclave ps --json` and read the per-session password from /tmp/enclave-vnc/vnc-password.

A supervisor keeps Xvnc, matchbox-window-manager, and a headful Chromium at $VNC_URL alive with restart loops. Xvnc listens on all container interfaces so the published port reaches it, and enforces a random password at the RFB layer (VncAuth). That password is what shapes the boundary: holding it is what grants control of the display, and because it is generated per session it reaches exactly one session's display, which is why the agent knowing it is harmless.

Exporting DISPLAY and BROWSER=vnc-open routes "open in browser" flows onto the contained display, where all real browsing stays on the session's gateway-restricted network.

#### How to test

1. Build the local enclave
2. Run enclave with the feature, e.g. `./bin/enclave --features +vnc --rebuild`
3. Check the port by executing `./bin/enclave ps --json`
4. Retrieve the vnc password: `docker exec <container-id> cat /tmp/enclave-vnc/vnc-password`
5. Connect: `xtigervncviewer 127.0.0.1:<port>`
6. Enter the password

You can ask the agent to open tabs or show other GUI tools in a way so that it's visible to you

image

If you want a more convenient way to test you can 
- check out [my fork](https://github.com/sdirix/enclave) which includes a separate feature to transform vnc to a websocket and includes an application to discover and open the respective views within browser tabs, OR
- use https://github.com/eclipse-enclave/enclave/pull/29

I created the latter after the first one and it's conceptually the better approach for Enclave itself I think. The HomeShell could then later have a built-in feature similar to the first web-based approach.

#### Breaking changes

#### Review checklist

### DIFF
--- a/extensions/features/vnc/README.md
+++ b/extensions/features/vnc/README.md
@@ -12,25 +12,21 @@ enclave --features +vnc …
 ```
 
 The feature needs a base image whose archive ships Chromium as a deb. The
-default Debian base does; on Ubuntu bases, which only ship Chromium as a
-snap, the feature's install fails (scoped to this feature, with a clear
-error).
+default Debian base does; on an Ubuntu base, which only ships Chromium as a
+snap, the install fails and the build stops with an error naming this feature.
 
 ## Connecting a VNC client
 
 The RFB port (container `5900`) is published with an OS-assigned host port on
-the loopback interface, so concurrent sessions get distinct ports. Find the
-published port with `enclave ps --json` (look for the `5900` container port),
-then read the per-session password out of the container and point your client
-at it:
+the loopback interface, so concurrent sessions get distinct ports. The session
+prints the resolved `vnc://localhost:<port>` at startup; `enclave ps --json`
+reports it too (look for container port `5900`). Read the per-session password
+out of the session and point your client at it:
 
 ```bash
-docker exec <container> cat /tmp/enclave-vnc/vnc-password
+enclave exec --name <session> -- cat /tmp/enclave-vnc/vnc-password
 vncviewer 127.0.0.1:<published-host-port>
 ```
-
-The password is required (see [Access control](#access-control)). It only ever
-gates that one session's display, so reading it out of the container is safe.
 
 ## What runs in the container
 
@@ -43,19 +39,17 @@ loops, logging to `/tmp/enclave-vnc/log/`:
    (no `-localhost`) so the host's loopback-published port reaches it.
 2. **matchbox-window-manager**: fullscreens every window (kiosk-style).
 3. **Chromium**: headful on the virtual display. It starts on a local
-   waiting page and stays there until a page is opened. When `$VNC_URL` is set
-   (default `about:blank`, meaning no target), a one-shot watcher probes it and
-   forwards it into the running browser once its TCP port accepts connections.
-   Loading the URL directly would instead park the display on a
-   connection-refused error page whenever the target server starts later than
-   the stack. Sessions can also drive the browser on demand via `vnc-open`,
-   which is how a consuming feature opens a URL it only knows at runtime.
-   Both the supervisor and `vnc-open` launch it through the shared
-   `/usr/local/bin/vnc-chromium` wrapper, which disables Chromium's own
-   sandbox (the default Docker seccomp profile blocks the unprivileged user
-   namespaces it needs, so the container remains the isolation boundary) and
-   its background networking (component updater, safe browsing, sync), which
-   would otherwise keep hitting gateway-denied domains.
+   waiting page and stays there until a page is opened. When
+   `$ENCLAVE_VNC_URL` is set, a one-shot watcher probes it and forwards it
+   into the running browser once its TCP port accepts connections. Loading the
+   URL directly would instead park the display on a connection-refused error
+   page whenever the target server starts later than the stack. Sessions can
+   also drive the browser on demand via `vnc-open`, which is how a consuming
+   feature opens a URL it only knows at runtime.
+
+Both the supervisor and `vnc-open` launch Chromium through the shared
+`/usr/local/bin/vnc-chromium` wrapper, so the browser behaves the same however
+it was started; that script's header documents the switches it sets and why.
 
 The feature entrypoint additionally exports `DISPLAY=:99` and
 `BROWSER=/usr/local/bin/vnc-open`, and `install.sh` registers `vnc-open` as
@@ -67,6 +61,11 @@ the supervisor's Chromium profile, waiting briefly for its singleton if the
 stack is still booting, so URLs open in the running browser instead of racing
 it, and logs to `/tmp/enclave-vnc/log/vnc-open.log`.
 
+A set `DISPLAY` is also how many tools decide a GUI is available, so with this
+feature enabled `gpg` pinentry, `SSH_ASKPASS`, and `GIT_ASKPASS` prompts render
+on the contained display rather than in the terminal. Check the VNC client if
+an interactive command appears to hang.
+
 ## Access control
 
 The supervisor generates a random password on first start and enforces it at
@@ -76,16 +75,18 @@ copies:
 - obfuscated auth file: `/tmp/enclave-vnc/rfb-passwd` (Xvnc)
 - plaintext: `/tmp/enclave-vnc/vnc-password` (mode 0600)
 
-That password is what shapes the boundary. Holding it is what grants control
-of the display, and nothing else does. It is generated per session, so it
-grants control of exactly one session's display and no other.
+Holding that password is what grants control of the display, and nothing else
+does. It is generated per session, so it reaches exactly one session's display
+and no other — which is why the (untrusted) agent knowing it is harmless, and
+why reading it out of the session is safe.
 
-The plaintext path is what a VNC client user reads (see
-[Connecting a VNC client](#connecting-a-vnc-client)), and it doubles as the
-**integration contract** for a trusted host-side viewer, which can pick the
-password up with `docker exec <container> cat /tmp/enclave-vnc/vnc-password`.
-Because its reach stops at that one session's own display, the (untrusted)
-agent knowing it is harmless.
+The plaintext path is the **integration contract** for a trusted host-side
+viewer: `/tmp/enclave-vnc/vnc-password` inside the session, alongside the
+container port `5900` binding that `enclave ps --json` reports. The path is the
+contract; how a viewer reads it is up to the backend it drives. `enclave exec`
+always allocates a TTY, so it serves the interactive flow above but not a
+headless one — a non-interactive viewer needs a backend-level read (for Docker,
+`docker exec`) until the CLI grows a non-TTY exec.
 
 ## Configuration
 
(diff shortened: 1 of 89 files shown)


planger:
Love it! Didn't get to try it out yet. Just one thought about the naming... I tend to prefer something like "enclave-vnc" instead of GUI to avoid any collisions with the upcoming UI of enclave (homeshell).

Edit: I'm mostly talking about https://github.com/eclipse-enclave/enclave/pull/29 but already wanted to mention it here as this is going to be discussed first and may be impacted by any naming discussions. :-)

sdirix:
I see, yes, "gui" could be misleading. What about "vnc-viewer"?

planger:
Yes, *vnc-viewer* sounds great to me. Clear and descriptive and no collision with any other UI component.

extensions/features/vnc/vnc-supervisor
```diff
@@ -0,0 +1,186 @@
...
+WAITING_PAGE=/usr/local/share/enclave-vnc/waiting.html
+BROWSER_START="about:blank"
+if [ -f "$WAITING_PAGE" ]; then
+    BROWSER_START="file://$WAITING_PAGE"
+fi
+
+# The container is the sandbox. Chromium's own sandbox needs unprivileged
+# user namespaces, which the default Docker seccomp profile blocks, so it is
+# disabled here. Everything Chromium loads egresses via the session gateway.
+run_forever browser chromium \
```
EclipseSourceAI: This loop restarts Chromium unconditionally, but a `chromium` that finds a live profile singleton just forwards its URL and exits 0. So if anything else owns the singleton (which is exactly what `vnc-open`'s fallback launch does once its 30s wait runs out), this respawns every 2 seconds forever, each time popping a new fullscreened waiting page and appending to the log. Worth checking for a live singleton before restarting instead of relying on the bounded wait in `vnc-open`.

extensions/features/vnc/vnc-open
```diff
@@ -0,0 +1,49 @@
...
+# bounded wait the launch proceeds regardless, because with no live singleton,
+# becoming the instance is the best remaining way to show the URL.
+(
+    (
+        for _ in $(seq 1 100); do
+            [ -e "$PROFILE/SingletonSocket" ] && break
+            sleep 0.3
+        done
+        exec chromium --no-sandbox --user-data-dir="$PROFILE" "$@"
+    ) >/dev/null 2>&1 &
```
EclipseSourceAI: Discarding stdout/stderr makes a failed open undiagnosable, and since `install.sh` registers this as the image-wide http/https handler, every `xdg-open`/`$BROWSER` caller in the session silently loses its URL (after a 30s stall) whenever the stack is down. The supervisor already logs per component to `$STATE_DIR/log/`; sending this there too would cost nothing.

extensions/features/vnc/spec.yaml
```diff
@@ -0,0 +1,40 @@
...
+description: >-
+  Virtual X display with a contained Chromium browser, served over VNC (RFB).
+  Opt-in. The raw RFB port is published so any VNC client can attach.
+defaultEnabled: false
+needsRoot: true
+failOnInstallError: true
+priority: 75
+aptPackages:
+  - tigervnc-standalone-server
+  - chromium
```
EclipseSourceAI: `chromium` is not an installable apt package on the Ubuntu base that `docs/tools.md` documents ([link](https://github.com/eclipse-enclave/enclave/blob/5e808201d3d2284e17c075f1cff4dcb4c214074c/docs/tools.md#L29-L29)), and all selected features' packages go through a single `apt-get install` ([link](https://github.com/eclipse-enclave/enclave/blob/5e808201d3d2284e17c075f1cff4dcb4c214074c/runtime-assets/build-scripts/install-feature-apt-packages.sh#L49-L54)), so one unavailable package fails the whole feature stage, not just vnc. Did you try `--features +vnc` against `--base-image ubuntu:24.04`?

extensions/features/vnc/README.md
```diff
@@ -0,0 +1,116 @@
+# vnc feature
+
+Opt-in feature that gives a session a **contained GUI**: a virtual X display
+(TigerVNC's `Xvnc`) running a fullscreened Chromium, served over **VNC (RFB)**.
+The raw RFB port is published on the host loopback, so you can attach any VNC
```
EclipseSourceAI, line 5: The host binding is loopback, but Xvnc listens on all interfaces inside the netns, and under network isolation that netns belongs to the gateway container on a shared Docker bridge. Other containers (including other sessions' gateways) can therefore hit 5900 directly with VncAuth as the only gate. That belongs in Residual risks. Longer term the gateway already has a loopback-forwarding path for exactly this shape of problem ([link](https://github.com/eclipse-enclave/enclave/blob/5e808201d3d2284e17c075f1cff4dcb4c214074c/gateway-entrypoint.sh#L262-L321)), though today it is only driven by tool Go handlers and would not cover unrestricted mode.

extensions/features/vnc/vnc-supervisor
```diff
@@ -0,0 +1,186 @@
...
+# native VNC client needs (fetch it with `docker exec <container> cat
+# $STATE_DIR/vnc-password`) and is also the integration contract a trusted
+# host-side viewer reads the same way. That password is what shapes the
+# boundary: holding it is what grants control of the display, and because it is
+# generated per session it reaches exactly one session's display.
+set -u
+
+DISPLAY_NUM="${VNC_DISPLAY:-:99}"
+GEOMETRY="${VNC_GEOMETRY:-1600x1000}"
+RFB_PORT="${VNC_RFB_PORT:-5900}"
```
EclipseSourceAI: This knob cannot actually be used: the published container port is hardcoded to 5900 in `spec.yaml`, so a consuming feature that sets `VNC_RFB_PORT` only makes the display unreachable. Either drop it or have the supervisor refuse to start when it does not match 5900.

extensions/features/vnc/install.sh
```diff
@@ -0,0 +1,32 @@
...
+# and the apt-installed chromium.desktop (plain `chromium`, no --no-sandbox)
+# would otherwise win, crash on the blocked user namespaces, and silently
+# drop the URL.
+install -D -m 644 "$dir/enclave-vnc-open.desktop" /usr/share/applications/enclave-vnc-open.desktop
+mkdir -p /etc/xdg
+printf '%s\n' \
+    '[Default Applications]' \
+    'x-scheme-handler/http=enclave-vnc-open.desktop' \
+    'x-scheme-handler/https=enclave-vnc-open.desktop' \
+    > /etc/xdg/mimeapps.list
```
EclipseSourceAI: This truncates an image-wide shared file rather than merging into it, so any other extension that registers a scheme handler gets silently dropped depending on install priority. Appending the two lines when the file already exists would be safer.

extensions/features/vnc/vnc-supervisor
```diff
@@ -0,0 +1,186 @@
+#!/bin/bash
```
EclipseSourceAI: `vnc-supervisor` and `vnc-open` have no `.sh` extension, so `make lint`'s shellcheck pass skips them ([link](https://github.com/eclipse-enclave/enclave/blob/5e808201d3d2284e17c075f1cff4dcb4c214074c/Makefile#L166-L172)). They are clean today (checked at all severities), so adding a path pattern for them, like the one that exists for `build-scripts/bin/*`, keeps them that way.

extensions/features/vnc/enclave-vnc-open.desktop
```diff
@@ -0,0 +1,20 @@
...
+# terms of the MIT License, which is available in the project root.
+#
+# SPDX-License-Identifier: MIT
+#
+# Default x-scheme-handler for http/https inside the session image, so
+# xdg-open (and tools built on it, like the npm "open" package) route URLs
+# into the VNC feature's Chromium instead of the apt-installed
+# chromium.desktop, whose plain `chromium %U` crashes without --no-sandbox
+# and silently drops the URL. Wired up via /etc/xdg/mimeapps.list in
+# install.sh.
```
EclipseSourceAI: The xdg-mime-resolves-before-`$BROWSER` rationale is now spelled out four times (here, `install.sh`, the `vnc-open` header, and the README). Keep it in one place and point at it.

extensions/features/vnc/vnc-open
```diff
@@ -0,0 +1,49 @@
...
+# silently drop the URL before it reaches the running instance. After the
+# bounded wait the launch proceeds regardless, because with no live singleton,
+# becoming the instance is the best remaining way to show the URL.
+(
+    (
+        for _ in $(seq 1 100); do
+            [ -e "$PROFILE/SingletonSocket" ] && break
+            sleep 0.3
+        done
+        exec chromium --no-sandbox --user-data-dir="$PROFILE" "$@"
```
EclipseSourceAI: When this wins the singleton it starts Chromium without the supervisor's flags (`--no-first-run`, `--hide-crash-restore-bubble`, `--password-store=basic`, `--disable-gpu`, `--disable-dev-shm-usage`), so the browser behaves differently depending on which path started it. Sharing one flag list between the two would avoid the drift.

extensions/features/vnc/vnc-supervisor
```diff
@@ -0,0 +1,186 @@
...
+# The container is the sandbox. Chromium's own sandbox needs unprivileged
+# user namespaces, which the default Docker seccomp profile blocks, so it is
+# disabled here. Everything Chromium loads egresses via the session gateway.
+run_forever browser chromium \
+    --no-sandbox \
+    --disable-gpu \
+    --disable-dev-shm-usage \
+    --no-first-run \
+    --no-default-browser-check \
+    --hide-crash-restore-bubble \
```
EclipseSourceAI: Chromium's component updater, safe browsing, and optimization-guide fetches will keep hitting gateway-denied domains and filling the network log. `--disable-background-networking --disable-component-update --disable-sync` would keep the contained browser quiet.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Opt-in `vnc` mixin: Xvnc + matchbox + fullscreened Chromium kept alive by an in-container supervisor, RFB published on an auto-assigned host port and gated by a per-session VncAuth password, plus `DISPLAY`/`BROWSER` and an image-wide http(s) scheme handler so "open in browser" lands on the contained display.

No Go changes, and it is the first extension to use both feature-level `commands.startup` and feature-level `ports`, so it exercises existing machinery rather than adding any. Docs and the golden surface snapshot are updated in the right places; `go test ./...`, `check-license-headers`, `validate-extensions` and shellcheck are all clean, and I verified the hand-rolled VncAuth derivation is byte-identical to a reference vncpasswd implementation.

Worth a human look at three things: the Chromium restart loop in `vnc-supervisor`, which can spin every 2s (and flood the display with waiting-page windows) whenever another process owns the profile singleton; the `chromium` apt package against the documented `ubuntu:24.04` base, where one unavailable package fails the shared feature apt step for every feature; and the reachability of the RFB port, which is broader than "host loopback" because Xvnc binds all interfaces in a netns that is shared with the gateway on a Docker bridge. The rest are smaller notes (unusable `VNC_RFB_PORT` knob, `mimeapps.list` clobber, shellcheck coverage gap, repeated rationale comments).

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Follow-up review after the two "address review comments" commits (b93af3d, 57727b8).

All ten points from my earlier review are addressed:
- The Chromium restart loop now checks for a live singleton before respawning, so it no longer fights `vnc-open`'s fallback launch.
- `vnc-open` now logs to `vnc-open.log` instead of discarding output.
- Chromium install moved out of the shared `aptPackages` step into `install.sh` with its own actionable error, so an Ubuntu base fails only this feature with a clear message instead of taking down the whole apt stage.
- The RFB-reachability risk (Xvnc on all interfaces, shared gateway bridge) is now documented in the feature README's Residual risks and cross-linked from a new section in `docs/security/README.md`.
- The unusable `VNC_RFB_PORT` knob is gone; the port is hardcoded with a comment explaining why.
- `install.sh` now merges into `mimeapps.list` instead of truncating it (verified idempotent).
- `vnc-supervisor`/`vnc-open` moved under `bin/`, and both `Makefile` and `lint-changed.sh` now shellcheck `extensions/*/bin/*`.
- The xdg-mime-before-`$BROWSER` rationale is centralized in `vnc-open`'s header, with everything else pointing back at it.
- `vnc-supervisor` and `vnc-open` now share one Chromium invocation via the new `vnc-chromium` wrapper, removing the flag drift.
- `vnc-chromium` adds `--disable-background-networking --disable-component-update --disable-sync`.

Also new: a `vnc_feature_test.go` covering the startup command shape and the install/spec argv, and a config golden-snapshot update for the apt package list change. I read through both commits in full, ran the new tests, `go build`, and shellcheck on the new `bin/` scripts; all clean. I didn't find anything new worth flagging inline.

- [Chromium restart loop could spin every 2s against vnc-open's fallback launch](https://github.com/eclipse-enclave/enclave/pull/24#discussion_r3703781178)
- [vnc-open discarded stdout/stderr, making a failed open undiagnosable](https://github.com/eclipse-enclave/enclave/pull/24#discussion_r3703781267)
- [chromium apt package fails the whole feature apt step on Ubuntu bases](https://github.com/eclipse-enclave/enclave/pull/24#discussion_r3703781353)
- [README didn't document the RFB port's reachability beyond loopback under network isolation](https://github.com/eclipse-enclave/enclave/pull/24#discussion_r3703781436)
- [VNC_RFB_PORT env knob was unusable since the published port is hardcoded](https://github.com/eclipse-enclave/enclave/pull/24#discussion_r3703781547)
- [install.sh truncated the shared mimeapps.list instead of merging](https://github.com/eclipse-enclave/enclave/pull/24#discussion_r3703781662)
- [vnc-supervisor/vnc-open lacked .sh extensions, skipping shellcheck](https://github.com/eclipse-enclave/enclave/pull/24#discussion_r3703781758)
- [xdg-mime-before-$BROWSER rationale was duplicated four times](https://github.com/eclipse-enclave/enclave/pull/24#discussion_r3703781853)
- [vnc-open started Chromium without the supervisor's flags, causing drift](https://github.com/eclipse-enclave/enclave/pull/24#discussion_r3703781920)
- [Chromium's background networking would fill the network log with gateway-denied requests](https://github.com/eclipse-enclave/enclave/pull/24#discussion_r3703781973)

xai (changes requested):
I can see the AI review requested you to change the variables from `VNC_...` to `ENCLAVE_VNC_...` but I don't think this works.

When I use something like:
```
environment:
  variables:
    ENCLAVE_VNC_GEOMETRY: "800x600"
```
then there will be a not-too-obvious warning: "warn: Ignoring environment.variables key "ENCLAVE_VNC_GEOMETRY": reserved for enclave internals"
and then the variables are skipped in `internal/runtime/runtime.go:956`.

Maybe it would be best to rename them back? There shouldn't be a clash with anything else inside of the image hopefully.

PR #25, opened 2026-07-28
Preserve config store timestamps during overlay

xai:
## Problem

Enclave keeps the tool's config directory (for Claude: `.claude/`) in a persistent store on the host. When a launch generates a fresh tool config, `overlayConfigDir` swaps the store's contents for the generated tree in four steps: copy the paths that must survive (like `projects/`) to a temp dir, delete everything in the store, copy the generated tree in, copy the saved paths back.

Those copies carried content and permissions, but wrote the files with the current time as their modification time. Claude stores its session transcripts under `.claude/projects/` and sorts resume candidates by mtime, so after such a launch every past session looked like it had just been touched and `claude --resume` showed them in arbitrary order.

## Solution

Preserve file and directory modification times in the copy paths of both the Docker and QEMU backends, covered by regression tests over the full overlay flow.

### DIFF
--- a/internal/backend/docker/overlay.go
+++ b/internal/backend/docker/overlay.go
@@ -57,7 +57,8 @@ func overlayConfigDir(targetDir string, sourceDir string, preservePaths []string
 // enumerated separately: checking ancestor matches instead of preserve staging
 // existence avoids treating an unrelated sibling as already handled.
 func stagePreservedPaths(targetDir string, preserveDir string, preservePaths []string) error {
-	return filepath.WalkDir(targetDir, func(p string, _ fs.DirEntry, err error) error {
+	stagedParents := make(map[string]struct{})
+	if err := filepath.WalkDir(targetDir, func(p string, _ fs.DirEntry, err error) error {
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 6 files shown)


internal/backend/docker/overlay.go
```diff
@@ -130,9 +131,9 @@ func copyPath(src string, dst string) error {
 				return err
 			}
 		}
-		return nil
+		return os.Chtimes(dst, info.ModTime(), info.ModTime())
```
sdirix, line 182: Only the copied node's own mtime is preserved, not that of parents `MkdirAll` creates on the way.

internal/backend/docker/overlay.go
```diff
@@ -130,9 +131,9 @@ func copyPath(src string, dst string) error {
...
 	default:
-		return copyRegularFile(src, dst, info.Mode().Perm())
+		return copyRegularFile(src, dst, info.Mode().Perm(), info.ModTime())
```
sdirix, line 184: The doc comment above still says only permissions are carried over.

internal/backend/qemu/storage.go
```diff
@@ -295,9 +295,12 @@ func copyPath(src string, dst string, mode fs.FileMode) error {
...
 			}
 		}
-		return nil
+		return os.Chtimes(dst, info.ModTime(), info.ModTime())
 	default:
-		return copyFile(src, dst, firstNonZeroMode(mode, info.Mode().Perm()))
+		if err := copyFile(src, dst, firstNonZeroMode(mode, info.Mode().Perm())); err != nil {
+			return err
+		}
+		return os.Chtimes(dst, info.ModTime(), info.ModTime())
```
sdirix, line 303: This `copyPath` is also the host to store seed path (https://github.com/eclipse-enclave/enclave/blob/e7551a24ab11e069bad25a9e9f2d79589ca6c22a/internal/backend/qemu/storage.go#L141), so `enclave auth import` now keeps the host file's mtime while docker's seed still stamps copies with now (https://github.com/eclipse-enclave/enclave/blob/e7551a24ab11e069bad25a9e9f2d79589ca6c22a/internal/backend/docker/storage.go#L211-L238). Nothing reads those mtimes today, so leaving it as is works for me. If you want the backends aligned, the same two `Chtimes` calls in `copyTreeNoFollow` do it.

sdirix (approved):
Works for me! Great!

PR #26, opened 2026-07-28
feat: add RPM package distribution

### LINKED ISSUE #10: Publish an RPM next to DEB
E.g. for the Fedora Linux users out there, please consider also publishing an RPM next to a DEB package.

xai:
#### What it does

- Adds an RPM spec and a `make rpm` target
- Adds rpm build tools to the devtools image
- Extracts the asset verification steps from the deb-build workflow to a reusable script.
- Adds an rpm-build workflow that publishes an RPM alongside the existing rolling release artifacts. Note that the test-fedora step uses a container because GitHub still does not provide Red Hat or Fedora runners.

Closes #10 

#### How to test

You can build the RPM on both Debian-based and Redhat/Fedora-based systems as long as the rpm build tools are installed (e.g., packages `rpm` in Debian, `rpm-build` in Fedora).

To test the package on Fedora:
0. Install dependencies and provide docker access to the user:
```bash
sudo dnf install docker go runc
sudo systemctl enable --now docker
sudo gpasswd -a "$USER" docker
```
Relogin or run `newgrp docker`.

1. Install the package:
```bash
dnf install ./dist/enclave-*.x86_64.rpm
```
2. Verify enclave is in `/usr/bin/enclave` and its asset tree is in `/usr/share/enclave/` and works as expected
3. Remove package again:
```bash
dnf remove enclave
```
4. Verify the executable and the asset tree is no longer installed.

#### Breaking changes

#### Review checklist


planger (approved):
Great, thank you very much! RPM build works on Ubuntu and I briefly inspected the RPM, but didn't test installing it on an actual RPM-based distro.

  Non-blocking, but I think it may deserve a consideration to build the RPM not using Debian's  `rpmbuild` on `ubuntu-24.04`, where `redhat-rpm-config` and the whole `brp-*` pipeline don't exist, but rather in a `fedora:44` container. That would give us real coverage for  `__brp_mangle_shebangs_exclude_from`, which no CI job executes today.

  The job itself is fairly small. `dnf install -y git golang make rpm-build cpio diffutils` has  to come before `actions/checkout`, otherwise the checkout falls back to the REST tarball and  `rolling-package-version.sh` has no git history to read. After that it is the same `make rpm`  and verify steps, using Fedora's `golang` rather than `setup-go`. I would also move  `upload-artifact` to that job so the `merge-multiple` download in `rolling.yml` does not  collect two RPMs, point `test-fedora` at it, and keep the Ubuntu job as evidence that  `make rpm` still works on a Debian host. The file name changes because `%{?dist}` resolves  there, so we get `enclave-<version>-1.fc44.x86_64.rpm`, which still matches the globs in the workflows and in the release notes.

PR #28, opened 2026-07-29
feat: publish macOS and Linux arm64 rolling-release binaries

planger:
#### What it does

Extends the rolling release from one Linux x86-64 binary to all four platforms Enclave supports natively, and documents the platform story: Windows through WSL2, and the x86-64-only QEMU backend.

The build matrix in `.github/workflows/rolling.yml` now produces:

| Artifact | Built on | Smoke-tested |
|----------|----------|--------------|
| `enclave-linux-amd64` | `ubuntu-24.04` | yes |
| `enclave-linux-arm64` | `ubuntu-24.04-arm` | yes |
| `enclave-darwin-arm64` | `macos-15` | yes |
| `enclave-darwin-amd64` | `macos-15` (cross-compiled) | no, emits a `::notice::` |

Everything builds with `CGO_ENABLED=0` and explicit `GOOS`/`GOARCH`, replacing the per-matrix-entry `cgo` toggle. The runner labels are pinned instead of `-latest`, so the host architecture an artifact is built and tested on cannot change under a moving alias.

Supporting changes:

- `scripts/smoke-standalone-binary.sh` runs on macOS. The asset cache path is derived per platform (the Apple layout under `~/Library/Caches/org.eclipse.enclave` instead of `XDG_CACHE_HOME`), and `mapfile` is gone, since Bash 3.2 on the macOS runners lacks it. The legacy-root decoy stays at the former XDG data root on both platforms, because that is the lookup it guards.
- `.github/workflows/ci.yml` gains a `smoke-macos` job, so the Darwin path in that script is covered on every pull request rather than first running after merge.
- `make cross-build` covers every published non-native target (`darwin/arm64`, `darwin/amd64`, `linux/arm64`) instead of only `darwin/arm64`. `windows/amd64` stays in as an unpublished portability guard.
- Release notes, `README.md`, and `website/docs/docs/getting-started.md` list the per-platform artifacts with install commands that are runnable as written on each platform. The macOS binaries are unsigned and not notarized, so the quarantine attribute is cleared before `install` can propagate it, and the command tolerates its absence after a `curl` or `gh release download`. Both also get a WSL2 section for Windows, including keeping the project inside the WSL filesystem rather than under `/mnt/c`.
- The website install guide replaced a TODO placeholder, so it now also documents the Fedora `.rpm` that `main` already publishes.
- Documents that the experimental `qemu` backend is x86-64 only: it requires `qemu-system-x86_64`, always builds an x86-64 Alpine guest, and gets KVM acceleration only on x86-64 Linux hosts. On arm64 hosts and on macOS it falls back to TCG emulation. Touches `README.md`, `docs/ARCHITECTURE.md`, `docs/cli-reference.md`, and the website page. No behavior change; this was already true but unstated.

No Go source changes.

#### How to test

CI covers most of it: the four `rolling` build jobs, plus the new `smoke-macos` job on pull requests.

Locally:

```bash
make cross-build
./scripts/smoke-standalone-binary.sh ./bin/enclave
```

Verifying the published artifacts needs a merge to `main`, since the rolling release only builds from there. Download each binary, check it against `checksums.txt`, and run `enclave` in a git repository on the matching platform.

#### Follow-ups

- `darwin/amd64` is cross-compiled on the arm64 runner and never executed. GitHub's Intel macOS labels are billed large runners, so this is accepted risk; the job logs a notice so the gap stays visible.
- macOS binaries are unsigned and not notarized. Signing needs an Apple Developer ID and org-level secrets, worth a separate issue if browser downloads should work without the `xattr` step.
- No arm64 `.deb` or `.rpm`. The release table marks both packages x86-64.
- No arm64 QEMU guest bundle.
- No native Windows build. WSL2 stays the documented path; #30 adds a launcher for it, not a port.

#### Breaking changes

#### Review checklist


xai (approved):
Looks good and all cross-compiled binaries I tested work fine!

I tested Windows (amd64) via WSL2 and Linux (arm64).
@kalyi tested macOS (arm64).

Currently, the `make cross-build` target does not produce actual binaries, it rather serves only as a portability check. To get actual cross-compiled binaries, users still have to run something like `CGO_ENABLED=0 GOOS=darwin GOARCH=arm64 go build -o enclave-darwin-arm64 ./cmd/enclave`. Fine by me, especially with rolling releases that include all binaries.

PR #29, opened 2026-07-29
feat: offer 'enclave vnc-viewer'

sdirix:
#### What it does

Resolves the current project's VNC-enabled session, its published RFB
binding, and its per-session password, then launches a host viewer. The
default is `xtigervncviewer` on Linux and macOS Screen Sharing via open,
overridable with the `vnc_viewer` config key.

The password reaches the viewer through the environment (`VNC_PASSWORD`,
`ENCLAVE_VNC_PASSWORD`) rather than argv, which /proc would expose to
every local user for the viewer's lifetime. TigerVNC reads `VNC_PASSWORD`
itself, so the default viewer needs no placeholder and nothing is
written to disk.

`gui_viewer` is a registry option without a CLI flag. It configures a host
command, so a one-shot override on the command line would just be the
command the user could run directly.

Based on https://github.com/eclipse-enclave/enclave/pull/24

#### How to test

1. Install `xtigervncviewer` if you don't have it already. `sudo apt install tigervnc-viewer`
2. Run an enclave with the vnc feature `./bin/enclave --features +vnc`
3. In the same directory run `./bin/enclave vnc-viewer`

#### Breaking changes

#### Review checklist


sdirix:
I will rework this command to be named `enclave vnc-viewer` instead of `enclave gui` to avoid any misunderstandings with HomeShell and making it more explicit.

sdirix:
@planger @xai Should we move this vnc feature and the vnc-viewer feature to the extension repository?

planger:
Yes, I'd be in favor of that in general: make sure enclave(-core) provides all the generic extensibility needed and let extensions mature outside and decide later. The maturity bar can be much lower in the enclave-extensions.

Of course the vnc viewer is a bit of a special case because it is a great generic feature, but still it might be easier to just move it to the extensions and have it merged asap.

xai:
It is a great feature that unlocks a ton of use cases for enclave users, so it is a really hard decision. But in the end, I agree with Philip, as the extension path will make it available and visible faster to users.

PR #32, opened 2026-07-30
fix: correct Pi metadata and provider support

xai:
#### What it does

Fixes the Pi tool extension so a Pi session comes up with correct metadata, the right provider set, and a working model catalog:

- Corrects the extension metadata and its provider IDs and auth storage names.
- Makes Anthropic OAuth opt-in rather than always released into the session.
- Restricts Mistral credential release to the cases that need it.
- Adds `pi.dev` to the gateway allowlist so Pi's model catalog update works under network isolation.

It also fixes an unrelated image-build defect found while testing this branch: `enclave --tool pi shell` died right after startup with

```
/usr/local/bin/entrypoint.sh: line 289: /usr/local/share/enclave/kit-init.sh: Permission denied
```

`Dockerfile` and `runtime-assets/microvm/alpine/build-bundle.sh` both copy `runtime-assets/kit-init.sh` into the image but never normalized its mode, unlike every other asset they copy. The entrypoint sources it as the unprivileged agent user under `set -e`, so a source checkout with a restrictive umask (files at `0660`) produced a root-owned unreadable file and killed the container before any feature entrypoint ran. Installed builds were unaffected: they build from embedded assets, which `appassets.FileMode` writes as `0644`.

#### How to test

Pi extension:

1. `make build && make test`
2. `bin/enclave --tool pi` — the session starts, and Pi's model catalog update succeeds with network isolation active.

Image permissions:

1. From a source checkout, ensure the asset is group-only readable: `chmod 0660 runtime-assets/kit-init.sh`.
2. `bin/enclave --tool pi shell` — before this change the container exits with the `Permission denied` error above; after it, the shell comes up.
3. `docker run --rm --entrypoint /bin/sh <image> -c 'ls -l /usr/local/share/enclave/kit-init.sh'` reports a world-readable mode.

#### Follow-ups

Two hardening items around the `kit-init.sh` failure, neither included here:

- `entrypoint.sh` guards this source with `[ -f ]` where the sibling asset loads (`net.sh`, `build-scripts/lib/common.sh`) use `[ -r ]`. With `-r` an unreadable asset would fail open and degrade instead of killing the container.
- `collectAppRootModeIssues` in `internal/app/build_permissions.go` warns about restrictive modes on the other copied runtime assets but omits `runtime-assets/kit-init.sh`, so nothing pointed at the offending file.

The image-permission fix is independent of the Pi changes and can be split into its own PR if you would rather keep this one scoped to Pi.

#### Breaking changes

#### Review checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -335,7 +335,8 @@ COPY entrypoint.sh /usr/local/bin/entrypoint.sh
 RUN chmod a+rx /usr/local/bin/entrypoint.sh \
         /usr/local/share/enclave/auth-reconcile.sh \
         /usr/local/share/enclave/net.sh && \
-    chmod a+r /usr/local/share/enclave/tmux-session.conf
+    chmod a+r /usr/local/share/enclave/tmux-session.conf \
+        /usr/local/share/enclave/kit-init.sh
 USER ${USERNAME}
 ENTRYPOINT ["/usr/local/bin/entrypoint.sh"]
 CMD ["/bin/bash"]
--- a/extensions/tools/pi/README.md
+++ b/extensions/tools/pi/README.md
@@ -1,7 +1,7 @@
 # Pi
 
 Pi is a coding agent by Mario Zechner (`@earendil-works/pi-coding-agent`). It
-supports multiple LLM providers (OpenAI, Anthropic, Google).
+supports multiple LLM providers (OpenAI, Anthropic, Google, Mistral).
 
 ## Configuration
 
@@ -17,25 +17,29 @@ supports multiple LLM providers (OpenAI, Anthropic, Google).
 | `OPENAI_API_KEY` | OpenAI API access |
 | `ANTHROPIC_API_KEY` | Anthropic API access |
 | `GEMINI_API_KEY` | Google Gemini API access |
+| `MISTRAL_API_KEY` | Mistral API access |
 
 ## Auth Files
 
 - `agent/auth.json`
 
-Pi uses Codex OAuth for OpenAI subscriptions. enclave auto-maps port 1455
-when the `openai-codex` entry is missing from `agent/auth.json`; add `-p 1455`
-(or `-p 1455:1455`) to re-login.
+Enclave recognizes Pi's `openai`, `openai-codex`, `anthropic`, `google`, and
+`mistral` credentials in this file. For browser-based login, enclave auto-maps
+port 1455 for Codex OAuth and port 53692 for Anthropic OAuth when the corresponding entry
+is missing. Add an explicit mapping for the relevant port to re-login.
 
 ## Network Access
 
-Allowlisted domains include OpenAI, Anthropic, Google, GitHub, and common
-package registries (npm, PyPI, Go, CDNs, TLS/OCSP).
+Allowlisted domains include Pi's model catalog, OpenAI, Anthropic, Google,
+Mistral, GitHub, and common package registries (npm, PyPI, Go, CDNs, TLS/OCSP).
 
 ## Settings
 
 Pi settings are seeded from `templates/settings.json` on first run. The upstream
 Pi settings schema supports `defaultProjectTrust`, but the default template stays
 empty so host settings passthrough behaves the same inside and outside Enclave.
+Enclave disables Pi's install telemetry and version check while leaving model
+catalog refresh enabled.
 
 Enclave runs Pi with `--approve` by default, so project-local Pi resources
 are trusted inside the sandbox without showing Pi's project trust prompt. This
--- a/extensions/tools/pi/README.md
+++ b/extensions/tools/pi/README.md
@@ -31,8 +31,9 @@ entry already exists.
 
 ## Network Access
 
-Allowlisted domains include Pi's model catalog, OpenAI, Anthropic, Google,
-Mistral, GitHub, and common package registries (npm, PyPI, Go, CDNs, TLS/OCSP).
+Allowlisted domains include `pi.dev` for Pi's model catalog, OpenAI, Anthropic,
+Google, Mistral, GitHub, and common package registries (npm, PyPI, Go, CDNs,
+TLS/OCSP).
 
 ## Settings
 
--- a/extensions/tools/pi/README.md
+++ b/extensions/tools/pi/README.md
@@ -10,6 +10,12 @@ supports multiple LLM providers (OpenAI, Anthropic, Google, Mistral).
 - **Config directory**: `~/.pi`
 - **Settings file**: `~/.pi/agent/settings.json`
 
+## Environment
+
+Enclave sets `PI_TELEMETRY=0` and `PI_SKIP_VERSION_CHECK=1` to disable Pi's
+install/update telemetry and version check while leaving model catalog refresh
+enabled. See Pi's upstream [environment-variable reference](https://github.com/earendil-works/pi/blob/main/packages/coding-agent/docs/environment-variables.md).
+
 ## API Keys
 
 | Variable | Purpose |
@@ -40,8 +46,6 @@ TLS/OCSP).
 Pi settings are seeded from `templates/settings.json` on first run. The upstream
 Pi settings schema supports `defaultProjectTrust`, but the default template stays
 empty so host settings passthrough behaves the same inside and outside Enclave.
-Enclave disables Pi's install telemetry and version check while leaving model
-catalog refresh enabled.
 
 Enclave runs Pi with `--approve` by default, so project-local Pi resources
 are trusted inside the sandbox without showing Pi's project trust prompt. This
--- a/extensions/tools/pi/README.md
+++ b/extensions/tools/pi/README.md
@@ -24,9 +24,10 @@ supports multiple LLM providers (OpenAI, Anthropic, Google, Mistral).
 - `agent/auth.json`
 
 Enclave recognizes Pi's `openai`, `openai-codex`, `anthropic`, `google`, and
-`mistral` credentials in this file. For browser-based login, enclave auto-maps
-port 1455 for Codex OAuth and port 53692 for Anthropic OAuth when the corresponding entry
-is missing. Add an explicit mapping for the relevant port to re-login.
+`mistral` credentials in this file. For browser-based login, Enclave auto-maps
+port 1455 when the `openai-codex` entry is missing. Anthropic OAuth is opt-in;
+add `-p 53692` when logging in. Add `-p 1455` to re-login with Codex when an
+entry already exists.
 
 ## Network Access
 
--- a/extensions/tools/pi/gateway-allowlist.conf
+++ b/extensions/tools/pi/gateway-allowlist.conf
@@ -10,6 +10,9 @@ no-resolv
 no-poll
 address=/#/
 
+# Pi model catalogs
+server=/pi.dev/8.8.8.8
+
 # OpenAI
 conf-file=/etc/dnsmasq.allowlists/fragments/openai.conf
 
@@ -18,6 +21,9 @@ conf-file=/etc/dnsmasq.allowlists/fragments/anthropic.conf
 
 # Google / Gemini
 conf-file=/etc/dnsmasq.allowlists/fragments/google.conf
+
+# Mistral
+conf-file=/etc/dnsmasq.allowlists/fragments/mistral.conf
 
 # Common Infrastructure
 conf-file=/etc/dnsmasq.allowlists/fragments/github.conf
--- a/extensions/tools/pi/gateway-allowlist.conf
+++ b/extensions/tools/pi/gateway-allowlist.conf
@@ -22,7 +22,8 @@ conf-file=/etc/dnsmasq.allowlists/fragments/anthropic.conf
 # Google / Gemini
 conf-file=/etc/dnsmasq.allowlists/fragments/google.conf
 
-# Mistral
+# Pi uses Mistral's general API; the shared fragment also allows Codestral,
+# which uses a separate credential.
 server=/api.mistral.ai/8.8.8.8
 
 # Common Infrastructure
--- a/extensions/tools/pi/gateway-allowlist.conf
+++ b/extensions/tools/pi/gateway-allowlist.conf
@@ -23,7 +23,7 @@ conf-file=/etc/dnsmasq.allowlists/fragments/anthropic.conf
 conf-file=/etc/dnsmasq.allowlists/fragments/google.conf
 
 # Mistral
-conf-file=/etc/dnsmasq.allowlists/fragments/mistral.conf
+server=/api.mistral.ai/8.8.8.8
 
 # Common Infrastructure
 conf-file=/etc/dnsmasq.allowlists/fragments/github.conf
--- a/extensions/tools/pi/spec.yaml
+++ b/extensions/tools/pi/spec.yaml
@@ -51,7 +51,6 @@ network:
     "*.anthropic.com": anthropic-api-key
     generativelanguage.googleapis.com: gemini-api-key
     api.mistral.ai: mistral-api-key
-    codestral.mistral.ai: mistral-api-key
   serviceAuth:
     openai-api-key: { headerName: authorization, valueFormat: "Bearer %s" }
     anthropic-api-key: { headerName: x-api-key }
--- a/extensions/tools/pi/spec.yaml
+++ b/extensions/tools/pi/spec.yaml
@@ -77,7 +77,8 @@ providers:
     authSession:
       checks:
         - { file: agent/auth.json, type: json_pointer_non_null, pointer: /anthropic }
-    oauthPorts: [{ port: "53692" }]
+    oauthPorts:
+      - { port: "53692", autoHintWhenNoSession: false, requireMappingWhenNoCredentials: false }
   - name: google
     credentials: [gemini-api-key]
     authFiles: [agent/auth.json]
--- a/extensions/tools/pi/spec.yaml
+++ b/extensions/tools/pi/spec.yaml
@@ -8,8 +8,8 @@
 schemaVersion: "1"
 kind: sandbox
 name: pi
-displayName: Anthropic Claude Pi
-description: Anthropic Claude Pi
+displayName: Pi Coding Agent
+description: Coding agent CLI with read, bash, edit, write tools and session management
 
 sandbox:
   entrypoint: { run: [pi] }
@@ -31,11 +31,17 @@ sandbox:
     - agent/skills/
     - agent/themes/
 
+environment:
+  variables:
+    PI_SKIP_VERSION_CHECK: "1"
+    PI_TELEMETRY: "0"
+
 credentials:
   sources:
     openai-api-key: { env: [OPENAI_API_KEY] }
     anthropic-api-key: { env: [ANTHROPIC_API_KEY] }
     gemini-api-key: { env: [GEMINI_API_KEY] }
+    mistral-api-key: { env: [MISTRAL_API_KEY] }
 
 network:
   serviceDomains:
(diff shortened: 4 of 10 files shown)


extensions/tools/pi/spec.yaml
```diff
@@ -44,21 +50,43 @@ network:
...
         - { file: agent/auth.json, type: json_pointer_non_null, pointer: /openai-codex }
     oauthPorts: [{ port: "1455" }]
   - name: anthropic
     credentials: [anthropic-api-key]
-  - name: gemini
+    authFiles: [agent/auth.json]
+    authSession:
+      checks:
+        - { file: agent/auth.json, type: json_pointer_non_null, pointer: /anthropic }
+    oauthPorts: [{ port: "53692" }]
```
EclipseSourceAI: Adding a second `oauthPorts` entry means pi auto-binds two host ports on most runs, not one. `OAuthPortHints` maps a port whenever the provider has no session, ignoring credentials ([link](https://github.com/eclipse-enclave/enclave/blob/5e808201d3d2284e17c075f1cff4dcb4c214074c/internal/tools/oauth_ports.go#L20-L49)), and `OAuthPortValidateRun` then bind-tests it, so a user with no Anthropic credential at all now fails hard if host 53692 happens to be in use. If Pi's Anthropic OAuth is an opt-in flow rather than the default, `autoHintWhenNoSession: false` keeps the port declared without binding it every session.

extensions/tools/pi/spec.yaml
```diff
@@ -44,21 +50,43 @@ network:
...
-  - name: openai-codex
+  - name: openai
     credentials: [openai-api-key]
     authFiles: [agent/auth.json]
     authSession:
-      mode: any
+      checks:
+        - { file: agent/auth.json, type: json_pointer_non_null, pointer: /openai }
+  - name: openai-codex
+    authFiles: [agent/auth.json]
```
EclipseSourceAI, line 68: Dropping `credentials: [openai-api-key]` here removes the only way `OPENAI_API_KEY` can satisfy the OAuth port check for this provider. `ProviderHasEnvCredential` is populated purely from `provider.CredentialSecrets` ([link](https://github.com/eclipse-enclave/enclave/blob/5e808201d3d2284e17c075f1cff4dcb4c214074c/internal/runtime/auth_manager.go#L222-L231)), and that flag is what lets `OAuthPortValidateRun` skip the mapping requirement ([link](https://github.com/eclipse-enclave/enclave/blob/5e808201d3d2284e17c075f1cff4dcb4c214074c/internal/tools/oauth_ports.go#L66-L79)). Was that intentional, or should the key stay listed on both OpenAI providers?

extensions/tools/pi/spec.yaml
```diff
@@ -31,11 +31,17 @@ sandbox:
     - agent/skills/
     - agent/themes/
 
+environment:
+  variables:
+    PI_SKIP_VERSION_CHECK: "1"
+    PI_TELEMETRY: "0"
```
EclipseSourceAI, line 37: These two aren't metadata or provider support, and the PR description doesn't mention them. They're consistent with the privacy defaults the tool guide asks for ([link](https://github.com/eclipse-enclave/enclave/blob/5e808201d3d2284e17c075f1cff4dcb4c214074c/docs/extensions/adding-a-tool.md#L121)), so no objection to the change itself, but a pointer to where these variable names come from upstream would help since nothing in this repo can confirm them.

extensions/tools/pi/spec.yaml
```diff
@@ -44,21 +50,43 @@ network:
...
 
 providers:
-  - name: openai-codex
+  - name: openai
     credentials: [openai-api-key]
     authFiles: [agent/auth.json]
     authSession:
-      mode: any
+      checks:
+        - { file: agent/auth.json, type: json_pointer_non_null, pointer: /openai }
```
EclipseSourceAI, line 66: These five pointers are the substance of the PR and a reviewer can't check them against anything in-tree. A wrong key fails silently: `checkProviderSession` just reports "no session", which shows up as the OAuth port being auto-mapped forever rather than an error. Can you link the upstream `auth.json` shape you verified against?

extensions/tools/pi/spec.yaml
```diff
@@ -44,21 +50,43 @@ network:
     api.anthropic.com: anthropic-api-key
     "*.anthropic.com": anthropic-api-key
     generativelanguage.googleapis.com: gemini-api-key
+    api.mistral.ai: mistral-api-key
+    codestral.mistral.ai: mistral-api-key
```
EclipseSourceAI: Codestral has its own API key, separate from `MISTRAL_API_KEY`. The gateway rewrites the configured header with the mapped secret on requests to declared hosts, so if Pi ever sends a Codestral-specific key to this host it gets replaced by the general Mistral key. Dropping this line leaves `codestral.mistral.ai` reachable via the DNS fragment without header injection.

extensions/tools/pi/spec.yaml
```diff
@@ -8,8 +8,8 @@
 schemaVersion: "1"
 kind: sandbox
 name: pi
-displayName: Anthropic Claude Pi
-description: Anthropic Claude Pi
+displayName: Pi Coding Agent
+description: Coding agent CLI with read, bash, edit, write tools and session management
```
EclipseSourceAI: 74 chars, and it enumerates Pi's built-in tools, which will drift whenever upstream adds one. Every other tool spec is a short noun phrase (longest is `Mistral Vibe CLI coding assistant` at 33), and this string is what `enclave extension list` prints in an aligned column ([link](https://github.com/eclipse-enclave/enclave/blob/5e808201d3d2284e17c075f1cff4dcb4c214074c/internal/app/extension_cmd.go#L87-L93)). Something like `Pi coding agent CLI` fits better.

extensions/tools/pi/README.md
```diff
@@ -17,25 +17,29 @@ supports multiple LLM providers (OpenAI, Anthropic, Google).
...
+Allowlisted domains include Pi's model catalog, OpenAI, Anthropic, Google,
+Mistral, GitHub, and common package registries (npm, PyPI, Go, CDNs, TLS/OCSP).
 
 ## Settings
 
 Pi settings are seeded from `templates/settings.json` on first run. The upstream
 Pi settings schema supports `defaultProjectTrust`, but the default template stays
 empty so host settings passthrough behaves the same inside and outside Enclave.
+Enclave disables Pi's install telemetry and version check while leaving model
+catalog refresh enabled.
```
EclipseSourceAI: This lands in the Settings section, which is specifically about `templates/settings.json` (the paragraph right above explains why that template stays empty). The telemetry and version-check controls are `spec.yaml` env vars, so they belong in their own paragraph or under Configuration.

extensions/tools/pi/README.md
```diff
@@ -17,25 +17,29 @@ supports multiple LLM providers (OpenAI, Anthropic, Google).
...
+Enclave recognizes Pi's `openai`, `openai-codex`, `anthropic`, `google`, and
+`mistral` credentials in this file. For browser-based login, enclave auto-maps
+port 1455 for Codex OAuth and port 53692 for Anthropic OAuth when the corresponding entry
+is missing. Add an explicit mapping for the relevant port to re-login.
 
 ## Network Access
 
-Allowlisted domains include OpenAI, Anthropic, Google, GitHub, and common
-package registries (npm, PyPI, Go, CDNs, TLS/OCSP).
+Allowlisted domains include Pi's model catalog, OpenAI, Anthropic, Google,
```
EclipseSourceAI: Name `pi.dev`. The opencode README spells out its equivalents rather than describing them ([link](https://github.com/eclipse-enclave/enclave/blob/5e808201d3d2284e17c075f1cff4dcb4c214074c/extensions/tools/opencode/README.md#L44-L46)), which is more useful for anyone writing a gateway allowlist override.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Cleans up the Pi extension: fixes the copy-pasted Claude metadata, splits `openai` (API key) from `openai-codex` (OAuth), adds `google`/`mistral` providers with explicit `auth.json` session pointers, allowlists `pi.dev`, and sets two env vars for telemetry/version-check suppression. Config and docs only, no Go changes. `make build`, `make test`, `make lint` and `make generate` are all clean and both golden files are correctly regenerated.

Focus on the provider table in `extensions/tools/pi/spec.yaml`. Two entries there change runtime behavior in ways the diff doesn't show: the new `oauthPorts: [53692]` on `anthropic` means enclave auto-binds a second host port on most sessions and fails when that port is taken, and dropping `credentials: [openai-api-key]` from `openai-codex` removes the `HasEnvCredential` escape hatch in `OAuthPortValidateRun`. The five `auth.json` pointer names are the substance of the fix and can't be verified from this repo, so they need an upstream reference. Remaining comments are doc placement and naming nits.

runtime-assets/microvm/alpine/build-bundle.sh
```diff
@@ -114,6 +114,7 @@ cp /src/runtime-assets/auth-reconcile.sh "$root/usr/local/share/enclave/auth-rec
 cp /src/runtime-assets/net.sh "$root/usr/local/share/enclave/net.sh"
 cp /src/runtime-assets/kit-init.sh "$root/usr/local/share/enclave/kit-init.sh"
 chmod 0755 "$root/usr/local/bin/entrypoint.sh" "$root/usr/local/share/enclave/auth-reconcile.sh" "$root/usr/local/share/enclave/net.sh"
+chmod 0644 "$root/usr/local/share/enclave/kit-init.sh"
```
EclipseSourceAI, line 117: This and the matching `Dockerfile` chmod have nothing to do with Pi metadata or provider support, and the PR description doesn't mention them. The diagnosis looks right (the entrypoint sources the file, so 0644 is correct), but it's a self-contained commit that would land faster as its own PR.

Dockerfile
```diff
@@ -334,7 +334,8 @@ COPY entrypoint.sh /usr/local/bin/entrypoint.sh
 RUN chmod a+rx /usr/local/bin/entrypoint.sh \
         /usr/local/share/enclave/auth-reconcile.sh \
         /usr/local/share/enclave/net.sh && \
-    chmod a+r /usr/local/share/enclave/tmux-session.conf
+    chmod a+r /usr/local/share/enclave/tmux-session.conf \
+        /usr/local/share/enclave/kit-init.sh
 USER ${USERNAME}
```
EclipseSourceAI, line 340: `collectAppRootModeIssues` checks `auth-reconcile.sh`, `net.sh` and `tmux-session.conf` but not `kit-init.sh` ([link](https://github.com/eclipse-enclave/enclave/blob/5e808201d3d2284e17c075f1cff4dcb4c214074c/internal/app/build_permissions.go#L51-L57)). The comment at the top of this file asks for these rules to stay in sync, and the restrictive-mode case this commit fixes is exactly what that warning exists to catch, so it should be listed there too.

extensions/tools/pi/gateway-allowlist.conf
```diff
@@ -19,6 +22,9 @@ conf-file=/etc/dnsmasq.allowlists/fragments/anthropic.conf
 # Google / Gemini
 conf-file=/etc/dnsmasq.allowlists/fragments/google.conf
 
+# Mistral
+conf-file=/etc/dnsmasq.allowlists/fragments/mistral.conf
```
EclipseSourceAI: This fragment also resolves `codestral.mistral.ai` ([link](https://github.com/eclipse-enclave/enclave/blob/5e808201d3d2284e17c075f1cff4dcb4c214074c/extensions/tools/mistral-vibe/README.md#L25-L25)). With the host no longer in `serviceDomains` and no `environment.proxyManaged` declared, `MISTRAL_API_KEY` reaches the container as a placeholder ([link](https://github.com/eclipse-enclave/enclave/blob/5e808201d3d2284e17c075f1cff4dcb4c214074c/internal/runtime/auth_manager.go#L235-L243)), so a Codestral model in Pi sends the placeholder and fails. Either scope the release per host or say in the README that Codestral isn't usable.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Follow-up round. Since the last review the author made Anthropic OAuth opt-in (`autoHintWhenNoSession: false` plus `requireMappingWhenNoCredentials: false`, both needed so `OAuthPortValidateRun` doesn't hard-fail), dropped the `codestral.mistral.ai` credential release, named `pi.dev` in the README, and added an unrelated `kit-init.sh` permission fix to the Dockerfile and the microVM bundle script. `make build`, `make test`, `make lint` and `make generate` are all clean; both golden files match.

Two areas for a human to focus on. First, the `kit-init.sh` commit: the diagnosis is right (the entrypoint sources the file, so 0644 fits), but it isn't Pi-related and it leaves `collectAppRootModeIssues` out of sync. Second, the Mistral change now leaves `codestral.mistral.ai` DNS-reachable through the shared fragment while `MISTRAL_API_KEY` arrives in the container as a placeholder.

Four findings from the previous round are still open and unanswered: `openai-codex` losing `credentials: [openai-api-key]` (the one with real runtime impact), the unverifiable `auth.json` pointer names, the long `description`, and the README Settings-section placement. GitHub rejected my attempt to resolve the threads that were fixed, so the Anthropic OAuth, `pi.dev` naming and Codestral threads can be closed manually.

runtime-assets/microvm/alpine/build-bundle.sh
xai: fixed pr description. up to a human reviewer if it should be split out or not.

extensions/tools/pi/gateway-allowlist.conf
xai: i think pi defines the Mistral provider with https://api.mistral.ai.

extensions/tools/pi/gateway-allowlist.conf
```diff
@@ -19,6 +22,9 @@ conf-file=/etc/dnsmasq.allowlists/fragments/anthropic.conf
 # Google / Gemini
 conf-file=/etc/dnsmasq.allowlists/fragments/google.conf
 
+# Mistral
+server=/api.mistral.ai/8.8.8.8
```
EclipseSourceAI, line 27: Good narrowing, but worth saying why inline instead of the fragment: `mistral.conf` also resolves `codestral.mistral.ai` ([link](https://github.com/eclipse-enclave/enclave/blob/5e808201d3d2284e17c075f1cff4dcb4c214074c/runtime-assets/gateway-allowlists/fragments/mistral.conf#L8-L10)), which has its own API key. Without a note here, a later cleanup that normalizes tool allowlists onto fragments (like mistral-vibe does) would quietly re-open it.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Third round. The only new commit since the last review narrows the Mistral allowlist to `api.mistral.ai` by replacing the shared `mistral.conf` include with an inline `server=` line, which makes DNS reach and credential release cover the same single host. That closes the Codestral gap properly. `make build`, `make test`, `make lint`, `make check-license-headers` and `make generate` are all clean, and both golden files match the spec.

I also traced the `kit-init.sh` fix end to end and the diagnosis holds: the microVM `chown -R agent` covers `/opt/enclave` but not `/usr/local/share/enclave`, which is why this one asset broke while its siblings didn't, and `debian/rules` plus the RPM spec both `chmod -R go+rX` over `runtime-assets`, so packaged installs really were unaffected. Nothing else copied in either build path is left unnormalized.

Where to focus: the provider table in `extensions/tools/pi/spec.yaml`. Four findings from earlier rounds are still open and unanswered, and I have not repeated them as comments. The one with real runtime impact is `openai-codex` losing `credentials: [openai-api-key]`: `OPENAI_API_KEY` no longer suppresses the port-1455 bind test, so a session hard-fails when that host port is occupied. The others are the unverifiable `auth.json` pointer and `PI_*` env var names, the 74-char `description` outlier, and the README Settings-section placement. Only one new comment this round, a maintainability nit on the inline Mistral entry.

- [make Anthropic OAuth opt-in instead of binding a port every session](https://github.com/eclipse-enclave/enclave/pull/32#discussion_r3703756713)
- [name `pi.dev` in the README instead of describing it](https://github.com/eclipse-enclave/enclave/pull/32#discussion_r3703757323)
- [Codestral left DNS-reachable without header injection](https://github.com/eclipse-enclave/enclave/pull/32#discussion_r3703757093)
- [Codestral reachable via the shared fragment while `MISTRAL_API_KEY` arrives as a placeholder](https://github.com/eclipse-enclave/enclave/pull/32#discussion_r3704532786)

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Fourth round. The only new commit since my last review adds the inline explanation on the `api.mistral.ai` allowlist entry that I asked for, so there are no new findings and no new inline comments this round. `make build`, `make test`, `make lint`, `make check-license-headers` and `make generate` are all clean, and both golden files match the spec.

I re-verified the provider rename fallout from scratch this round: provider names are pure labels in the runtime (auth-state map keys and error strings), nothing in Go or docs hardcodes `gemini`, and no host state path is keyed by provider name, so `gemini` -> `google` is safe. `--no-api-key` / `--ephemeral` suppression also still covers `openai-api-key` through the new `openai` provider, so dropping `credentials` from `openai-codex` doesn't regress that path.

Where a maintainer should focus: the four earlier findings that are still open and unanswered, which I have not repeated as comments. The one with real runtime impact is `openai-codex` losing `credentials: [openai-api-key]`, since `OPENAI_API_KEY` no longer lets the port-1455 bind test be skipped (both `codex` and `opencode` keep the key on their OAuth-port provider). The rest are unverifiable-from-this-repo upstream names (the five `auth.json` pointers, the two `PI_*` variables, port `53692`), the 74-char `description` outlier, and the README Settings-section placement. Separately, `collectAppRootModeIssues` still omits `runtime-assets/kit-init.sh`, and whether to split the `kit-init.sh` commit into its own PR is a maintainer call the author has already deferred.

- [say why the Mistral allowlist entry is inline instead of using the shared fragment](https://github.com/eclipse-enclave/enclave/pull/32#discussion_r3707936653)

extensions/tools/pi/spec.yaml
xai: The omission is intentional and correct. Pi treats these as distinct providers:

- openai uses OPENAI_API_KEY.
- openai-codex uses ChatGPT Plus/Pro OAuth credentials.

extensions/tools/pi/spec.yaml
xai: The variables are valid and explicitly documented in Pi’s environment-variable reference (https://github.com/earendil-works/pi/blob/main/packages/coding-agent/docs/environment-variables.md):

- PI_SKIP_VERSION_CHECK disables the pi.dev version request.
- PI_TELEMETRY=0 disables install/update telemetry and provider-attribution headers.

this will stay in.

extensions/tools/pi/spec.yaml
xai: The pointers are correct:

- Pi stores credentials under the provider ID at the top level of auth.json.
- Upstream provider IDs are exactly openai, openai-codex, anthropic, google, and mistral.
- The provider documentation (https://github.com/earendil-works/pi/blob/main/packages/coding-agent/docs/providers.md) confirms four API-key entries; the openai-codex provider definition (https://github.com/earendil-works/pi/blob/main/packages/ai/src/providers/openai-codex.ts) confirms the OAuth provider ID.

extensions/tools/pi/README.md
```diff
@@ -10,26 +10,36 @@ supports multiple LLM providers (OpenAI, Anthropic, Google).
...
 
 - `agent/auth.json`
 
-Pi uses Codex OAuth for OpenAI subscriptions. enclave auto-maps port 1455
-when the `openai-codex` entry is missing from `agent/auth.json`; add `-p 1455`
-(or `-p 1455:1455`) to re-login.
+Enclave recognizes Pi's `openai`, `openai-codex`, `anthropic`, `google`, and
+`mistral` credentials in this file. For browser-based login, Enclave auto-maps
+port 1455 when the `openai-codex` entry is missing. Anthropic OAuth is opt-in;
+add `-p 53692` when logging in. Add `-p 1455` to re-login with Codex when an
```
EclipseSourceAI, line 35: Since `autoHintWhenNoSession: false` means `OAuthPortHints` never surfaces port 53692, this README is the only place the `-p 53692` requirement shows up. `docs/auth.md` carries exactly this kind of per-tool note for codex and opencode ([link](https://github.com/eclipse-enclave/enclave/blob/5e808201d3d2284e17c075f1cff4dcb4c214074c/docs/auth.md#L119-L128)), so a one-liner for Pi there would make it findable.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Fifth round. The one new commit addresses three earlier comments: the `description` is short and no longer enumerates Pi's tools, the telemetry/version-check note moved out of the Settings section into a dedicated `## Environment` section with an upstream link, and `runtime-assets/kit-init.sh` is now listed in `collectAppRootModeIssues` with a matching test case. `make build`, `make test`, `make lint`, `make check-license-headers` and `make generate` are all clean, and both golden files match the spec.

The `kit-init.sh` entry uses the non-executable check, which matches how the entrypoint loads it (sourced, not exec'd), and sits in the existing alphabetical list. Using `environment.variables` in the spec rather than exporting from `entrypoint.d/setup.sh` is the documented mechanism for sandbox specs. I also confirmed the empty `mode` on the new `authSession` blocks defaults to `any`, so dropping `mode: any` is behavior-neutral with one check per provider.

Where a maintainer should focus: nothing new in the code itself, only one doc-discoverability comment about the opt-in Anthropic OAuth port. The remaining open items are maintainer calls, not defects: whether to split the `kit-init.sh` commit into its own PR (the author has deferred that decision), and the `[ -f ]` vs `[ -r ]` guard in `entrypoint.sh`, which the PR description already lists as an intentional follow-up.

- [shorten the `description` so it stops enumerating Pi's tools](https://github.com/eclipse-enclave/enclave/pull/32#discussion_r3703757154)
- [move the telemetry/version-check note out of the Settings section](https://github.com/eclipse-enclave/enclave/pull/32#discussion_r3703757229)
- [list `runtime-assets/kit-init.sh` in `collectAppRootModeIssues`](https://github.com/eclipse-enclave/enclave/pull/32#discussion_r3704532619)
- [link the upstream `auth.json` shape the provider pointers were verified against](https://github.com/eclipse-enclave/enclave/pull/32#discussion_r3703757010)

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Sixth round. The only commit since my last review (`04affaf`) adds a "Pi Anthropic OAuth note" to `docs/auth.md`, telling users to pass `-p 53692` when they intend to log in with Anthropic OAuth. That's exactly the doc-discoverability gap I flagged last round (since `autoHintWhenNoSession: false` means the port is never auto-hinted), it matches the existing Codex OAuth note's format, and the port number matches `oauthPorts` in `extensions/tools/pi/spec.yaml`. No new findings.

Everything else is unchanged since round five. The remaining open items are maintainer calls the author already deferred: whether to split the `kit-init.sh` commit into its own PR, and the `[ -f ]` vs `[ -r ]` guard in `entrypoint.sh` (explicitly called out as a follow-up in the PR description).

- [document the `-p 53692` requirement for Pi's opt-in Anthropic OAuth](https://github.com/eclipse-enclave/enclave/pull/32#discussion_r3711555501)

tortmayr (approved):
LGTM! 👍🏼

PR #34, opened 2026-07-31
fix: resolve secret env aliases by layer priority

sgraband:
#### What it does
Aliases were compared for equality before layer precedence was applied, so a stale value in the persisted env store hard-failed the run instead of losing to a rotated value in the layered secrets files. Rotating a token that is declared under two aliases (GH_TOKEN/GITHUB_TOKEN) wedged every project whose persisted store still held the old value.

The highest-priority layer now wins outright. Only aliases resolved from the same layer must agree, since that is the one case the user has to fix; the error names the layer. Cross-layer drift is warned about and heals on the next run, because the winning value is injected into every alias and written back to the persisted store.

#### How to test
Change your GH_TOKEN/GITHUB_TOKEN to a temporary value (if it was set beforehand) and see that the build succeeds, beforehand it would fail with an error.

#### Breaking changes

#### Review checklist

### DIFF
--- a/docs/extensions/README.md
+++ b/docs/extensions/README.md
@@ -393,11 +393,8 @@ Secrets are split across two `spec.yaml` sections:
 - `credentials.sources.<id>` declares the credential itself: `env` (one or
   more env-var aliases for the same credential) and the enclave-native
   `apiKey` bool (`false` for OAuth/session tokens; omitted/`true` means it's
-  an API key).
-- Env aliases name one value. Precedence: host env, layered secrets files,
-  persisted store; the winner is injected into *every* alias, so a stale
-  persisted value heals after one run. Aliases set in the *same* layer with
-  different values are an error; across layers it is a warning.
+  an API key). See [Authentication](../auth.md#env-aliases) for how multiple
+  aliases are resolved against each other.
 - `credentials.sources.<id>.file` optionally sources the secret from a host
   file: `path` (supports `~`) plus a `parser` — empty for the trimmed raw file
   contents, or `json:<dot.path>` (e.g. `json:auth.token`) to extract a scalar
--- a/docs/extensions/README.md
+++ b/docs/extensions/README.md
@@ -394,6 +394,10 @@ Secrets are split across two `spec.yaml` sections:
   more env-var aliases for the same credential) and the enclave-native
   `apiKey` bool (`false` for OAuth/session tokens; omitted/`true` means it's
   an API key).
+- Env aliases name one value. Precedence: host env, layered secrets files,
+  persisted store; the winner is injected into *every* alias, so a stale
+  persisted value heals after one run. Aliases set in the *same* layer with
+  different values are an error; across layers it is a warning.
 - `credentials.sources.<id>.file` optionally sources the secret from a host
   file: `path` (supports `~`) plus a `parser` — empty for the trimmed raw file
   contents, or `json:<dot.path>` (e.g. `json:auth.token`) to extract a scalar
--- a/internal/runtime/active_secrets.go
+++ b/internal/runtime/active_secrets.go
@@ -12,6 +12,7 @@ import (
 	"os"
 	"sort"
 
+	"enclave/internal/logx"
 	"enclave/internal/model"
 	"enclave/internal/secretfile"
 )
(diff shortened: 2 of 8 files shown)


internal/runtime/active_secrets.go
```diff
@@ -121,28 +122,29 @@ func resolveActiveSecretValue(secret activeSecret, hostHome string, layeredSecre
...
 	values := make([]aliasValue, 0, len(secret.EnvVars))
 	for _, envVar := range secret.EnvVars {
 		if value := os.Getenv(envVar); value != "" {
-			values = append(values, aliasValue{envVar: envVar, value: value, source: "env", priority: 1, found: true})
+			values = append(values, aliasValue{envVar: envVar, value: value, source: "env", priority: 1})
 			continue
 		}
 		if value, ok := layeredSecrets[envVar]; ok && value != "" {
-			values = append(values, aliasValue{envVar: envVar, value: value, source: "secrets", priority: 2, found: true})
+			values = append(values, aliasValue{envVar: envVar, value: value, source: "secrets", priority: 2})
```
EclipseSourceAI: `auth.ResolveLayeredSecrets` flattens `global.env`, `global/<tool>.env` and `projects/<hash>/<tool>.env` into one map ([link](https://github.com/eclipse-enclave/enclave/blob/5e808201d3d2284e17c075f1cff4dcb4c214074c/internal/auth/auth.go#L40-L44)), so all three collapse into `priority: 2` here. `GH_TOKEN` left in `global.env` plus a rotated `GITHUB_TOKEN` in the project file is then a same-layer conflict and still hard-fails, which is the same rotation breakage one level down. Worth a follow-up, since the file order already has a documented precedence.

internal/runtime/active_secrets.go
```diff
@@ -152,14 +154,21 @@ func resolveEnvAliasValue(secret activeSecret, layeredSecrets map[string]string,
...
 			chosen = value
 		}
 	}
 
+	for _, value := range values {
+		if value.value == chosen.value {
+			continue
+		}
+		if value.priority == chosen.priority {
+			return "", "", false, fmt.Errorf("secret %q has conflicting values across env aliases (%s vs %s, both set in the %s layer)", secret.ID, chosen.envVar, value.envVar, chosen.source)
```
EclipseSourceAI: The check only enforces agreement inside the *winning* layer, not inside every layer: with three or more aliases, two conflicting values sitting in a non-winning layer only warn. That is fine as a rule (the persisted layer must be allowed to drift so it can heal), but the docs bullet in this PR states the stronger version.

internal/runtime/active_secrets.go
```diff
@@ -152,14 +154,21 @@ func resolveEnvAliasValue(secret activeSecret, layeredSecrets map[string]string,
...
+		}
+		logx.Warnf("Secret %s: using %s from the %s layer; %s in the %s layer holds a different value and is ignored.", secret.ID, chosen.envVar, chosen.source, value.envVar, value.source)
```
EclipseSourceAI: "the secrets layer" and "the persisted layer" are internal source ids, not names users see anywhere. `docs/auth.md` calls them layered secrets files (three of them) and the host-side env store ([link](https://github.com/eclipse-enclave/enclave/blob/5e808201d3d2284e17c075f1cff4dcb4c214074c/docs/auth.md#L130-L140)). Same for the error above, which asks the user to fix a conflict without saying which file to open.

docs/extensions/README.md
```diff
@@ -394,6 +394,10 @@ Secrets are split across two `spec.yaml` sections:
   more env-var aliases for the same credential) and the enclave-native
   `apiKey` bool (`false` for OAuth/session tokens; omitted/`true` means it's
   an API key).
+- Env aliases name one value. Precedence: host env, layered secrets files,
+  persisted store; the winner is injected into *every* alias, so a stale
+  persisted value heals after one run. Aliases set in the *same* layer with
```
EclipseSourceAI: This is stronger than the code: same-layer disagreement is only fatal in the layer that wins, elsewhere it warns. Also this bullet documents behavior in the middle of a list where every other bullet documents a `credentials.sources.<id>` key, and the user-visible half (which error, which warning) fits better under `## Secrets` in `docs/auth.md` where the layer precedence is already spelled out ([link](https://github.com/eclipse-enclave/enclave/blob/5e808201d3d2284e17c075f1cff4dcb4c214074c/docs/auth.md#L130-L140)).

internal/runtime/auth_manager_test.go
```diff
@@ -323,8 +323,51 @@ func TestInjectDeclaredSecretsUsesPersistedFallbackForAliases(t *testing.T) {
 	}
 }
 
-func TestInjectDeclaredSecretsRejectsConflictingAliasValues(t *testing.T) {
+// A stale persisted alias must lose to the layered secrets files rather than
+// wedge the run, which is the drift that broke token rotation. Every alias then
+// carries the winning value, so the persisted store heals.
+func TestInjectDeclaredSecretsPrefersLayeredSecretsOverStalePersistedAlias(t *testing.T) {
```
EclipseSourceAI, line 329: Both new tests go through `injectDeclaredSecrets`, but the resolution tests live in `active_secrets_test.go` ([link](https://github.com/eclipse-enclave/enclave/blob/5e808201d3d2284e17c075f1cff4dcb4c214074c/internal/runtime/active_secrets_test.go#L29-L96)). A table test straight on `resolveEnvAliasValue` would cover env-beats-secrets, secrets-beats-persisted and the non-winning-layer cases for almost nothing. The integration test here is still worth keeping since it is the one that proves every alias gets the winner.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Fixes a real wedge: `resolveEnvAliasValue` compared alias values for equality before applying layer precedence, so a stale persisted `GH_TOKEN` hard-failed the run instead of losing to a rotated `GITHUB_TOKEN` in the layered secrets files. The fix picks the highest-priority alias first and only errors when a same-priority alias disagrees; cross-layer drift warns and heals, since `SecretValues` carries the winner for every alias into the persisted store (and `Persist` defaults on).

Approach and scope look right, one function plus tests plus a doc bullet. `make test` and `make lint` are both green here.

Main thing to look at: the three layered secrets files are merged into one map by `auth.ResolveLayeredSecrets`, so they all share `priority: 2`, and alias drift between `global.env` and the project file still hard-fails, the same rotation breakage one level down. Rest of the comments are the docs bullet claiming a stronger rule than the code enforces, layer names in the messages not matching what users see in the docs, and where the new tests live.

internal/runtime/active_secrets.go
sgraband: Right, and that's deliberate: enforcing agreement inside non-winning layers would reintroduce exactly the wedge this PR fixes, since two disagreeing aliases in a stale persisted store would fail the run even though both are about to be overwritten. Only the winning layer is checked. docs/auth.md now states that version instead of the stronger one.

xai (approved):
Thanks for catching this! Fix works fine and looks good 👍

PR #35, opened 2026-07-31
feat: point gitlab-cli at a self-hosted host via GITLAB_HOST

### LINKED ISSUE #4: Let gitlab-cli set the GitLab host directly
Pointing glab at a self-hosted GitLab currently needs a manual workaround: declare `GITLAB_HOST` as a secret so it reaches the container, then edit the mixin spec to add the host to `serviceAuth.hosts`. The host should be a normal setting on the `gitlab-cli` feature instead. Setting it should write glab's `config.yml` host at startup and add the host to `serviceAuth.hosts` and the allowlist, so the gateway injects the token without a spec edit. The goal: one setting makes glab talk to the self-hosted instance with token injection working end to end.

sgraband:
#### What it does
Reaching a self-hosted GitLab needed two manual steps: declare GITLAB_HOST so it reached the container, then edit the mixin spec to put the host in serviceAuth.hosts. Both are now unnecessary.

serviceAuth entries gain hostsFromCredential, naming a credential whose resolved value is a host. When it resolves, its value replaces that service's declared hosts. Replacement rather than addition is deliberate: an instance-specific token must not be released to gitlab.com as well.

The override is resolved during auth setup, before the effective policy is resolved and cached, so the host flows into the allow set through the existing ReleaseHosts union and stays reachable. gitlab-cli declares GITLAB_HOST as a non-apiKey credential and wires all three tokens to it.

Closes #4

#### Breaking changes

#### Review checklist

### DIFF
--- a/docs/extensions/README.md
+++ b/docs/extensions/README.md
@@ -429,12 +429,16 @@ Secrets are split across two `spec.yaml` sections:
 - `network.serviceAuth.<service-id>.hostsFromCredential` names another
   `credentials.sources` id whose resolved value is a host, for services that
   can point at a self-hosted instance. When that credential resolves, its value
-  **replaces** the declared `hosts` for this service — it is not additive,
-  since an instance-specific token must not be released to the public default
-  host as well. The replacement host joins the allow set like a declared one,
-  so one env var is enough to make the tool reach the instance with token
-  injection working. An unset credential keeps the declared hosts; a value that
-  is not a usable host warns and is ignored. Values may be a bare host,
+  becomes the **only** host this service's token is released to, replacing the
+  declared `hosts` and any `serviceDomains` entries pointing at the service —
+  an instance-specific token must not be released to the public default host as
+  well. The network allow set is *not* narrowed: the selected host is added to
+  it, and the declared hosts stay reachable. So one env var is enough to make
+  the tool reach the instance with token injection working. An unset credential
+  keeps the declared hosts as the release targets, and the resolved value is not
+  written to the persisted env store, so it never outlives the run that set it.
+  A value that is not a usable host warns and is ignored; wildcards are
+  rejected, since the point is to name one instance. Values may be a bare host,
   `host:port`, or a full URL — the scheme, port and path are stripped.
 
 The placeholder convention is Go `fmt`-style `%s`, not `{secret}`. An empty
--- a/docs/extensions/README.md
+++ b/docs/extensions/README.md
@@ -428,13 +428,20 @@ Secrets are split across two `spec.yaml` sections:
   `serviceDomains` or from its own `hosts` list.
 - `network.serviceAuth.<service-id>.hostsFromCredential` names another
   `credentials.sources` id whose resolved value is a host, for services that
-  can point at a self-hosted instance. When that credential resolves, its value
+  can point at a self-hosted instance. It does not stand in for the hosts the
+  pairing above requires: the service still needs `hosts` or `serviceDomains`
+  entries, which remain the release targets whenever the credential is unset.
+  When that credential resolves, its value
   becomes the **only** host this service's token is released to, replacing the
   declared `hosts` and any `serviceDomains` entries pointing at the service —
   an instance-specific token must not be released to the public default host as
-  well. The network allow set is *not* narrowed: the selected host is added to
-  it, and the declared hosts stay reachable. So one env var is enough to make
-  the tool reach the instance with token injection working. An unset credential
+  well. Unlike a declared host, which the gateway also applies to everything
+  beneath it, the selected host is matched exactly: the token reaches
+  `gitlab.example.com` and not `runner.gitlab.example.com`. The network allow
+  set is *not* narrowed: the selected host is added to it (with the usual
+  subdomain reach, so the instance's other hosts stay resolvable), and the
+  declared hosts stay reachable. So one env var is enough to make the tool
+  reach the instance with token injection working. An unset credential
   keeps the declared hosts as the release targets, and the resolved value is not
   written to the persisted env store, so it never outlives the run that set it.
   A value that is not a usable host warns and is ignored; wildcards are
--- a/docs/extensions/README.md
+++ b/docs/extensions/README.md
@@ -426,6 +426,16 @@ Secrets are split across two `spec.yaml` sections:
   `network.allowedDomains` instead. The pairing is required in both directions:
   a `serviceAuth` entry also needs at least one host, either from
   `serviceDomains` or from its own `hosts` list.
+- `network.serviceAuth.<service-id>.hostsFromCredential` names another
+  `credentials.sources` id whose resolved value is a host, for services that
+  can point at a self-hosted instance. When that credential resolves, its value
+  **replaces** the declared `hosts` for this service — it is not additive,
+  since an instance-specific token must not be released to the public default
+  host as well. The replacement host joins the allow set like a declared one,
+  so one env var is enough to make the tool reach the instance with token
+  injection working. An unset credential keeps the declared hosts; a value that
+  is not a usable host warns and is ignored. Values may be a bare host,
+  `host:port`, or a full URL — the scheme, port and path are stripped.
 
 The placeholder convention is Go `fmt`-style `%s`, not `{secret}`. An empty
 `valueFormat` means "inject the raw secret value" with no wrapping (see
--- a/extensions/features/gitlab-cli/README.md
+++ b/extensions/features/gitlab-cli/README.md
@@ -8,3 +8,28 @@ requests from the command line. Opt-in (disabled by default).
 ## Installation
 
 Downloads the latest `.deb` release from GitLab's release API. Requires root.
+
+## Auth
+
+`GITLAB_TOKEN`/`GITLAB_ACCESS_TOKEN`, `OAUTH_TOKEN` and
+`JOB_TOKEN`/`CI_JOB_TOKEN` are resolved from the host environment, the layered
+secrets files, or the persisted env store, and injected as env vars. `spec.yaml`
+maps each to the header the gateway injects.
+
+## Self-hosted instances
+
+Set `GITLAB_HOST` to point `glab` at a self-hosted instance:
+
+```bash
+GITLAB_HOST=gitlab.example.com enclave --features +gitlab-cli
+```
+
+That single env var selects the instance for `glab`, retargets token injection,
+and joins the network allowlist — no spec edit or `allow_domains` entry. See
+`serviceAuth.hostsFromCredential` in `docs/extensions/README.md` for the
+accepted value forms and why the host replaces the `gitlab.com` defaults instead
+of adding to them.
+
+To set it persistently, put it in a secrets file rather than the shell — global
+in `~/.local/state/enclave/secrets/global.env`, or per project under
+`~/.local/state/enclave/projects/<hash>/`.
--- a/extensions/features/gitlab-cli/README.md
+++ b/extensions/features/gitlab-cli/README.md
@@ -24,12 +24,14 @@ Set `GITLAB_HOST` to point `glab` at a self-hosted instance:
 GITLAB_HOST=gitlab.example.com enclave --features +gitlab-cli
 ```
 
-That single env var selects the instance for `glab`, retargets token injection,
-and joins the network allowlist — no spec edit or `allow_domains` entry. See
-`serviceAuth.hostsFromCredential` in `docs/extensions/README.md` for the
-accepted value forms and why the host replaces the `gitlab.com` defaults instead
-of adding to them.
+That single env var selects the instance for `glab`, retargets token injection
+at it, and joins the network allowlist — no spec edit or `allow_domains` entry.
+The tokens then go only to that instance, while `gitlab.com` stays reachable.
+See `serviceAuth.hostsFromCredential` in `docs/extensions/README.md` for the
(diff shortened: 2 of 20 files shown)


sdirix:
@EclipseSourceAI

internal/runtime/active_secrets.go
```diff
@@ -163,6 +166,79 @@ func resolveEnvAliasValue(secret activeSecret, layeredSecrets map[string]string,
...
+	}
+	// NormalizeHost strips the port and IPv6 brackets but does not validate the
+	// labels, so the pattern validator runs after it to reject junk that would
+	// otherwise become a bogus allowlist entry.
+	host, err := domainpattern.NormalizeHost(trimmed)
+	if err == nil {
+		host, err = domainpattern.Normalize(host)
+	}
+	if err != nil {
+		logx.Warnf("Secret %s: value %q is not a usable host (%v); keeping declared hosts.", secretID, value, err)
```
EclipseSourceAI: This prints the credential value verbatim at warn level. If `hostsFromCredential` ever points at a token (or `GITLAB_HOST` gets a token value by mistake), the secret lands in the terminal. Everywhere else secret-derived values go through `util.RedactSecret`, and at debug level: https://github.com/eclipse-enclave/enclave/blob/5e808201d3d2284e17c075f1cff4dcb4c214074c/internal/runtime/auth_manager.go#L208-L208

internal/runtime/runtime.go
```diff
@@ -982,15 +988,39 @@ func (r *Runtime) specProxyManaged() []string {
...
+		allowed = append(allowed, r.releaseHosts(feature.Secrets)...)
 	}
 	return allowed, denied
 }
 
+// releaseHosts is model.ReleaseHosts with any resolved
+// serviceAuth.hostsFromCredential replacement applied, so a host selected at
+// runtime becomes resolvable exactly like a declared one. The loaded spec is
+// left untouched.
+func (r *Runtime) releaseHosts(secrets map[string]model.SecretConfig) []string {
```
EclipseSourceAI, line 1010: The replacement narrows the DNS allow set too, not just the token release targets. Only `claude`'s built-in allowlist pulls in the gitlab fragment (https://github.com/eclipse-enclave/enclave/blob/5e808201d3d2284e17c075f1cff4dcb4c214074c/extensions/tools/claude/gateway-allowlist.conf#L27-L27), so for `codex`/`opencode`/`theia` the feature's release hosts are the only source of `gitlab.com` and setting `GITLAB_HOST` makes gitlab.com unresolvable for the whole session. The rationale in the PR description only needs the release hosts narrowed, so `releaseHosts` could keep the declared hosts while `releaseHostsFor` uses only the override.

internal/runtime/auth_manager.go
```diff
@@ -168,6 +168,9 @@ func (m authManager) injectDeclaredSecrets(hooks auth.Hooks, authCtx auth.Contex
 	for _, secret := range suppressedActiveSecrets {
 		logx.Debugf("Suppressed declared API key secret %s due to %s", secret.ID, suppressionReason)
 	}
+	// Must precede shouldUseSecretReleases: that call resolves and caches the
+	// effective policy, which unions the release hosts into the allow set.
+	m.releaseHostOverrides = resolveReleaseHostOverrides(eligibleSecrets, m.host.Home, layeredSecrets, persistedEnv)
```
EclipseSourceAI: The invariant is enforced only by this comment, and `resolveEffectivePolicy` memoizes on the Runtime (https://github.com/eclipse-enclave/enclave/blob/5e808201d3d2284e17c075f1cff4dcb4c214074c/internal/runtime/runtime.go#L1507-L1511). It holds today, but a future policy lookup anywhere in `prepareMounts` (which runs before `prepareVolumes`) would silently cache an allow set without the override while the token still targets it. Either reset `policyResolved` here when overrides are non-empty, or fail loudly if the policy is already resolved.

extensions/features/gitlab-cli/README.md
```diff
@@ -8,3 +8,28 @@ requests from the command line. Opt-in (disabled by default).
...
+
+That single env var selects the instance for `glab`, retargets token injection,
+and joins the network allowlist — no spec edit or `allow_domains` entry. See
+`serviceAuth.hostsFromCredential` in `docs/extensions/README.md` for the
+accepted value forms and why the host replaces the `gitlab.com` defaults instead
+of adding to them.
+
+To set it persistently, put it in a secrets file rather than the shell — global
+in `~/.local/state/enclave/secrets/global.env`, or per project under
+`~/.local/state/enclave/projects/<hash>/`.
```
EclipseSourceAI: Wrong path. Per-project secrets live at `~/.local/state/enclave/secrets/projects/<hash>/<tool>.env` (https://github.com/eclipse-enclave/enclave/blob/5e808201d3d2284e17c075f1cff4dcb4c214074c/docs/auth.md#L136-L138), and that file is keyed by the tool (e.g. `claude.env`), not by the feature, which is worth spelling out since this is a mixin.

extensions/features/gitlab-cli/README.md
```diff
@@ -8,3 +8,28 @@ requests from the command line. Opt-in (disabled by default).
...
+## Self-hosted instances
+
+Set `GITLAB_HOST` to point `glab` at a self-hosted instance:
+
+```bash
+GITLAB_HOST=gitlab.example.com enclave --features +gitlab-cli
+```
+
+That single env var selects the instance for `glab`, retargets token injection,
+and joins the network allowlist — no spec edit or `allow_domains` entry. See
```
EclipseSourceAI: True at session start, but `enclave network apply` recomputes the desired bundle from the tool profile only, with no features and no runtime overrides (https://github.com/eclipse-enclave/enclave/blob/5e808201d3d2284e17c075f1cff4dcb4c214074c/internal/policy/effective_resolver.go#L134-L143), then rewrites it (https://github.com/eclipse-enclave/enclave/blob/5e808201d3d2284e17c075f1cff4dcb4c214074c/internal/app/network_cmd.go#L291-L295). So `network status` reports drift and an apply cuts the self-hosted host out of a running session. The gap already exists for `gitlab.com`, but a follow-up issue would be good since the override makes it the only host the token can reach.

docs/extensions/README.md
```diff
@@ -408,6 +408,16 @@ Secrets are split across two `spec.yaml` sections:
   describe how the gateway injects that credential as an HTTP header when it
   proxies requests to those hosts. The service-id in `serviceAuth` and
   `serviceDomains` is the same key used under `credentials.sources`.
+- `network.serviceAuth.<service-id>.hostsFromCredential` names another
+  `credentials.sources` id whose resolved value is a host, for services that
+  can point at a self-hosted instance. When that credential resolves, its value
+  **replaces** the declared `hosts` for this service — it is not additive,
```
EclipseSourceAI: It replaces the union of `hosts` and the `serviceDomains`-derived hosts, not just the declared `hosts`: both go into the same list in `buildSecrets` (https://github.com/eclipse-enclave/enclave/blob/5e808201d3d2284e17c075f1cff4dcb4c214074c/internal/config/spec_map.go#L115-L124). Matters for specs like `github-cli` that express hosts via `serviceDomains`.

extensions/features/gitlab-cli/spec.yaml
```diff
@@ -18,8 +18,11 @@ credentials:
     gitlab-token: { env: [GITLAB_TOKEN, GITLAB_ACCESS_TOKEN] }
     gitlab-oauth-token: { env: [OAUTH_TOKEN] }
     gitlab-job-token: { env: [JOB_TOKEN, CI_JOB_TOKEN] }
+    # Not a credential: glab reads GITLAB_HOST to pick the instance, and
+    # hostsFromCredential below retargets token injection at it.
+    gitlab-host: { env: [GITLAB_HOST], apiKey: false }
```
EclipseSourceAI, line 23: Because this is a declared credential, the resolved `GITLAB_HOST` is written to the persisted env store under the default `--persist`, so later runs without the env var keep the self-hosted override (and keep gitlab.com out of the release hosts). That contradicts "An unset credential keeps the declared hosts" in the docs, so either document how to clear it or skip persistence for non-`apiKey` host credentials.

internal/runtime/active_secrets.go
```diff
@@ -163,6 +166,79 @@ func resolveEnvAliasValue(secret activeSecret, layeredSecrets map[string]string,
...
+		}
+		if !found {
+			continue
+		}
+		host := normalizeReleaseHost(ref, value)
+		if host == "" {
+			continue
+		}
+		for _, id := range secretIDs {
+			overrides[id] = []string{host}
```
EclipseSourceAI, line 246: A successful override is completely silent, and the injection debug log redacts the value, so there is no way to see which host was picked. An info line here would help, similar to how ad-hoc allowlisted domains are reported: https://github.com/eclipse-enclave/enclave/blob/5e808201d3d2284e17c075f1cff4dcb4c214074c/internal/runtime/network_manager.go#L81-L81

internal/runtime/active_secrets.go
```diff
@@ -163,6 +166,79 @@ func resolveEnvAliasValue(secret activeSecret, layeredSecrets map[string]string,
...
+	}
+	if index := strings.IndexAny(trimmed, "/?#"); index >= 0 {
+		trimmed = trimmed[:index]
+	}
+	// NormalizeHost strips the port and IPv6 brackets but does not validate the
+	// labels, so the pattern validator runs after it to reject junk that would
+	// otherwise become a bogus allowlist entry.
+	host, err := domainpattern.NormalizeHost(trimmed)
+	if err == nil {
+		host, err = domainpattern.Normalize(host)
```
EclipseSourceAI, line 278: `domainpattern.Normalize` accepts wildcard patterns, so `GITLAB_HOST=*.example.com` becomes a wildcard release target and allowlist entry. That works against the "names the one instance the token belongs to" rationale, so rejecting `*` here seems right.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.** Generated automatically, so it may contain mistakes. Feel free to ignore any comment you disagree with (noting *why* helps future reviews).
>
> This review does **not** approve the change. A human still needs to review it and sign off on the overall architecture and design.
>
> To get an updated review after pushing changes, re-request a review from this account.
>

Adds `network.serviceAuth.<id>.hostsFromCredential` so a service's HTTP-release hosts can be selected at runtime from another declared credential, and wires `gitlab-cli`'s three tokens to a new `gitlab-host` (`GITLAB_HOST`) credential. The plumbing fits the existing shape well: spec validation sits next to the existing `serviceAuth` cross-reference check, the runtime reuses `resolveActiveSecretValue` and `domainpattern`, and `releaseHosts` clones the secret map so the loaded spec is not mutated. Build, tests, `make lint` and `make generate` are all clean.

Two areas deserve a human look. First, replacement narrows the DNS allow set as well as the token release targets, so for tools whose built-in allowlist lacks the gitlab fragment (everything except `claude`) setting `GITLAB_HOST` makes `gitlab.com` unreachable for the whole session, which the stated rationale does not require. Second, the override lands in a mutable `Runtime` field whose correctness depends on running before the memoized effective-policy resolution, enforced only by a comment.

The rest are smaller: an unredacted credential value in a warning, a wrong per-project secrets path in the feature README, `network apply` recomputing the bundle without the override, and the host sticking via the persisted env store. Details inline.

extensions/features/gitlab-cli/README.md
sgraband: Confirmed, and it predates this change: network apply rebuilds from the tool profile only, so feature release hosts already drift. The override makes it more visible rather than new. Tracking separately rather than widening this PR into the apply path.

internal/runtime/active_secrets.go
```diff
@@ -185,6 +190,99 @@ func resolveAlias(envVar string, secretsLayers []auth.SecretsLayer, persistedEnv
...
+			continue
+		}
+		host := normalizeReleaseHost(ref, value)
+		if host == "" {
+			continue
+		}
+		logx.Infof("Host credential %s selects %s for %s", ref, host, strings.Join(secretIDs, ", "))
+		for _, id := range secretIDs {
+			overrides[id] = []string{host}
+		}
```
xai, line 247: Selected host releases the token to every subdomain, not one instance"
The selected host is stored in the override as a plain domain pattern and reaches the gateway as rule.Hosts. The gateway checks release targets with hostAllowed (`internal/gateway/mitm/proxy.go:726, 747`), which calls `domainpattern.MatchNormalizedHost`, and that function treats a non-wildcard pattern as apex-plus-all-subdomains (`internal/domainpattern/pattern.go:99-102`). So `GITLAB_HOST=gitlab.example.com` releases the token to `anything.gitlab.example.com`. Possible fix is small: mark selector-derived release hosts as exact (e.g. an exact-hosts field on `model.SecretReleaseEntry` threaded through `normalizeRule` in `proxy.go`, matched by equality in `rewriteHeaders`) while the allow-set entry keeps its current semantics, plus a test that a subdomain of the selected host is denied. If you consciously prefer to keep suffix matching, the docs' exclusivity claims and the wildcard-rejection comment must be reworded so they no longer promise it.

xai (changes requested):
Looks good and works!

Fable found one issue that is worth fixing before the merge, see the inline comment in `active_secrets.go`.

xai (approved):
Thanks for the super quick follow-up!
Everything addressed!

PR #37, opened 2026-08-03
Add project-scoped feature state stores

xai:
#### What it does

Adds an opt-in state store for features that need to keep project data across sessions and tools. A mixin can declare `state: true`; in persistent sessions Enclave creates `~/.local/state/enclave/projects/<hash>/features/<feature>/state/` and mounts it at `~/.enclave-feature-state/<feature>/`.

While sourcing that feature's setup scripts, the entrypoint exposes the mount as `ENCLAVE_FEATURE_STATE_DIR`. It unsets the variable before sourcing the next feature and before starting the agent. Ephemeral sessions do not create or mount feature state.

Feature state is represented in both backend store layers, although current QEMU sessions disable mixins and do not mount it. It is independent of auth scope and remains separate from feature auth. `enclave cleanup` removes it with the rest of the project data unless `--keep feature-state` is set, including during `cleanup --all`.

Mixin validation rejects `state` on sandbox specs. The sandbox extension name `features` is now reserved because it would collide with the project feature-state namespace. The image build also makes `kit-init.sh` world-readable, so startup does not depend on the checkout's umask.

#### How to test

Current QEMU sessions disable mixins, so this manual test uses Docker. It creates a temporary `state-probe` extension that records which tools start with its state store mounted.

Run the setup commands from the repository root. `enclave shell` is interactive; run the checks below at the container prompt.

```bash
repo="$PWD"
make build
enclave_bin="$repo/bin/enclave"
test_root="$(mktemp -d)"
project_dir="$test_root/project"

export XDG_CONFIG_HOME="$test_root/config"
export XDG_STATE_HOME="$test_root/state"

probe_dir="$XDG_CONFIG_HOME/enclave/extensions/features/state-probe"
mkdir -p "$probe_dir/feature-entrypoint.d" "$project_dir"
cat >"$probe_dir/spec.yaml" <<'YAML'
schemaVersion: "1"
kind: mixin
name: state-probe
displayName: State Probe
description: Manual feature-state probe
defaultEnabled: false
state: true
YAML
cat >"$probe_dir/feature-entrypoint.d/setup.sh" <<'SH'
if [ -n "${ENCLAVE_FEATURE_STATE_DIR:-}" ]; then
    printf '%s\n' "${TOOL:-unknown}" >> "$ENCLAVE_FEATURE_STATE_DIR/runs"
fi
SH

cd "$project_dir"
"$enclave_bin" shell --tool claude --features state-probe --rebuild
```

Inside the Claude shell:

```bash
cat "$HOME/.enclave-feature-state/state-probe/runs"  # claude
printf '%s\n' "${ENCLAVE_FEATURE_STATE_DIR-unset}"   # unset
exit
```

Start the same extension with Codex:

```bash
"$enclave_bin" shell --tool codex --features state-probe --rebuild
```

Inside the Codex shell, the shared file should contain both tool names:

```bash
cat "$HOME/.enclave-feature-state/state-probe/runs"
# claude
# codex
exit
```

Back on the host, clean up and run the automated checks:

```bash
cd "$repo"
rm -rf "$test_root"
unset XDG_CONFIG_HOME XDG_STATE_HOME
make test
make lint
```

The unit tests cover stateless and ephemeral sessions, cleanup retention, and both backend store layers.

#### Follow-ups

None.

#### Breaking changes

The sandbox extension name `features` is now reserved.

#### Review checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -335,7 +335,8 @@ COPY entrypoint.sh /usr/local/bin/entrypoint.sh
 RUN chmod a+rx /usr/local/bin/entrypoint.sh \
         /usr/local/share/enclave/auth-reconcile.sh \
         /usr/local/share/enclave/net.sh && \
-    chmod a+r /usr/local/share/enclave/tmux-session.conf
+    chmod a+r /usr/local/share/enclave/tmux-session.conf \
+        /usr/local/share/enclave/kit-init.sh
 USER ${USERNAME}
 ENTRYPOINT ["/usr/local/bin/entrypoint.sh"]
 CMD ["/bin/bash"]
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -296,9 +296,15 @@ fi
 # feature's install never ran, so its runtime entrypoint must not fire either.
 # enclave_feature_enabled fails open when the manifest is absent (or the
 # helper is unavailable), preserving the prior all-features behavior.
+enclave_feature_state_root="${ENCLAVE_FEATURE_STATE_ROOT:-}"
 for ext_dir in "$enclave_features_dir"/*/; do
+    feature_name="$(basename "${ext_dir%/}")"
     if command -v enclave_feature_enabled >/dev/null 2>&1; then
-        enclave_feature_enabled "$(basename "${ext_dir%/}")" || continue
+        enclave_feature_enabled "$feature_name" || continue
+    fi
+    unset ENCLAVE_FEATURE_STATE_DIR
+    if [ -n "$enclave_feature_state_root" ] && [ -d "$enclave_feature_state_root/$feature_name" ]; then
+        export ENCLAVE_FEATURE_STATE_DIR="$enclave_feature_state_root/$feature_name"
     fi
     if [ -d "${ext_dir}feature-entrypoint.d" ]; then
         for script in "${ext_dir}feature-entrypoint.d"/*.sh; do
@@ -306,7 +312,9 @@ for ext_dir in "$enclave_features_dir"/*/; do
             [ -f "$script" ] && . "$script"
         done
     fi
+    unset ENCLAVE_FEATURE_STATE_DIR
 done
+unset ENCLAVE_FEATURE_STATE_ROOT enclave_feature_state_root
 
 # Seed extension init files declared in spec.yaml commands.initFiles, copy any
 # files/workspace trees into the project (never clobbering host files), then run
--- a/internal/app/cleanup.go
+++ b/internal/app/cleanup.go
@@ -176,6 +176,10 @@ func resolveEphemeralStoreDirs(run model.RunOptions, cleanup model.CleanupOption
 			tools = []string{run.Tool}
 		}
 		for _, tool := range tools {
+			// The project-level features namespace is not a tool config-store root.
+			if config.HostProjectToolDir(home, hash, tool) == config.HostStoreFeatureStateRootDir(home, hash) {
+				continue
+			}
 			storeRoot := config.HostStoreConfigRootDir(home, tool, hash)
 			for _, key := range listSubdirs(storeRoot) {
 				if key == persistentConfigStoreKey {
@@ -214,12 +218,15 @@ func resolveCleanupDirs(run model.RunOptions, cleanup model.CleanupOptions, home
 	if cleanup.CleanupAll {
 		dirs := []cleanupDir{
 			{Kind: "cache", Path: config.HostCacheDir(home)},
-			// The state projects tree holds every project's config/env stores
-			// and history, so a full cleanup removes them all at once.
-			{Kind: "history", Path: config.HostProjectsDir(home)},
 			// The image inbox is global (not project-scoped), so it is only
 			// removed by a full cleanup. Held images are user-imported content.
 			{Kind: "inbox", Path: config.HostImageInboxDir(home)},
+		}
+		if cleanup.CleanupKeepHist || cleanup.CleanupKeepFeatureState {
+			dirs = append(dirs, selectiveAllProjectCleanupDirs(home, cleanup)...)
+		} else {
+			// Without selective project-state retention, remove the whole tree.
+			dirs = append(dirs, cleanupDir{Kind: "history", Path: config.HostProjectsDir(home)})
 		}
 		// Shared tool/feature auth stores live outside the projects tree; they
 		// are removed on a full cleanup unless `--keep auth` is set.
@@ -236,7 +243,42 @@ func resolveCleanupDirs(run model.RunOptions, cleanup model.CleanupOptions, home
 		{Kind: "history", Path: config.HostStoreConfigRootDir(home, run.Tool, project.Hash)},
 		{Kind: "history", Path: config.HostStoreEnvDir(home, run.Tool, project.Hash)},
 		{Kind: "memory", Path: config.HostProjectMemoryDir(home, project.Hash, run.Tool)},
+		{Kind: "feature-state", Path: config.HostStoreFeatureStateRootDir(home, project.Hash)},
 	}
+}
+
+// selectiveAllProjectCleanupDirs expands the projects tree only when a keep
+// flag needs to retain one project-scoped namespace while deleting another.
+// Projects without matching retained data can still be removed as a whole.
+func selectiveAllProjectCleanupDirs(home string, cleanup model.CleanupOptions) []cleanupDir {
+	var dirs []cleanupDir
+	for _, hash := range listSubdirs(config.HostProjectsDir(home)) {
+		projectDir := config.HostProjectDir(home, hash)
+		featureStateRoot := config.HostStoreFeatureStateRootDir(home, hash)
+		children := listSubdirs(projectDir)
+		keepsChild := false
+		for _, child := range children {
+			path := filepath.Join(projectDir, child)
+			if path == featureStateRoot {
+				keepsChild = keepsChild || cleanup.CleanupKeepFeatureState
+			} else {
+				keepsChild = keepsChild || cleanup.CleanupKeepHist
+			}
+		}
+		if !keepsChild {
+			dirs = append(dirs, cleanupDir{Kind: "project", Path: projectDir})
+			continue
+		}
+		for _, child := range children {
+			path := filepath.Join(projectDir, child)
+			kind := "history"
+			if path == featureStateRoot {
+				kind = "feature-state"
+			}
+			dirs = append(dirs, cleanupDir{Kind: kind, Path: path})
+		}
+	}
+	return dirs
 }
 
 // authStoreCleanupDirs enumerates the shared tool and feature auth store
@@ -269,6 +311,9 @@ func cleanupDirsForRemoval(run model.RunOptions, cleanup model.CleanupOptions, h
 	}
 	if cleanup.CleanupKeepAuth {
 		dirs = filterDirs(dirs, "auth")
+	}
+	if cleanup.CleanupKeepFeatureState {
+		dirs = filterDirs(dirs, "feature-state")
 	}
 	return dirs
 }
--- a/internal/app/cleanup_test.go
+++ b/internal/app/cleanup_test.go
@@ -232,17 +232,20 @@ func TestResolveEphemeralStoreDirsSkipsDefaultKey(t *testing.T) {
 	}
 }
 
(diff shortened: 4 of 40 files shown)


Dockerfile
```diff
@@ -335,7 +335,8 @@ COPY entrypoint.sh /usr/local/bin/entrypoint.sh
 RUN chmod a+rx /usr/local/bin/entrypoint.sh \
         /usr/local/share/enclave/auth-reconcile.sh \
         /usr/local/share/enclave/net.sh && \
-    chmod a+r /usr/local/share/enclave/tmux-session.conf
+    chmod a+r /usr/local/share/enclave/tmux-session.conf \
+        /usr/local/share/enclave/kit-init.sh
```
EclipseSourceAI, line 339: The QEMU bundle needs the same fix. `build-bundle.sh` copies `kit-init.sh` with plain `cp` and its chmod line only covers entrypoint/auth-reconcile/net.sh ([link](https://github.com/eclipse-enclave/enclave/blob/5e808201d3d2284e17c075f1cff4dcb4c214074c/runtime-assets/microvm/alpine/build-bundle.sh#L115-L116)), and the later `chown -R agent` doesn't touch `/usr/local/share/enclave`. Since the bundle bakes an empty `.installed-features` manifest specifically to gate mixins off and that gate lives in `kit-init.sh`, an unreadable copy fails open and runs every baked feature's hooks.

entrypoint.sh
```diff
@@ -296,17 +296,25 @@ fi
...
     fi
     if [ -d "${ext_dir}feature-entrypoint.d" ]; then
         for script in "${ext_dir}feature-entrypoint.d"/*.sh; do
             # shellcheck disable=SC1090 # Sourcing feature entrypoints at runtime.
             [ -f "$script" ] && . "$script"
         done
     fi
+    unset ENCLAVE_FEATURE_STATE_DIR
 done
+unset ENCLAVE_FEATURE_STATE_ROOT enclave_feature_state_root
```
EclipseSourceAI, line 317: Unsetting the root here means neither `ENCLAVE_FEATURE_STATE_ROOT` nor `ENCLAVE_FEATURE_STATE_DIR` is visible to the `initFiles` / `workspace` / `startup` loops that follow ([link](https://github.com/eclipse-enclave/enclave/blob/5e808201d3d2284e17c075f1cff4dcb4c214074c/entrypoint.sh#L316-L332)). A stateful service mixin declaring `commands.startup` can then only find its store if a `feature-entrypoint.d` script planted a symlink first, and `initFiles` can't reference it at all (envsubst whitelist is WORKDIR/HOME/USER). Either keep the root exported through those loops or say so in the extension docs.

internal/runtime/runtime.go
```diff
@@ -1463,6 +1464,27 @@ func (r *Runtime) addFeatureAuthMounts(mounts *mountAccumulator, featureStores m
...
+		if !ok {
+			continue
+		}
+		mountPath := mountRoot + "/" + feat.Name
+		mounts.AddStore(backend.PersistentStore{Kind: store.Kind, Key: store.Key, ContainerPath: mountPath})
+		mounted = true
+		logx.Infof("%s feature state mounted", util.TitleCase(feat.Name))
+	}
+	if mounted {
+		mounts.AddEnv(model.EnvFeatureStateRoot, mountRoot)
```
EclipseSourceAI, line 1484: Exporting only the root and letting the entrypoint infer per-feature opt-in from `[ -d "$root/$feature_name" ]` diverges from how feature auth does it: that one ships a declared map, `ENCLAVE_FEATURE_AUTH_MAP` ([link](https://github.com/eclipse-enclave/enclave/blob/5e808201d3d2284e17c075f1cff4dcb4c214074c/internal/runtime/runtime.go#L1445-L1462), consumed at [entrypoint.sh#L170](https://github.com/eclipse-enclave/enclave/blob/5e808201d3d2284e17c075f1cff4dcb4c214074c/entrypoint.sh#L170-L171)). With the directory probe a feature that declared `state: true` but whose mount got skipped looks identical to one that never opted in.

internal/app/cleanup.go
```diff
@@ -236,7 +243,42 @@ func resolveCleanupDirs(run model.RunOptions, cleanup model.CleanupOptions, home
...
+		for _, child := range children {
+			path := filepath.Join(projectDir, child)
+			if path == featureStateRoot {
+				keepsChild = keepsChild || cleanup.CleanupKeepFeatureState
+			} else {
+				keepsChild = keepsChild || cleanup.CleanupKeepHist
+			}
+		}
+		if !keepsChild {
+			dirs = append(dirs, cleanupDir{Kind: "project", Path: projectDir})
```
EclipseSourceAI, line 269: `Kind: "project"` matches none of the filters in `cleanupDirsForRemoval`, so this whole-project shortcut is unfilterable by any `--keep`. It happens to be safe because only `history` and `feature-state` gate the projects tree today, but it will silently ignore any future project-scoped keep kind. The `keepsChild` loop above also reduces to two checks: does the project have a `features` child, does it have any other child.

internal/backend/types.go
```diff
@@ -90,12 +90,25 @@ type Mount struct {
...
-	StoreKindEnv         StoreKind = "env"
+	StoreKindConfig       StoreKind = "config"
+	StoreKindAuth         StoreKind = "auth"
+	StoreKindFeatureAuth  StoreKind = "feature-auth"
+	StoreKindFeatureState StoreKind = "feature-state"
+	StoreKindEnv          StoreKind = "env"
 )
 
+// FeatureStoreLabel returns the user-facing label for a feature-owned store.
+func (kind StoreKind) FeatureStoreLabel() string {
```
EclipseSourceAI, line 101: This is log-message formatting hanging off a shared domain type, and it returns `"feature store"` for `config`/`auth`/`env`, which aren't feature stores. It's only used by the two `logx` lines in the Docker and QEMU prepare paths; a local helper or a small map in each backend keeps the presentation string out of `backend`.

internal/app/cleanup_test.go
```diff
@@ -169,6 +232,20 @@ func TestResolveEphemeralStoreDirsSkipsDefaultKey(t *testing.T) {
 	}
 }
 
+func TestResolveEphemeralStoreDirsSkipsFeatureStateNamespace(t *testing.T) {
```
EclipseSourceAI: This sets `run.Tool = "features"`, which is the non-`--all` branch, and a sandbox named `features` is now rejected at spec load so that can't happen in practice. The branch the guard actually protects is `--all`, where `tools` comes from `listSubdirs(HostProjectDir(...))`. Set `CleanupAll: true` and add a real tool dir alongside the feature-state tree.

internal/runtime/ports_test.go
```diff
@@ -226,8 +226,8 @@ func TestApplyFeaturePortsPublishesEnabledFeatureDeclarations(t *testing.T) {
 	r := &Runtime{
 		profile: model.Profile{Name: "claude"},
 		features: []model.Extension{
-			{Name: "diffity", Ports: []model.PortConfig{
-				{Container: 5391, HostAllocation: model.HostAllocationAuto, Publish: true, Label: "Diffity"},
+			{Name: "preview", Ports: []model.PortConfig{
```
EclipseSourceAI, line 229: Renaming the `diffity` fixture to `preview` is unrelated to feature state. Separate commit.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Adds an opt-in project-scoped state store for feature mixins (`state: true`), mounted at `~/.enclave-feature-state/<feature>/` and keyed by feature + project hash but not by tool, so tools in the same project share it.

The store integration is clean: a new `StoreKind`, one `hoststore` case, one path helper, and both backends pick it up through the generalized `StorePrep.FeatureStores`. Nice side effects are that `hoststore.DirFor` no longer silently falls through to the config layout for unknown kinds, and feature auth and feature state are no longer mutually exclusive. Owner and project hash both go through `validateStoreSegment`, and the docs (`docs/runtime/stores.md`) are accurate about the cross-tool trust channel and the `--ephemeral` / `--auth-scope` interactions.

Worth a human look at three spots: the `kit-init.sh` permission fix in the Dockerfile has no counterpart in `runtime-assets/microvm/alpine/build-bundle.sh`, where the same umask-dependent copy defeats the empty-manifest feature gate; `entrypoint.sh` unsets the state root before the `initFiles`/`startup` loops, so a stateful service mixin can't locate its store from a startup command; and the restructured `--all` path in `internal/app/cleanup.go` introduces a `"project"` cleanup kind that no `--keep` filter matches.

`make build`, `make test`, and `make lint` all pass.

internal/runtime/ports_test.go
xai: i don't care, this will be done now :)

internal/backend/docker/prepare.go
```diff
@@ -34,14 +34,15 @@ func (b *Backend) PrepareStores(ctx context.Context, prep backend.StorePrep) (ba
 		b.prepareConfigStore(*prep.Config)
 	}
 	if prep.Auth != nil {
-		b.ensureStoreDir(backend.StoreKindAuth, prep.Auth.Key, util.TitleCase(prep.Auth.Key.Owner)+" shared auth store")
+		b.ensureStoreDir(prep.Auth.Kind, prep.Auth.Key, util.TitleCase(prep.Auth.Key.Owner)+" shared auth store")
```
EclipseSourceAI: Taking the kind from `prep.Auth.Kind` buys nothing here (`prep.Auth` is by definition the shared tool auth store) and adds a failure mode: with the new erroring `default` case in `DirFor`, a caller that forgets `Kind` now silently loses the shared auth store, since `ensureStoreDir` only logs the resolve failure at debug level ([link](https://github.com/eclipse-enclave/enclave/blob/5e808201d3d2284e17c075f1cff4dcb4c214074c/internal/backend/docker/prepare.go#L88-L92)). Same in `qemu/prepare.go`. Keeping `backend.StoreKindAuth` hardcoded for this one field would be safer.

internal/config/spec_loader.go
```diff
@@ -49,6 +49,12 @@ func LoadSpec(paths model.Paths, name string, kind string) (specDocument, error)
 	if doc.Name != name {
 		return specDocument{}, fmt.Errorf("%s name must be %q", specPath, name)
 	}
+	if kind == KindSandbox && name == projectFeatureStateDirName {
```
EclipseSourceAI, line 52: The reservation only fires once the spec is actually loaded. `ListTools` just checks for a spec file ([link](https://github.com/eclipse-enclave/enclave/blob/5e808201d3d2284e17c075f1cff4dcb4c214074c/internal/config/extension.go#L93-L104)), so a user tool dir named `features` still shows up in `--help` and completion and only errors when someone selects it. Filtering it there too would make the docs claim ("the sandbox name `features` is reserved") hold end to end.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Follow-up review of the project-scoped feature state store.

Two commits landed since the last pass: the QEMU bundle now chmods `kit-init.sh` readable, and the ephemeral-cleanup guard test was rewritten to actually hit the `--all` branch (it now trips on a feature named `config-store`, which is exactly the collision the guard exists for). Both look right.

The store integration itself stays small and well placed: one `StoreKind`, one `hoststore` case, two path helpers, and both backends pick it up via the generalized `StorePrep.FeatureStores`. Docs across stores.md, persistence.md, ARCHITECTURE.md, cli-reference.md and extensions/README.md match the code, including the concurrency and cross-tool caveats. `make build`, `make test`, `make lint` pass.

Where a maintainer should focus: the restructured `--all` path in `internal/app/cleanup.go` (the `"project"` kind is unfilterable by any `--keep`), the entrypoint env lifecycle around the `initFiles`/`startup` loops, and the blast radius of `enclave cleanup --tool <x>`, which now wipes the whole project's feature state including data other tools use. Also worth deciding whether the new cross-tool write channel deserves a mention in `docs/security/` rather than only `docs/runtime/stores.md`. No shipped extension sets `state: true` yet, so the path is exercised only by tests.

Two new low-severity comments inline; the four earlier open threads still apply.

- [kit-init.sh needs the same world-readable fix in the QEMU bundle](https://github.com/eclipse-enclave/enclave/pull/37#discussion_r3706311067)
- [feature-state namespace guard test should cover the --all branch](https://github.com/eclipse-enclave/enclave/pull/37#discussion_r3706311555)

PR #38, opened 2026-08-04
feat: install zsh completions

planger:
## Problem

`make install` and the `.deb`/`.rpm` packages generated and shipped only the bash completion script, so zsh users got no completion for `enclave` at all and fell back to filename completion.

The bash side works without any user setup because bash-completion 2.x ships a dynamic loader whose first search directory is a user-level XDG path (`${XDG_DATA_HOME:-$HOME/.local/share}/bash-completion/completions`). Zsh has no counterpart: its default `fpath` is compiled in and contains only root-owned directories, with no XDG user entry and no lazy loader.

## Change

`enclave completion zsh` already emitted a correct `#compdef` script; it was simply never installed. Now it is:

- `make install` installs `_enclave` to `~/.local/share/zsh/site-functions/`, and `make uninstall` removes it.
- `.deb` installs to `/usr/share/zsh/vendor-completions/`, `.rpm` to `/usr/share/zsh/site-functions/`. The paths differ because Debian reserves `vendor-completions` for distro packages and does not keep `/usr/share/zsh/site-functions` on `fpath`, while Fedora uses `site-functions`. Both are on the respective default `fpath`, so package installs need no user action.
- `scripts/verify-package-assets.sh` now asserts that both completions are present in the built package, accepting either zsh layout. Completion files previously had no packaging coverage.
- `scripts/clean-state.sh` knows the new install locations.

A user-level install cannot be made zero-config the way the bash one is, since no writable directory is on zsh's default `fpath`. Rather than editing `~/.zshrc` from a build target, `make install` prints a single hint line with the `fpath` line to add, guarded by `command -v zsh` so it stays silent where zsh is absent and can never fail the target. The README documents the same for both shells.

## Verification

- `make install-binary` and `make uninstall` into a throwaway `HOME`: both completion files placed and removed, hint renders `$fpath` literally, silent and exit 0 when zsh is missing.
- Installed layout works end to end: with the directory on `fpath`, `compinit` registers `_comps[enclave]=_enclave`, and completion offers subcommands with descriptions, per-subcommand flags (`enclave ps --` gives `--all`, `--json`, `--name`, `--tool`) and enum values (`--tool` gives the seven tool profiles).
- `debian/rules override_dh_auto_install` staged into a temp dir: `verify-package-assets.sh` passes, including the new assertions, and stages `usr/share/zsh/vendor-completions/_enclave`.
- `make build`, `make test`, `make check-license-headers` all pass. The generated zsh script carries the license header, so it is covered by the header check like the bash one.

## Known limitations

- The rpm path was not exercised locally: no `rpmbuild` on the dev host, and `debhelper` was missing so `make deb-quick` could not run either. The packaging CI jobs cover both, and the new `verify-package-assets.sh` assertion will fail loudly if either path is wrong.
- Completions are still Linux only. `make install` on macOS installs no completion for either shell, unchanged by this PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,11 @@ LINT_TMP_DIR := ./.tmp
 LINT_GO_CACHE := $(LINT_TMP_DIR)/go-build
 LINT_GOLANGCI_CACHE := $(LINT_TMP_DIR)/golangci-lint-cache
 BASE_REF ?=
+# Completion install roots follow the XDG base directory spec, like
+# scripts/clean-state.sh, so install and cleanup agree when XDG_DATA_HOME is set.
+DATA_HOME ?= $(or $(XDG_DATA_HOME),$(HOME)/.local/share)
+ZSH_COMPLETION_DIR := $(DATA_HOME)/zsh/site-functions
+BASH_COMPLETION_DIR := $(DATA_HOME)/bash-completion/completions
 
 build: check-go
 	mkdir -p $(BIN_DIR)
@@ -64,13 +69,13 @@ install-binary: clean-legacy-assets
 	mv -f "$(INSTALL_BIN)/$(BINARY).new" "$(INSTALL_BIN)/$(BINARY)"
 ifeq ($(UNAME_S),Linux)
 	# freedesktop shell completion (Linux only).
-	mkdir -p "$(HOME)/.local/share/bash-completion/completions"
-	cp $(COMPLETIONS_DIR)/enclave "$(HOME)/.local/share/bash-completion/completions/enclave"
-	mkdir -p "$(HOME)/.local/share/zsh/site-functions"
-	cp $(COMPLETIONS_DIR)/_enclave "$(HOME)/.local/share/zsh/site-functions/_enclave"
+	mkdir -p "$(BASH_COMPLETION_DIR)"
+	cp $(COMPLETIONS_DIR)/enclave "$(BASH_COMPLETION_DIR)/enclave"
+	mkdir -p "$(ZSH_COMPLETION_DIR)"
+	cp $(COMPLETIONS_DIR)/_enclave "$(ZSH_COMPLETION_DIR)/_enclave"
 	@# zsh has no user-level directory on its default fpath, so a completion
 	@# installed under the home directory is only found once ~/.zshrc adds it.
-	@command -v zsh >/dev/null 2>&1 && echo 'zsh completion installed; it is only picked up if ~/.zshrc has "fpath=(~/.local/share/zsh/site-functions $$fpath)" before compinit' || true
+	@command -v zsh >/dev/null 2>&1 && echo 'zsh completion installed; it is only picked up if ~/.zshrc has "fpath=($(ZSH_COMPLETION_DIR) $$fpath)" before whatever runs compinit' || true
 endif
 	@echo "Installed $(INSTALL_BINARY_LABEL) to $(INSTALL_BIN)/$(BINARY)"
 
@@ -80,8 +85,9 @@ clean-legacy-assets:
 uninstall:
 	rm -f "$(INSTALL_BIN)/$(BINARY)"
 ifeq ($(UNAME_S),Linux)
-	rm -f "$(HOME)/.local/share/bash-completion/completions/enclave"
-	rm -f "$(HOME)/.local/share/zsh/site-functions/_enclave"
+	rm -f "$(BASH_COMPLETION_DIR)/enclave"
+	rm -f "$(ZSH_COMPLETION_DIR)/_enclave"
+	@command -v zsh >/dev/null 2>&1 && echo 'zsh completion removed; the matching "fpath=($(ZSH_COMPLETION_DIR) ...)" line in ~/.zshrc is now unused' || true
 endif
 	@echo "Uninstalled $(BINARY)"
 
--- a/Makefile
+++ b/Makefile
@@ -37,6 +37,7 @@ check-go:
 completions:
 	mkdir -p $(COMPLETIONS_DIR)
 	$(BIN_DIR)/$(BINARY) completion bash > $(COMPLETIONS_DIR)/enclave
+	$(BIN_DIR)/$(BINARY) completion zsh > $(COMPLETIONS_DIR)/_enclave
 
 # make install now installs only the self-contained binary. This fixed path is
 # used solely for conservative cleanup of assets staged by older source
@@ -65,6 +66,11 @@ ifeq ($(UNAME_S),Linux)
 	# freedesktop shell completion (Linux only).
 	mkdir -p "$(HOME)/.local/share/bash-completion/completions"
 	cp $(COMPLETIONS_DIR)/enclave "$(HOME)/.local/share/bash-completion/completions/enclave"
+	mkdir -p "$(HOME)/.local/share/zsh/site-functions"
+	cp $(COMPLETIONS_DIR)/_enclave "$(HOME)/.local/share/zsh/site-functions/_enclave"
+	@# zsh has no user-level directory on its default fpath, so a completion
+	@# installed under the home directory is only found once ~/.zshrc adds it.
+	@command -v zsh >/dev/null 2>&1 && echo 'zsh completion installed; it is only picked up if ~/.zshrc has "fpath=(~/.local/share/zsh/site-functions $$fpath)" before compinit' || true
 endif
 	@echo "Installed $(INSTALL_BINARY_LABEL) to $(INSTALL_BIN)/$(BINARY)"
 
@@ -75,6 +81,7 @@ uninstall:
 	rm -f "$(INSTALL_BIN)/$(BINARY)"
 ifeq ($(UNAME_S),Linux)
 	rm -f "$(HOME)/.local/share/bash-completion/completions/enclave"
+	rm -f "$(HOME)/.local/share/zsh/site-functions/_enclave"
 endif
 	@echo "Uninstalled $(BINARY)"
 
--- a/README.md
+++ b/README.md
@@ -133,6 +133,20 @@ is on your `PATH`. On macOS the default is `/usr/local/bin/`; copying there may
 need `sudo` or a writable `/usr/local/bin`. Override the destination with
 `make install INSTALL_BIN=...`.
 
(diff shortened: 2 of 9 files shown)


scripts/verify-package-assets.sh
```diff
@@ -29,6 +29,13 @@ diff -qr "$repo_root/docs" "$app_root/docs"
 [ -f "$package_root/usr/share/doc/enclave/LICENSE.md" ]
 [ -f "$package_root/usr/share/doc/enclave/NOTICE.md" ]
 
+# Completions are only found if they sit in a directory the shell searches by
+# default, and that directory differs per distro: Debian uses
+# zsh/vendor-completions, Fedora zsh/site-functions.
+[ -f "$package_root/usr/share/bash-completion/completions/enclave" ]
+[ -f "$package_root/usr/share/zsh/vendor-completions/_enclave" ] ||
+    [ -f "$package_root/usr/share/zsh/site-functions/_enclave" ]
```
EclipseSourceAI: The `||` means neither path is actually pinned: a `.deb` that installed into `site-functions` (which is not on Debian's `fpath`) passes, and an `.rpm` into `vendor-completions` (dead on Fedora) also passes. Since the script already takes the package root as `$1`, an optional second arg for the expected zsh dir, passed from the two workflows, would make this catch what the comment says it catches.

Makefile
```diff
@@ -64,6 +65,11 @@ ifeq ($(UNAME_S),Linux)
 	# freedesktop shell completion (Linux only).
 	mkdir -p "$(HOME)/.local/share/bash-completion/completions"
 	cp $(COMPLETIONS_DIR)/enclave "$(HOME)/.local/share/bash-completion/completions/enclave"
+	mkdir -p "$(HOME)/.local/share/zsh/site-functions"
+	cp $(COMPLETIONS_DIR)/_enclave "$(HOME)/.local/share/zsh/site-functions/_enclave"
```
EclipseSourceAI: This hardcodes `$(HOME)/.local/share`, but `clean-state.sh` cleans `$XDG_DATA_HOME` too ([link](https://github.com/eclipse-enclave/enclave/blob/5e808201d3d2284e17c075f1cff4dcb4c214074c/scripts/clean-state.sh#L284-L287)), so with `XDG_DATA_HOME` set the two disagree. Same for the bash line above ([link](https://github.com/eclipse-enclave/enclave/blob/5e808201d3d2284e17c075f1cff4dcb4c214074c/Makefile#L64-L66)), so if you want this fixed it's one `DATA_HOME ?= $(or $(XDG_DATA_HOME),$(HOME)/.local/share)` variable for both.

Makefile
```diff
@@ -74,6 +80,7 @@ uninstall:
 	rm -f "$(INSTALL_BIN)/$(BINARY)"
 ifeq ($(UNAME_S),Linux)
 	rm -f "$(HOME)/.local/share/bash-completion/completions/enclave"
+	rm -f "$(HOME)/.local/share/zsh/site-functions/_enclave"
```
EclipseSourceAI: `install` tells the user to add an `fpath` line to `~/.zshrc`, and `uninstall` deletes the file without mentioning that line, leaving a stale entry pointing at a now-empty directory. A matching one-line hint here would close the loop.

README.md
```diff
@@ -97,6 +97,20 @@ is on your `PATH`. On macOS the default is `/usr/local/bin/`; copying there may
...
 `make install INSTALL_BIN=...`.
 
+On Linux it also installs shell completions: bash into
+`~/.local/share/bash-completion/completions/`, which is picked up automatically,
+and zsh into `~/.local/share/zsh/site-functions/`, which zsh only searches once
+you add it to `fpath` in `~/.zshrc`:
+
+```zsh
+fpath=(~/.local/share/zsh/site-functions $fpath)
+autoload -U compinit && compinit
```
EclipseSourceAI: Anyone on oh-my-zsh or prezto already runs `compinit`, so pasting this adds a second call. Worth saying explicitly that the `fpath=` line has to come before whatever runs `compinit` (the Makefile hint says this, the README doesn't) and that the `autoload` line is only needed if nothing else already does it.

Makefile
```diff
@@ -36,6 +36,7 @@ build: check-go
 completions:
 	mkdir -p $(COMPLETIONS_DIR)
 	$(BIN_DIR)/$(BINARY) completion bash > $(COMPLETIONS_DIR)/enclave
+	$(BIN_DIR)/$(BINARY) completion zsh > $(COMPLETIONS_DIR)/_enclave
```
EclipseSourceAI, line 45: fish is now the only shell the CLI can generate for that nothing installs, and the README advertises `enclave completion fish` right next to the install docs. `~/.config/fish/completions/` and `/usr/share/fish/vendor_completions.d/` are both on fish's default path with no user setup, so it's simpler than the zsh case. Fine as a follow-up, up to the maintainer whether it belongs here.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Installs the already-generated zsh completion script across `make install`, the `.deb`, and the `.rpm`, plus cleanup and packaging-verification coverage. The distro path split (`vendor-completions` on Debian, `site-functions` on Fedora) is correct: Debian's default `fpath` carries the former and deliberately omits `/usr/share/zsh/site-functions`.

Scope is tight and the tricky part was already handled on main: `writeCompletionWithHeader` keeps `#compdef` on line 1 so `compinit` still registers the file, with a test guarding it. `make build`, `make test`, and `make lint` (shellcheck plus the license-header check) all pass here, and I confirmed `make install-binary` / `make uninstall` place and remove both completions, with the zsh hint silent and exit 0 when no `zsh` is on `PATH`.

Worth a look: the new assertion in `scripts/verify-package-assets.sh` accepts either zsh layout, so it does not actually pin the per-distro path the way the PR description suggests. The rest are minor (hardcoded `~/.local/share` vs `XDG_DATA_HOME`, README `compinit` guidance, fish parity).

scripts/verify-package-assets.sh
planger: Correct, the `||` made the assertion vacuous. `verify-package-assets.sh` now takes the expected zsh directory as a second required argument (not optional, so a missing value fails loudly instead of silently skipping the check), and the workflows pass `usr/share/zsh/vendor-completions` for the `.deb` and `usr/share/zsh/site-functions` for both `.rpm` jobs. Verified against synthetic package roots: each layout passes only against its own expected directory and is rejected against the other, where all four combinations passed before.

Makefile
planger: Confirmed, and the bash line had the same problem. Added `DATA_HOME ?= $(or $(XDG_DATA_HOME),$(HOME)/.local/share)` with `BASH_COMPLETION_DIR` and `ZSH_COMPLETION_DIR` derived from it, used by both install and uninstall, and the printed hint now shows the resolved path. An install/uninstall round trip with `XDG_DATA_HOME=/tmp/xdgtest/share` puts both completions under that root and removes them again; with it unset the paths still resolve to `~/.local/share`. `LEGACY_INSTALL_DIR` stays hardcoded on purpose, since it only cleans up assets staged by older source installs.

Makefile
planger: Fair, the asymmetry was there. `uninstall` now prints a matching one-liner noting the `fpath` entry is unused. A stale entry pointing at a missing directory is harmless to zsh, so this is a tidiness hint rather than a fix, but it costs one line and closes the loop.

README.md
planger: Right, the snippet assumed nothing else calls `compinit`. The README now keeps only the `fpath` line in the code block, states it has to come before whatever runs `compinit`, notes that oh-my-zsh and prezto call it themselves, and makes `autoload -U compinit && compinit` conditional on a plain setup that does not. Also documents that both paths honor `XDG_DATA_HOME`.

Makefile
planger: Leaving this out. The PR is scoped to zsh, which needed the per-distro `fpath` split and the user-side `fpath` setup; fish is a separate change with its own paths and no such subtleties, and bundling it would widen a packaging PR that is otherwise easy to review. Worth a follow-up issue, which is the maintainer's call.

sdirix (commented):
I used this for a while in my own enclave build. This is how it looks like for me:

Image

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Only one commit landed since my last pass (`d431ec4`), addressing every point from that review.

- `verify-package-assets.sh` now takes the expected zsh dir as a required arg, and both workflows pass the right one (`vendor-completions` for `.deb`, `site-functions` for both `.rpm` jobs). Checked against `debian/rules` and `packaging/rpm/enclave.spec`, the install paths match what the workflows now assert.
- `Makefile` derives `DATA_HOME` from `XDG_DATA_HOME` with a fallback, and `BASH_COMPLETION_DIR`/`ZSH_COMPLETION_DIR` are used consistently in install/uninstall/hint. Matches `scripts/clean-state.sh`'s handling of the same variable.
- `uninstall` now prints the matching fpath-is-now-unused hint.
- README keeps only the `fpath` line in the snippet, states it must come before `compinit`, and calls out oh-my-zsh/prezto.

Fish completion install was intentionally left out as a separate, simpler follow-up, which is a reasonable scope call and not something to push on further.

No new issues found in the delta. sdirix's screenshot from real-world use looks like a straightforward confirmation, nothing to act on there.

- [pin the packaging CI check to each distro's actual zsh completion dir instead of accepting either](https://github.com/eclipse-enclave/enclave/pull/38#discussion_r3714198615)
- [make `install`/`uninstall` honor `XDG_DATA_HOME` instead of hardcoding `~/.local/share`](https://github.com/eclipse-enclave/enclave/pull/38#discussion_r3714198720)
- [have `uninstall` print a matching hint about the now-stale `fpath` entry](https://github.com/eclipse-enclave/enclave/pull/38#discussion_r3714198862)
- [clarify in the README that the `fpath` line must precede `compinit`, and that frameworks already call it](https://github.com/eclipse-enclave/enclave/pull/38#discussion_r3714199014)

xai:
Code looks good, just needs a rebase before we can merge!

xai (approved):
LGTM! Thanks for testing this in a real setup, @sdirix!

Issue #41, opened 2026-08-10
Deleting disposable cache data can break Enclave sessions


kalyi:
### Bug Description:

Enclave's cache is disposable, so deleting it should affect performance only.
Instead, cache deletion can affect functional correctness in two ways:

1. A container may fail to start because a newly recreated package-cache bind
   source is reported as missing, even though it exists on the host. An
   immediate retry succeeds.
2. If the extracted asset cache is deleted while a session is running,
   post-exit auth reconciliation may fail because `auth-reconcile.sh` is no
   longer available.

The startup failure was observed with Docker Desktop on macOS:

```text
invalid mount config for type bind: bind source path does not exist:
/host_mnt/Users/<user>/Library/Caches/org.eclipse.enclave/<tool>/<hash>/npm
```

#### Container startup

1. Delete the Enclave cache root. On macOS:

   ```sh
   rm -rf "$HOME/Library/Caches/org.eclipse.enclave"
   ```

2. Start Enclave from a project directory.
3. On an affected system, startup fails with a missing bind-source error even
   though the reported directory exists on the host. Rerunning Enclave without
   deleting anything succeeds.

#### Auth finalization

1. Start a persistent session with shared auth using an installation whose
   embedded assets are extracted into the cache.
2. Delete the Enclave cache after the container has started.
3. Exit the session. Post-exit reconciliation reports that
   `runtime-assets/auth-reconcile.sh` no longer exists.

### Additional Information

- Operating System: macOS Tahoe 26.6.1
- Enclave Version: main

PR #42, opened 2026-08-11
Fix cache bind mounts

kalyi:
- Keep cache deletion from breaking session startup by using source-creating host bind mounts for disposable cache directories while required mounts remain strict.
- Remove remaining runtime dependencies on disposable cache data, including auth reconciliation and the optional nvm cache.
- Surface gateway build output to make failures diagnosable.

#### What it does

Fixes issue [#41](https://github.com/eclipse-enclave/enclave/issues/41)

#### How to test

See issue [#41](https://github.com/eclipse-enclave/enclave/issues/41)

#### Breaking changes

#### Review checklist

### DIFF
--- a/internal/backend/qemu/bundle.go
+++ b/internal/backend/qemu/bundle.go
@@ -210,6 +210,11 @@ func (b *Backend) buildRuntimeMounts(req backend.Request, controlDir string, fil
 		if mount.Type != "" && mount.Type != backend.MountTypeBind {
 			return nil, fmt.Errorf("qemu backend: unsupported mount type %q", mount.Type)
 		}
+		if mount.CreateSourceDir {
+			if err := os.MkdirAll(mount.Source, 0o700); err != nil {
+				return nil, fmt.Errorf("qemu backend: create disposable mount source %q: %w", mount.Source, err)
+			}
+		}
 		info, err := os.Stat(mount.Source)
 		if err != nil {
 			return nil, fmt.Errorf("qemu backend: inspect mount source %q: %w", mount.Source, err)
--- a/internal/backend/types.go
+++ b/internal/backend/types.go
@@ -85,6 +85,12 @@ type Mount struct {
 	Source        string
 	ContainerPath string
 	ReadOnly      bool
+	// CreateSourceDir marks a disposable directory bind mount (package caches):
+	// cache data whose loss must never prevent a session from starting. Backends
+	// create a missing source as an empty directory instead of failing. Required
+	// project, config, auth, and file mounts must not set it, so their missing
+	// sources stay hard errors.
+	CreateSourceDir bool
 }
 
 type StoreKind string
--- a/internal/gateway/gateway.go
+++ b/internal/gateway/gateway.go
@@ -10,7 +10,6 @@ package gateway
 import (
 	"context"
 	"fmt"
-	"io"
 	"io/fs"
 	"os"
 	"path/filepath"
@@ -187,11 +186,14 @@ func buildGatewayImage(paths model.Paths, profile model.Profile, allowlistPath s
 			model.GatewayLabelAgent: profile.Name,
 		},
 	}
-	if err := docker.Build(context.Background(), req, io.Discard); err != nil {
+	// Build output goes to the terminal like the runtime image build, so a
+	// gateway build failure is attributable instead of surfacing later as an
+	// unrelated container-start error.
+	if err := docker.Build(context.Background(), req, os.Stdout); err != nil {
 		// Some Docker BuildKit setups fail DNS resolution in the default build
 		// network for Alpine index fetches. Retry once with host build network.
 		req.NetworkMode = "host"
-		if retryErr := docker.Build(context.Background(), req, io.Discard); retryErr != nil {
+		if retryErr := docker.Build(context.Background(), req, os.Stdout); retryErr != nil {
 			return fmt.Errorf("failed to build gateway image: %w (retry with host build network failed: %v)", err, retryErr)
 		}
 		logx.Warnf("Gateway build failed on default build network; retry with host build network succeeded")
--- a/internal/runtime/runtime.go
+++ b/internal/runtime/runtime.go
@@ -300,7 +300,9 @@ func (r *Runtime) prepareMounts() (*mountAccumulator, error) {
 	r.addSSHMount(mountArgs)
 	r.addImageInboxMount(mountArgs)
 	r.addSessionMonitorEnv(mountArgs)
-	r.addCacheMounts(mountArgs)
+	if err := r.addCacheMounts(mountArgs); err != nil {
+		return nil, err
+	}
 	r.addHistoryMounts(mountArgs)
 	r.addMemoryMounts(mountArgs)
 	r.addToolConfigMounts(mountArgs)
@@ -1063,7 +1065,9 @@ func (r *Runtime) addImageInboxMount(mounts *mountAccumulator) {
 		logx.Warnf("Failed to create image inbox directory %s: %v", inboxDir, err)
 		return
 	}
-	mounts.AddMount(bindMount(inboxDir, model.ContainerImageInboxDir, true))
+	// The inbox lives under the disposable cache root; an empty recreated inbox
+	// must not block a session start.
+	mounts.AddMount(disposableDirMount(inboxDir, model.ContainerImageInboxDir, true))
 	mounts.AddEnv(model.EnvImageInbox, model.ContainerImageInboxDir)
 	logx.Infof("Host image inbox mounted read-only at %s", model.ContainerImageInboxDir)
 }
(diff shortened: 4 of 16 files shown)


xai:
@EclipseSourceAI

internal/runtime/runtime.go
```diff
@@ -1095,39 +1099,47 @@ func (r *Runtime) addSessionMonitorEnv(mounts *mountAccumulator) {
...
-	// Yarn cache
-	mounts.AddMount(bindMount(filepath.Join(cacheDir, "yarn"), r.containerHome+"/.cache/yarn", false))
-	// Bun cache
-	mounts.AddMount(bindMount(filepath.Join(cacheDir, "bun"), r.containerHome+"/.bun", false))
-	// nvm installed Node.js versions
-	mounts.AddMount(bindMount(filepath.Join(cacheDir, "nvm"), r.containerHome+"/.nvm/versions", false))
+	for _, entry := range packageCacheDirs(r.containerHome) {
+		source := filepath.Join(cacheDir, entry[0])
+		if err := os.MkdirAll(source, 0o700); err != nil {
+			return fmt.Errorf("create package cache directory %s: %w", source, err)
```
EclipseSourceAI, line 1138: A failed `MkdirAll` here now aborts the whole session start, which works against the invariant this PR establishes (losing cache data must never block a session) and against the sibling mount helpers, which warn and skip: [addImageInboxMount](https://github.com/eclipse-enclave/enclave/blob/6ab6e6fc4ca8c7381c021e8e683cef6e69010507/internal/runtime/runtime.go#L1057-L1065). The mount is marked `CreateSourceDir` anyway, so warning and skipping that one cache keeps a full or read-only cache filesystem from taking the run down with it.

internal/runtime/runtime.go
```diff
@@ -1095,39 +1099,47 @@ func (r *Runtime) addSessionMonitorEnv(mounts *mountAccumulator) {
 	mounts.AddEnv(model.EnvSessionMonitorUser, r.containerUser)
 }
 
-func (r *Runtime) addCacheMounts(mounts *mountAccumulator) {
+// packageCacheDirs maps each package-cache directory under the per-project
+// cache root to its in-container mount point (relative to the container home).
+func packageCacheDirs(containerHome string) [][2]string {
```
EclipseSourceAI, line 1104: `[][2]string` with `entry[0]`/`entry[1]` at the call site reads worse than the named-field table used for exactly this a few lines further down ([addToolConfigMounts](https://github.com/eclipse-enclave/enclave/blob/6ab6e6fc4ca8c7381c021e8e683cef6e69010507/internal/runtime/runtime.go#L1183-L1192)). A `[]struct{ cacheName, containerPath string }` would match it.

internal/backend/types.go
```diff
@@ -85,6 +85,12 @@ type Mount struct {
 	Source        string
 	ContainerPath string
 	ReadOnly      bool
+	// CreateSourceDir marks a disposable directory bind mount (package caches):
+	// cache data whose loss must never prevent a session from starting. Backends
+	// create a missing source as an empty directory instead of failing. Required
+	// project, config, auth, and file mounts must not set it, so their missing
+	// sources stay hard errors.
```
EclipseSourceAI, line 92: The "missing sources stay hard errors" guarantee only holds when SELinux is not enforcing. `applySELinuxMounts` moves every bind mount into `Binds` ([formatBind](https://github.com/eclipse-enclave/enclave/blob/6ab6e6fc4ca8c7381c021e8e683cef6e69010507/internal/docker/selinux_mounts.go#L47-L55)), and `buildRunArgs` renders those as `--volume` ([here](https://github.com/eclipse-enclave/enclave/blob/6ab6e6fc4ca8c7381c021e8e683cef6e69010507/internal/docker/cli.go#L204-L206)), so on those hosts required mounts are already source-creating. The wording should be scoped, or the strictness made explicit for required mounts.

internal/gateway/gateway.go
```diff
@@ -187,11 +186,14 @@ func buildGatewayImage(paths model.Paths, profile model.Profile, allowlistPath s
 			model.GatewayLabelAgent: profile.Name,
 		},
 	}
-	if err := docker.Build(context.Background(), req, io.Discard); err != nil {
+	// Build output goes to the terminal like the runtime image build, so a
+	// gateway build failure is attributable instead of surfacing later as an
+	// unrelated container-start error.
+	if err := docker.Build(context.Background(), req, os.Stdout); err != nil {
```
EclipseSourceAI, line 192: The gateway `BuildRequest` never sets `Progress`, so `--progress quiet` is honored for the runtime image build ([build.go](https://github.com/eclipse-enclave/enclave/blob/6ab6e6fc4ca8c7381c021e8e683cef6e69010507/internal/app/build.go#L550)) but silently ignored here now that the output is no longer discarded.

internal/gateway/gateway.go
```diff
@@ -187,11 +186,14 @@ func buildGatewayImage(paths model.Paths, profile model.Profile, allowlistPath s
...
 		// Some Docker BuildKit setups fail DNS resolution in the default build
 		// network for Alpine index fetches. Retry once with host build network.
 		req.NetworkMode = "host"
-		if retryErr := docker.Build(context.Background(), req, io.Discard); retryErr != nil {
+		if retryErr := docker.Build(context.Background(), req, os.Stdout); retryErr != nil {
```
EclipseSourceAI, line 196: On the common BuildKit-DNS path the user now gets a full failing build log followed by a successful one, with only a warning at the end explaining it. Buffering the first attempt and flushing it only if the retry also fails keeps the recovered case quiet without losing attribution for real failures.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Fixes #41 by making the disposable package caches survive a deleted cache root. `backend.Mount` gains `CreateSourceDir`, which the Docker backend renders as `--volume src:dst` instead of the strict `--mount type=bind` (the only CLI form that creates a missing source), and the QEMU backend handles with an explicit `MkdirAll`. Cache sources are now created host-side in a loop driven by `packageCacheDirs`, the image inbox is marked disposable too, the auth reconcile script is snapshotted into memory at backend construction instead of being bind-mounted from the extracted asset cache, and the node-dev entrypoint tolerates an unwritable nvm version cache.

The approach looks sound and the tests cover the interesting bits. I verified the claim in `docs/persistence.md` that no required runtime mount comes from the cache tree: skills, gateway config, TLS and history/memory mounts all resolve under the state root, so the scoping holds.

Main points for a maintainer:
- `addCacheMounts` turning a `MkdirAll` failure into a hard session-start error is the one place the code contradicts the invariant the PR is establishing, and it diverges from the warn-and-skip style of the neighbouring mount helpers.
- The `--volume` strictness argument in the `CreateSourceDir` doc comment does not hold on SELinux-enforcing hosts, where every bind is already rendered as `--volume`.
- The gateway build output change is unrelated to the cache fix and rides in the same commit; it also ignores `--progress` and now prints a full failed build log on the recovered retry path.

xai:
Related: I noticed that if someone uses the `ssh-init` command (which one should NOT), it places the keys in the cache location. I created #48 to track this and am working on a branch to fix it.

internal/backend/qemu/bundle.go
```diff
@@ -210,6 +210,11 @@ func (b *Backend) buildRuntimeMounts(req backend.Request, controlDir string, fil
 		if mount.Type != "" && mount.Type != backend.MountTypeBind {
 			return nil, fmt.Errorf("qemu backend: unsupported mount type %q", mount.Type)
 		}
+		if mount.CreateSourceDir {
+			if err := os.MkdirAll(mount.Source, 0o700); err != nil {
+				return nil, fmt.Errorf("qemu backend: create disposable mount source %q: %w", mount.Source, err)
+			}
+		}
```
xai, line 217: This MkdirAll never runs for the case it exists for. prepareGuestRuntime calls prepareFileMounts (line 162) before buildRuntimeMounts (line 167), and prepareFileMounts stats every bind-mount source and returns a hard error when one is missing (lines 272-275). A CreateSourceDir mount whose source is absent therefore fails with `qemu backend: inspect file mount source ...` before this block executes, and the directory is never created. Confirmed with a scratch test: prepareGuestRuntime with a single CreateSourceDir bind whose source does not exist returns that error and the source stays missing. The contract on backend.Mount.CreateSourceDir (internal/backend/types.go:88-93, 'Backends create a missing source as an empty directory instead of failing') is thus honored by the Docker backend only. Today the gap is masked because the runtime pre-creates every such source host-side (addCacheMounts, addImageInboxMount), leaving only the delete-between-MkdirAll-and-stat window exposed, but merging dead code that implies support invites a future change to drop the runtime-side creation and break qemu. Fix either way: (a) run the CreateSourceDir creation pass over req.Mounts at the top of prepareGuestRuntime, before prepareFileMounts, and add a qemu test asserting a missing disposable source is created rather than rejected; or (b) delete this block and reword the CreateSourceDir comment to state that the runtime pre-creates sources and Docker's --volume rendering only covers daemon-view lag.

xai (changes requested):
One item we should change, see inline comment.

Edit: also needs a rebase.

xai:
Ping @kalyi

xai:
@EclipseSourceAI

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Follow-up review. All three of my earlier findings on `internal/runtime/runtime.go` and `internal/backend/types.go` are now fixed: `addCacheMounts` warns and skips on a failed `MkdirAll` instead of hard-failing, the cache dir list is now a named `packageCache{cacheName, containerPath}` slice instead of `[][2]string`, and the `CreateSourceDir` doc comment on `backend.Mount` now scopes the "hard errors" guarantee to exclude SELinux-enforcing hosts.

The gateway build-output change is no longer part of this PR (the diff against `main` no longer touches `internal/gateway/gateway.go`), so my two comments there no longer apply to this branch; the underlying concern was independently fixed on `main` in `244773f` (build output is now buffered and only surfaced on failure, not streamed live), so I'm listing them as resolved too.

xai's requested fix on `internal/backend/qemu/bundle.go` (source creation now runs before `prepareFileMounts`, with a regression test) is also in place.

The new commit since then, `538cdfd` (auth reconcile script snapshotted at backend construction instead of bind-mounted), looks correct and has solid test coverage. Build and the touched test packages (`runtime`, `backend/...`, `docker`) pass.

- [addCacheMounts hard-failing on MkdirAll instead of warn-and-skip](https://github.com/eclipse-enclave/enclave/pull/42#discussion_r3757550114)
- [[][2]string cache list should be a named-field struct](https://github.com/eclipse-enclave/enclave/pull/42#discussion_r3757550231)
- [CreateSourceDir doc comment's strictness claim doesn't hold under SELinux](https://github.com/eclipse-enclave/enclave/pull/42#discussion_r3757550336)
- [gateway build ignores --progress on the runtime image build path](https://github.com/eclipse-enclave/enclave/pull/42#discussion_r3757551823)
- [gateway prints a full failing build log on the recovered retry path](https://github.com/eclipse-enclave/enclave/pull/42#discussion_r3757552439)

xai (approved):
Thanks for the fixes and the rebase, @kalyi!

Code looks good and I tested it successfully with both docker and podman 🎉

PR #45, opened 2026-08-11
fix: fail when Claude installation is missing

### LINKED ISSUE #43: Claude image build can succeed without Claude installed
### Bug Description:

The Claude installation step can complete successfully even when the native installer does not provide a usable claude executable. Docker then produces an image that fails at runtime with:

  exec: claude: not found

### Steps to Reproduce:

1. Build the Claude image in an environment where the native installer exits successfully without installing claude.
2. Observe that the image build still succeeds.
3. Start Enclave with Claude.
4. The container fails because claude is not found.
5. Rerunning Enclave, also with --rebuild, doesn't fix it (see [#44](https://github.com/eclipse-enclave/enclave/issues/44))

### Additional Information

- Operating System: macOS Tahoe 26.6.1
- Enclave Version: current main

kalyi:
#### What it does

Fixes [#43](https://github.com/eclipse-enclave/enclave/issues/43).

#### How to test

See [#43](https://github.com/eclipse-enclave/enclave/issues/43).

#### Breaking changes

#### Review checklist

### DIFF
--- a/extensions/tools/claude/install.sh
+++ b/extensions/tools/claude/install.sh
@@ -11,10 +11,13 @@ set -e
 
 curl -fsSL https://claude.ai/install.sh | bash
 
-if command -v claude >/dev/null 2>&1; then
-    echo "Claude installed at: $(which claude)"
-    claude --version
+if ! command -v claude >/dev/null 2>&1; then
+    echo "Claude Code install failed: claude binary not found" >&2
+    exit 1
 fi
+
+echo "Claude installed at: $(which claude)"
+claude --version
 
 # Install sandbox runtime for Claude Code sandbox support
 enclave-agent-npm-install @anthropic-ai/sandbox-runtime


xai:
@EclipseSourceAI

extensions/tools/claude/install.sh
```diff
@@ -11,10 +11,13 @@ set -e
 
 curl -fsSL https://claude.ai/install.sh | bash
```
EclipseSourceAI, line 12: `set -e` never catches a failure here: the pipeline's status is bash's, and bash fed an empty stdin exits 0. That's the actual reason a broken install looked successful. Adding `pipefail` on line 10 would surface the real curl error instead of only the generic "binary not found" downstream.

extensions/tools/claude/install.sh
```diff
@@ -11,10 +11,13 @@ set -e
...
 
-if command -v claude >/dev/null 2>&1; then
-    echo "Claude installed at: $(which claude)"
-    claude --version
+if ! command -v claude >/dev/null 2>&1; then
+    echo "Claude Code install failed: claude binary not found" >&2
+    exit 1
 fi
```
EclipseSourceAI, line 17: This verify-and-report block is the same logic `enclave-install-npm-tool` already has ([link](https://github.com/eclipse-enclave/enclave/blob/6ab6e6fc4ca8c7381c021e8e683cef6e69010507/runtime-assets/build-scripts/bin/enclave-install-npm-tool#L22-L27)), and mistral-vibe carries a third copy ([link](https://github.com/eclipse-enclave/enclave/blob/6ab6e6fc4ca8c7381c021e8e683cef6e69010507/extensions/tools/mistral-vibe/install.sh#L17-L22)). A small `enclave-verify-tool-binary <bin> <label>` next to the other build-script bins would cover all three.

extensions/tools/claude/install.sh
```diff
@@ -11,10 +11,13 @@ set -e
...
 
+echo "Claude installed at: $(which claude)"
```
EclipseSourceAI, line 19: Test uses `command -v` but the print uses `which`; `enclave-install-npm-tool` sticks to `command -v` for both ([link](https://github.com/eclipse-enclave/enclave/blob/6ab6e6fc4ca8c7381c021e8e683cef6e69010507/runtime-assets/build-scripts/bin/enclave-install-npm-tool#L22-L27)). Also the failure message says "Claude Code" while this one says "Claude".

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Small, well-targeted fix: the Claude tool install script now hard-fails when the `claude` binary is missing after the native installer runs, instead of silently skipping the version check and returning 0.

The approach is right and fits the existing build flow. `enclave-install-tool` only appends the tool to `.installed-tools` after `install.sh` exits 0, and `checkToolInstalled` reads that manifest at runtime, so previously a broken Claude install produced an image that claimed the tool was present. It also mirrors what `enclave-install-npm-tool` and the mistral-vibe script already do, and `enclave-install-tool` exports `$HOME/.local/bin` into PATH before running the script, so the `command -v` check sees the native installer's output on both the Docker and the Alpine microvm paths. Scope is clean, no drive-by changes, shellcheck passes.

Points worth a look: the `curl | bash` pipeline still swallows a curl failure without `pipefail`, so the build fails with a generic message rather than the real cause, and the verify-and-report block is now the third copy of the same logic in the repo. Both are in inline comments.

extensions/tools/claude/install.sh
xai: usually i don't find that abstractions like that are super great for very short stuff. in this case, it might help future tools to "find" the verification logic and use it. not sure, your decision.

PR #46, opened 2026-08-11
fix: refresh tool layers on forced rebuild

### LINKED ISSUE #44: --rebuild can reuse cached tool installation layers
### Bug Description:

--rebuild starts an image build but does not force the selected tool’s installation layer to refresh. BuildKit can therefore reuse a cached tool layer, preserving an outdated or broken installation.

### Steps to Reproduce:

1. Build a tool image and populate the BuildKit cache.
2. Make a newer tool version available without changing local build inputs, or retain a cached broken installation layer (see, e.g., [#43](https://github.com/eclipse-enclave/enclave/issues/43)).
3. Run Enclave with --rebuild.
4. Observe that the tool installation step is reported as cached and the previous installation remains in the rebuilt image.

### Additional Information

- Operating System: macOS Tahoe 26.6.1
- Enclave Version: current main

kalyi:
#### What it does

Fixes [#44](https://github.com/eclipse-enclave/enclave/issues/44).

#### How to test

See [#44](https://github.com/eclipse-enclave/enclave/issues/44).

#### Breaking changes

#### Review checklist

### DIFF
--- a/internal/app/build.go
+++ b/internal/app/build.go
@@ -816,7 +816,7 @@ func devcontainerFeatureSelectionMessage(features string) string {
 }
 
 // planAgentUpdatesForTools decides, per tool, whether the agent CLI install
-// must be refreshed. With forceAll set (the `update` command), every tool is
+// must be refreshed. With forceAll set (`update` or `--rebuild`), every tool is
 // force-updated; otherwise each tool is refreshed only when its update interval
 // has elapsed and an online probe reports a changed upstream fingerprint.
 func planAgentUpdatesForTools(forceAll bool, tools []string, home string, now time.Time, resolveAutomatic automaticToolUpdateResolver) (agentUpdatePlan, error) {
--- a/internal/app/build_microvm.go
+++ b/internal/app/build_microvm.go
@@ -57,7 +57,7 @@ func ensureQEMUBundle(input *CommandInput, opts model.Options, buildCfg buildCon
 		}
 		return buildCfg, 0
 	}
-	plan, err := resolveRuntimeImageBuildPlan(input.Ctx.Paths, buildCfg, opts.BuildOptions, opts.Tool, host.Home, false, time.Now().UTC())
+	plan, err := resolveRuntimeImageBuildPlan(input.Ctx.Paths, buildCfg, opts.BuildOptions, opts.Tool, host.Home, opts.ForceRebuild, time.Now().UTC())
 	if err != nil {
 		logx.Errorf("%v", err)
 		return buildConfig{}, 1
--- a/internal/app/command_run.go
+++ b/internal/app/command_run.go
@@ -265,9 +265,10 @@ func ensureRuntimeImage(input *CommandInput, opts model.Options, buildCfg *build
 		}
 		return resolved, 0
 	}
-	// Normal runs never force an agent update; the `update` command is the
-	// explicit path. Automatic interval-based refresh still applies below.
-	buildPlan, err := resolveRuntimeImageBuildPlan(input.Ctx.Paths, resolved, opts.BuildOptions, opts.Tool, host.Home, false, time.Now().UTC())
+	// A forced image rebuild must also invalidate the selected tool install
+	// layers; otherwise BuildKit can reproduce a known-bad image entirely from
+	// cache. Automatic interval-based refresh still applies to normal runs.
+	buildPlan, err := resolveRuntimeImageBuildPlan(input.Ctx.Paths, resolved, opts.BuildOptions, opts.Tool, host.Home, opts.ForceRebuild, time.Now().UTC())
 	if err != nil {
 		logx.Errorf("%v", err)
 		return buildConfig{}, 1


xai:
@EclipseSourceAI

internal/app/command_run.go
```diff
@@ -265,9 +265,10 @@ func ensureRuntimeImage(input *CommandInput, opts model.Options, buildCfg *build
 		}
 		return resolved, 0
 	}
-	// Normal runs never force an agent update; the `update` command is the
-	// explicit path. Automatic interval-based refresh still applies below.
-	buildPlan, err := resolveRuntimeImageBuildPlan(input.Ctx.Paths, resolved, opts.BuildOptions, opts.Tool, host.Home, false, time.Now().UTC())
+	// A forced image rebuild must also invalidate the selected tool install
+	// layers; otherwise BuildKit can reproduce a known-bad image entirely from
+	// cache. Automatic interval-based refresh still applies to normal runs.
+	buildPlan, err := resolveRuntimeImageBuildPlan(input.Ctx.Paths, resolved, opts.BuildOptions, opts.Tool, host.Home, opts.ForceRebuild, time.Now().UTC())
```
EclipseSourceAI, line 271: `forceAll` does two things: it bumps the per-tool stamp (which is what actually busts the cached tool layer) and it sets the force flag, which flips npm to `prefer-online` with no offline retry ([enclave-install-tool L56-L62](https://github.com/eclipse-enclave/enclave/blob/6ab6e6fc4ca8c7381c021e8e683cef6e69010507/runtime-assets/build-scripts/bin/enclave-install-tool#L56-L62)). So `--rebuild` now requires registry access and can bump agent CLI versions, which is a bigger promise than "rebuild the image". If only the layer refresh is wanted, a fresh stamp alone does that; a maintainer should decide which semantic `--rebuild` should have.

internal/app/command_run.go
```diff
@@ -265,9 +265,10 @@ func ensureRuntimeImage(input *CommandInput, opts model.Options, buildCfg *build
 		}
 		return resolved, 0
 	}
-	// Normal runs never force an agent update; the `update` command is the
-	// explicit path. Automatic interval-based refresh still applies below.
-	buildPlan, err := resolveRuntimeImageBuildPlan(input.Ctx.Paths, resolved, opts.BuildOptions, opts.Tool, host.Home, false, time.Now().UTC())
+	// A forced image rebuild must also invalidate the selected tool install
+	// layers; otherwise BuildKit can reproduce a known-bad image entirely from
```
EclipseSourceAI, line 269: This makes `--rebuild` behave like `update` for agent CLIs, including resetting the auto-update interval via the post-build stamp commit. The docs still describe `update` as the way to refresh an agent now ([tools.md L74-L86](https://github.com/eclipse-enclave/enclave/blob/6ab6e6fc4ca8c7381c021e8e683cef6e69010507/docs/tools.md#L74-L86)) and DEV.md recommends `--rebuild` as the plain asset-refresh flag ([DEV.md L96-L102](https://github.com/eclipse-enclave/enclave/blob/6ab6e6fc4ca8c7381c021e8e683cef6e69010507/docs/DEV.md#L96-L102)), so both need an update per the docs rule in AGENTS.md.

internal/app/build_microvm.go
```diff
@@ -57,7 +57,7 @@ func ensureQEMUBundle(input *CommandInput, opts model.Options, buildCfg buildCon
 		}
 		return buildCfg, 0
 	}
-	plan, err := resolveRuntimeImageBuildPlan(input.Ctx.Paths, buildCfg, opts.BuildOptions, opts.Tool, host.Home, false, time.Now().UTC())
+	plan, err := resolveRuntimeImageBuildPlan(input.Ctx.Paths, buildCfg, opts.BuildOptions, opts.Tool, host.Home, opts.ForceRebuild, time.Now().UTC())
```
EclipseSourceAI, line 60: The qemu bundle build never sees the plan: `buildQEMUBundle` takes no `agentUpdatePlan` ([build_microvm.go L230](https://github.com/eclipse-enclave/enclave/blob/6ab6e6fc4ca8c7381c021e8e683cef6e69010507/internal/app/build_microvm.go#L230)) and the script calls `enclave-install-tool` without the force arg ([build-bundle.sh L139](https://github.com/eclipse-enclave/enclave/blob/6ab6e6fc4ca8c7381c021e8e683cef6e69010507/runtime-assets/microvm/alpine/build-bundle.sh#L139)). The only observable effect is that `--rebuild` now resets the agent update stamps for a bundle whose install was not actually forced, and `plan.AgentUpdates.NeedsRebuild` in the gate below becomes redundant with `opts.ForceRebuild`. Intended, or should this call site keep passing `false`?

internal/app/build.go
```diff
@@ -816,7 +816,7 @@ func devcontainerFeatureSelectionMessage(features string) string {
 }
 
 // planAgentUpdatesForTools decides, per tool, whether the agent CLI install
-// must be refreshed. With forceAll set (the `update` command), every tool is
+// must be refreshed. With forceAll set (`update` or `--rebuild`), every tool is
```
EclipseSourceAI, line 819: A forced update writes a new stamp but leaves the stored fingerprint untouched (backfill only fills in missing ones, [build.go L997-L1008](https://github.com/eclipse-enclave/enclave/blob/6ab6e6fc4ca8c7381c021e8e683cef6e69010507/internal/app/build.go#L997-L1008)), so roughly an interval after each forced build the probe still reports "changed" and triggers another forced rebuild. That was contained to `enclave update` before; with `--rebuild` on the same path it will hit far more often, so it is worth a follow-up.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Small, focused fix for #44: `--rebuild` now feeds `opts.ForceRebuild` into `resolveRuntimeImageBuildPlan` as `forceAll`, so the per-tool agent update plan bumps stamps and sets the force flag instead of letting BuildKit serve the tool install layers from cache. Same wiring is applied to the qemu bundle path, plus a doc-comment update. Build and `go test ./internal/app/...` pass locally.

The mechanics are right, but reusing `forceAll` couples two things: busting the tool layer (stamp change) and forcing an online npm resolve with no offline retry. That makes `--rebuild` network-dependent and version-bumping, and it resets the auto-update interval, which is close to what `update` does. That semantic call is the main thing to look at, along with the qemu call site, where the plan never reaches the bundle build and the only effect is the stamp commit.

Two smaller points: the docs (`docs/tools.md`, `docs/DEV.md`) still describe the old split between `--rebuild` and `update`, and there is no test covering the new wiring even though `planAgentUpdatesForTools` itself is well covered, so a positional bool flip here would go unnoticed.

internal/app/command_run.go
xai: I would still use `update` to refresh an agent, so nothing wrong here.

internal/app/command_run.go
xai: Yes, `--rebuild` fetches from the registry instead of trying to resolve `@latest` from a possible stale cache. There is a `--no-rebuild` switch anyway in case someone doesn't want that behavior.

internal/app/build_microvm.go
xai: For qemu, every bundle build is already a fresh online install, so that's fine.

xai (approved):
LGTM 👍

PR #47, opened 2026-08-11
fix: make build the default target

kalyi:
#### What it does

Makes `build` the default Make target. Currently, plain `make` runs only the silent Go-version check and produces no binary or output, which is confusing. After this change, make behaves like make build.

#### How to test

`make`

Verify that it runs the Go version check and builds bin/enclave, matching:

`make build`

#### Follow-ups

none

#### Breaking changes

#### Review checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,7 @@
 # SPDX-License-Identifier: MIT
 
 .PHONY: build cross-build install install-binary clean-legacy-assets uninstall clean clean-lint-cache test deb deb-quick rpm lint lint-changed lint-report fmt vet generate lint-tools check-go completions check-license-headers
+.DEFAULT_GOAL := build
 
 BIN_DIR ?= ./bin
 BINARY ?= enclave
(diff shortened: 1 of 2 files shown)


sdirix:
@EclipseSourceAI

Makefile
```diff
@@ -21,6 +21,10 @@ LINT_GO_CACHE := $(LINT_TMP_DIR)/go-build
 LINT_GOLANGCI_CACHE := $(LINT_TMP_DIR)/golangci-lint-cache
 BASE_REF ?=
 
+build: check-go
```
EclipseSourceAI, line 25: Works, but it makes the default goal depend on target ordering, which silently breaks the next time someone adds a rule above `build`. `.DEFAULT_GOAL := build` next to the [`.PHONY` line](https://github.com/eclipse-enclave/enclave/blob/6ab6e6fc4ca8c7381c021e8e683cef6e69010507/Makefile#L8-L8) gets the same behavior explicitly and keeps the diff to one added line.

Makefile
EclipseSourceAI, line 1: Plain `make` is now a real entry point, but [DEV.md](https://github.com/eclipse-enclave/enclave/blob/6ab6e6fc4ca8c7381c021e8e683cef6e69010507/docs/DEV.md#L15-L19) only documents `make build`. A one-liner there would help, since the point of the change is discoverability.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Moves the `build` rule above `check-go` in the `Makefile` so it becomes GNU Make's default goal (previously plain `make` only ran the silent Go version check). Verified with `make -n`: the default goal is now `build` and the recipe matches `make build`.

No side effects on the packaging or CI paths: `debian/rules` overrides `dh_auto_build` with its own `go build`, the RPM spec builds directly in `%build`, and no workflow invokes bare `make`.

Two points for a maintainer: whether the implicit ordering-based default goal is preferred over `.DEFAULT_GOAL := build`, and whether `docs/DEV.md` should mention the new default.

Makefile
kalyi: Added.

xai (approved):
LGTM 👍

PR #52, opened 2026-08-14
feat: add enclave network log

planger:
#### What it does

Adds `enclave network log`, a reader for the gateway audit log, and fixes the log format it reads.

The gateway already wrote audit events, but nothing could read them and the format had three defects a reader would otherwise have to work around: raw dnsmasq text appended by a shell tail, no session identity in a file shared by concurrent sessions, and unbounded growth.

- `internal/netlog` is the single source of truth for the JSONL contract (event schema, append, scan, filter, aggregate, follow, rotate).
- `enclave network log` supports row, `--follow`, `--summary`, and `--json` modes, with `--since`, `--verdict`, `--domain`, `--type`, `--tool`, `--session`, and `--all-running` filters. Terminal output is the aligned human form; `--json` is the machine contract, emitting the raw JSONL event stream or, with `--summary`, the aggregate as a single object.
- DNS denials move out of the shell entrypoint into `internal/gateway/dnsaudit`, running as its own process so they are recorded even when the proxy is disabled. Written against captured dnsmasq 2.91 output, which showed the shell filter's NODATA case reported allowed domains as denied: dnsmasq returns NODATA-IPv6 for every AAAA lookup of an allowlisted host without an IPv6 record. NODATA is dropped; NXDOMAIN, SERVFAIL, and REFUSED are kept, distinguishing a policy blackhole from an upstream failure.
- Reads are scoped by the event session field, and gateway paths are deduplicated so a shared file is read once instead of reporting every event twice.
- Rotation copies and truncates instead of renaming, above a fixed `netlog.MaxLogBytes` of 32MB. The gateway bind-mounts the log as a single file, so a rename left a running session appending to `network.log.1`, where `--follow` never looks. Concurrent session starts serialize on a lock file.
- Line splitting is bounded, so a torn write cannot grow a long-lived follower's buffer without limit. The scanner reports the offset it stopped at and `--follow` resumes from there, closing the window where an event appended while the backlog printed was lost or shown twice.
- Docs: `--network-log coarse` records one event per TLS connection, not one per request, and successful DNS lookups are never recorded, so absence of events was easy to misread as absence of traffic. The website also claimed DNS queries were logged.

Also: normalize domains through `domainpattern` everywhere so the filter and the aggregate agree, and add shell completion for the filter flags.

#### On scope

The first version of this branch also shipped a configurable `network_log_max_size` and a tab-separated machine output form. Both were dropped in 6a6c15f as more machinery than the feature earned. The size option needed a size parser, a project-scope guardrail and generated option plumbing to express one number nobody had asked to change, so it is a constant now. The TSV form was a second consumer-facing column contract sitting next to `--json`, which is the documented one, so `--plain` is gone and `--summary --json` covers the machine-readable aggregate. That removed about 620 lines.

#### How to test

`make build` and `make test` pass. `make lint` was not run: `golangci-lint` is not installed in this environment. `go vet ./...` is clean.

Manually, in a project directory:

1. Start a session and generate traffic, including a denied domain:
   ```
   enclave run -- claude
   # inside: curl https://example.com ; curl https://blocked.example
   ```
2. `enclave network log` shows pass and deny rows for that project and tool, including after the session exits.
3. `enclave network log --summary` aggregates per domain; `--json` emits the raw JSONL and `--summary --json` a single aggregate object.
4. `enclave network log -f` in a second terminal streams new events live while the session runs.
5. Filters: `--verdict deny`, `--domain '*.example.com'`, `--type dns`, `--since 10m`, `--since session`.
6. Run two sessions of the same project concurrently and check `--session <container>` and `--all-running` scope correctly and do not duplicate events.
7. Request-level detail: run with `--network-log requests` and confirm per-request HTTP events appear.
8. Rotation: grow a project's `network.log` past 32MB (`truncate -s 33M` works, only the size is checked), start a session, and confirm the previous generation lands in `network.log.1` while a running session keeps appending to the same file and `--follow` keeps working.

#### Follow-ups

- `--project` scope for network mutation commands is still unsupported, unchanged by this PR.
- Coarse mode records one event per TLS connection; per-request visibility requires `--network-log requests`, which forces MITM for all allowlisted hosts. Documented under [Coverage and granularity](docs/networking.md#coverage-and-granularity).
- The rotation cap is a constant. If someone needs to tune it, adding the config option back is a contained change.

#### Breaking changes

The gateway audit log format changed (raw dnsmasq lines are gone, events carry session identity). The log was not readable before, so no consumer should exist, but it is a format change to a file on disk.

#### Review checklist


internal/util/size.go
```diff
@@ -0,0 +1,41 @@
...
+
+// Binary size units, so "4.2 KB" means 4300 bytes rather than 4200.
+const (
+	unitKB = 1024
+	unitMB = 1024 * unitKB
+	unitGB = 1024 * unitMB
+)
+
+// FormatBytes renders a byte count for human output.
+func FormatBytes(bytes int64) string {
```
EclipseSourceAI: This duplicates `formatBytes` in `internal/app/cleanup.go` ([link](https://github.com/eclipse-enclave/enclave/blob/eb9261a8bc2c306986bff30678ddf1812e39c2d1/internal/app/cleanup.go#L390-L406)), same units, same output for everything below 10 units. Since you are adding the shared one anyway, drop the `app` copy and point its callers here.

internal/netlog/aggregate.go
```diff
@@ -0,0 +1,96 @@
...
+
+	for _, event := range events {
+		if event.IsSessionMarker() {
+			continue
+		}
+		// The same normalization the --domain filter applies, so the two views of
+		// one log cannot disagree about what counts as the same host.
+		domain, err := domainpattern.NormalizeHost(event.Domain)
+		if err != nil {
+			continue
```
EclipseSourceAI: Events with no domain are dropped from the aggregate entirely, including their verdict counts. The proxy writes deny events with an empty domain (`tls-clienthello` and `invalid-host`), so `--summary` reports `total_deny: 0` for a log where `--verdict deny` prints rows. Either count them under a placeholder domain or at least fold them into the totals.

gateway-entrypoint.sh
```diff
@@ -520,19 +523,17 @@ start_dnsmasq() {
...
-    tail -n 0 -F "$DNSMASQ_LOG_FILE" | while IFS= read -r line; do
-        case "$line" in
-            *NXDOMAIN*|*NODATA*|*SERVFAIL*|*REFUSED*)
-                printf '%s\n' "$line" >> "$NETWORK_LOG_FILE"
-                ;;
-        esac
-    done &
-    DNSMASQ_TAIL_PID="$!"
+    log "Starting DNS audit translator"
+    enclave-gateway-proxy -dns-audit "$DNSMASQ_LOG_FILE" >>"$DNS_AUDIT_LOG_FILE" 2>&1 &
```
EclipseSourceAI: The translator runs as root while the two other long-lived children in this script are dropped to unprivileged users with `su-exec` ([proxy](https://github.com/eclipse-enclave/enclave/blob/eb9261a8bc2c306986bff30678ddf1812e39c2d1/gateway-entrypoint.sh#L508), [dnsmasq](https://github.com/eclipse-enclave/enclave/blob/eb9261a8bc2c306986bff30678ddf1812e39c2d1/gateway-entrypoint.sh#L516)). It only tails a world-readable log and appends to a 666 file, so `su-exec "$PROXY_USER:$PROXY_USER"` should work here too.

internal/app/network_log_cmd.go
```diff
@@ -0,0 +1,382 @@
...
+		if err != nil {
+			return networkLogScope{}, fmt.Errorf("resolve project context: %w", err)
+		}
+		return networkLogScope{
+			paths: []string{config.HostProjectNetworkLogPath(home, project.Hash, input.Options.Tool)},
+			label: currentProjectToolLabel(project.Hash, input.Options.Tool),
+		}, nil
+	}
+
+	if err := checkDocker(); err != nil {
```
EclipseSourceAI, line 146: `--session` needs Docker and a live gateway, so a session that has exited cannot be selected even though its events are on disk and carry the session name. The whole point of the default scope is that exited sessions stay readable. Falling back to `filter.Session = sessionName` against the current project's log when no gateway matches would close that gap.

internal/netlog/filter.go
```diff
@@ -0,0 +1,123 @@
...
+
+	normalized.Type = strings.ToLower(strings.TrimSpace(f.Type))
+	switch normalized.Type {
+	case "", TypeDNS, TypeHTTP, TypeTCP:
+	default:
+		return f, fmt.Errorf("invalid type %q (use %s, %s or %s)", f.Type, TypeDNS, TypeHTTP, TypeTCP)
+	}
+
+	if pattern := strings.TrimSpace(f.Domain); pattern != "" {
+		domain, err := domainpattern.Normalize(pattern)
```
EclipseSourceAI: `domainpattern.Normalize` enforces allowlist safety rules that make no sense for a read-only filter: `--domain '*.com'` is rejected with "wildcard suffix must include at least two labels" ([link](https://github.com/eclipse-enclave/enclave/blob/eb9261a8bc2c306986bff30678ddf1812e39c2d1/internal/domainpattern/pattern.go#L56)). Broadening a query is not a policy decision.

internal/app/network_log_cmd.go
```diff
@@ -0,0 +1,382 @@
...
+		// unreadable lines once, and a live stream must not be drowned out.
+		OnSkipped: func() { logx.Debugf("Skipped an unreadable network log line") },
+	})
+	err := follower.Run(ctx, func(event netlog.Event) {
+		if writeErr != nil || !filter.Match(event) {
+			return
+		}
+		if asJSON {
+			writeErr = writeNetworkLogJSON(out, event)
+		} else if _, writeErr = out.WriteString(netlog.RenderEvent(event, render)); writeErr == nil {
```
EclipseSourceAI: A session marker in follow mode renders without the blank-line separation and without the pass/deny counts that `WriteEvents` gives it, so the boundary looks different depending on whether it came from the backlog or the live stream. Reusing the same separator logic for a marker here would keep one output format.

internal/gateway/gateway_proxy_build_inputs.txt
```diff
@@ -7,11 +7,13 @@ internal/appassets
 internal/config
 internal/domainpattern
 internal/gateway/bundle
+internal/gateway/dnsaudit
 internal/gateway/mitm
 internal/gateway/tlsstore
 internal/git
 internal/logx
 internal/model
+internal/netlog
```
EclipseSourceAI, line 16: The gateway proxy only needs `Event` and `Appender` (plus `Follower` for dnsaudit), but this pulls in `render.go`, `aggregate.go`, `filter.go` and `rotate.go` along with `internal/logx`. Splitting the reader side into its own package would keep the sidecar's build inputs and the embedded asset tree to what it actually uses.

docs/DEV.md
```diff
@@ -166,6 +166,11 @@ For CLI-only options (not configurable via config files), omit `DefaultsField`
 and use `Apply: ApplyNone` in `options_def.go` (for example:
 `--force-base-image` and `--no-rebuild`).
 
+For config-only options, omit `CLIFlags` instead (for example:
```
EclipseSourceAI, line 233: This paragraph is left over from the `network_log_max_size` option that 6a6c15f dropped, and has nothing to do with the network log reader. Better as its own docs commit.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Adds `enclave network log`, a reader for the gateway audit log, plus the format fixes that make it readable: a new `internal/netlog` package owning the JSONL contract (event schema, append, scan, filter, aggregate, follow, rotate), a `dnsaudit` translator that replaces the shell `tail | grep` with a real process, session stamping on every event, and copy-and-truncate rotation at 32 MB. The docs work correcting what `coarse` mode actually records is the most valuable part of the change and is accurate.

The design holds up well. Rotation by copy-and-truncate under a lock is the right call given the bind-mount, the scanner offset handed to `--follow` genuinely closes the lost/duplicated-event window, and dropping NODATA from the DNS translation fixes a real false positive in the old shell filter. `go build ./...` and the tests for the touched packages pass here.

- `Aggregate` drops events whose domain fails to normalize, which silently removes the proxy's domainless deny events (`tls-clienthello`, `invalid-host`) from `--summary` totals. `--summary` and `--verdict deny` then disagree about how much was blocked.
- The DNS audit translator is started as root in `gateway-entrypoint.sh` while dnsmasq and the proxy next to it are dropped to unprivileged users.
- `--session` requires Docker and a running gateway, even though the events on disk carry the session name and the default scope is explicitly readable after exit.
- `internal/util.FormatBytes` duplicates the existing `formatBytes` in `internal/app/cleanup.go`.
- `internal/netlog` is linked into the gateway proxy build inputs wholesale, including the terminal rendering and reader side the sidecar never uses.

Scope is otherwise tight; the only stray change is a `docs/DEV.md` paragraph left over from the config option that was dropped in 6a6c15f.

internal/util/size.go
planger: Done — `util.FormatBytes` is now generic over `int64 | uint64`, the `app` copy is gone, and its call sites point here. A type parameter rather than conversions at the call sites because every `app` caller passes `uint64` (`realDockerRootFreeSpace`, `docker.BuildCacheUsage`, `BuildCachePruneReport.SpaceReclaimed`), so casting would have meant five gosec G115 suppressions.

One visible change worth flagging: the build-cache messages now drop the decimal above ten units (`18 GB` rather than `18.4 GB`), which is the shared formatter's behaviour. The storage-preflight messages are unaffected — both thresholds are ≤ 10 GB, so the value they print always keeps its decimal.

internal/netlog/aggregate.go
planger: Good catch — `--summary` and `--verdict deny` really did disagree. Those events are now grouped under the empty domain instead of being skipped, so they reach both the per-domain rows and the totals. The human summary labels the row `(no domain)` and sorts it below named domains it ties with; the JSON keeps `"domain": ""` rather than inventing a host name a consumer could mistake for a real one. Covered by tests and documented in `docs/networking.md`.

gateway-entrypoint.sh
planger: Fixed: `su-exec "$PROXY_USER:$PROXY_USER"`. It needs nothing root gave it — it opens the dnsmasq log read-only through the follower and appends to the mode-666 network log, exactly what the MITM proxy already does as that user, and the `>>` redirect fd is opened by the root shell before the exec on both the initial start and the supervisor restarts. `su-exec` execs rather than forks, so `$!` stays the real pid and `stop_child_pid`/`is_pid_running` keep working.

internal/app/network_log_cmd.go
planger: Fixed. Gateway discovery still runs first and deliberately looks at every running gateway, which is how `--session` can name a running session of another project; when nothing matches, or Docker is unreachable, the read falls back to this project's log with the session filter derived from the name — stripping the gateway suffix is the exact inverse of how the stamp is built, so both accepted name forms resolve to the same session.

Two things fell out of wiring that up:

- A name that appears on no event is now an error rather than an empty success. Every gateway start writes a marker carrying its session, so absence is conclusive: `No events for session <name> in current project/tool (<hash>/<tool>); check the session name`.
- That message names the log that was searched, because a session of another tool lives in a different file and its name alone would not say where the read looked.

`docs/cli-reference.md` and `docs/networking.md` updated.

internal/netlog/filter.go
planger: Agreed — that was a policy rule leaking into a read. Added `domainpattern.NormalizeQuery`, and `Normalize` is now that plus the one allowlist-only check, so the relationship is explicit: the allowlist rules are the query rules plus the requirement that a wildcard covers two labels. No allowlist caller changed behaviour. Bare `*` stays rejected — `MatchNormalizedHost` cannot match it, so accepting it would silently return nothing; omitting the flag is how you match everything.

Worth noting the breadth was already reachable: `--domain com` was always accepted and matches every `.com` host, because a non-wildcard pattern implicitly covers subdomains. So this removes an inconsistency rather than granting a new query.

internal/app/network_log_cmd.go
planger: Fixed. `RenderEvent` is gone, replaced by `netlogview.WriteEvent`, which shares `renderSessionHeader`/`renderHumanRow` and the write helper with `WriteEvents`. A live boundary now gets the same blank-line separation, and a live row is byte-identical to the backlog row for the same event — there is a test asserting exactly that rather than a pinned column layout. Counts stay omitted because the session the marker opens has not produced any yet.

One asymmetry left on purpose: a followed marker always emits its leading blank line, while `WriteEvents` suppresses it at index 0, so `--follow` on an empty backlog opens with a blank line. Plumbing "is this the first line of output" through for that seemed worse than the blank line.

internal/gateway/gateway_proxy_build_inputs.txt
planger: Split: `filter.go`, `aggregate.go` and `render.go` moved to a new `internal/netlogview`, leaving `internal/netlog` with the format plus the file handling. `go list -deps ./cmd/enclave-gateway-proxy` no longer reaches the query side, and since the manifest and `assets_embed.go` are whitelists, neither needed a change — the sidecar simply stops staging and linking it. `netlog` also no longer imports `internal/logx`.

Two things stayed where they were, deliberately: `scan.go` and the line splitter, because the `Follower` that `dnsaudit` uses needs `ParseLine` and the same line bound, and `rotate.go`, which sits next to `append.go` as file handling and is called by the host-side launcher. `internal/logx` still reaches the sidecar, but through `internal/util` (its SELinux notice), which the proxy uses directly — separating that would be a different cut.

docs/DEV.md
planger: You are right that it came in with `network_log_max_size` and outlived it. I have kept it anyway: the remaining text describes a mechanism that still exists — `host_config_paths` has no `CLIFlags`, and `applyProjectDefaultsGuardrails` guards it — so it is accurate documentation rather than a stale leftover, and I would rather not spend a commit isolating four lines. Happy to pull it out if you prefer this PR to touch only the reader.

internal/netlogview/render.go
```diff
@@ -0,0 +1,289 @@
...
+	// to grow four times per event.
+	line.Grow(160)
+	line.WriteString(" ")
+	line.WriteString(opts.colorize(pad(timestamp, widthTime), logx.ColorDim))
+	line.WriteString("  ")
+	line.WriteString(opts.colorize(glyph, glyphColor))
+	line.WriteString("  ")
+	line.WriteString(pad(humanKind(event), widthMethod))
+	line.WriteString(" ")
+	line.WriteString(opts.colorize(pad(humanDomain(event), widthDomain), logx.ColorCyan))
```
xai, line 185: **Escape untrusted host names before terminal rendering:**
The human row writes `event.Domain` directly to `stdout`, and `RenderSummary` does the same. This field is attacker-controlled even for denied connections: the `TLS ClientHello` parser accepts a nonempty SNI through `NormalizeHost` without rejecting control bytes, then records that host on an allowlist denial. JSON escaping protects the log file but is undone by ParseLine. A host containing an OSC sequence such as `evil\x1b]0;controlled\a.example` survives normalization and is emitted as raw terminal controls even with `Color=false` (confirmed with the current renderer). A sandboxed process can therefore manipulate the host terminal when the user inspects the audit log. Escape control characters in all event-derived display cells before padding/coloring, including summary domains, and add a crafted-SNI regression test.

internal/netlogview/render.go
```diff
@@ -0,0 +1,289 @@
...
+// session's verdict counts.
+func WriteEvents(out io.StringWriter, events []netlog.Event, opts RenderOptions) error {
+	for index, event := range events {
+		if event.IsSessionMarker() {
+			separator := ""
+			if index > 0 {
+				separator = "\n"
+			}
+			pass, deny := countUntilNextMarker(events[index+1:])
+			if err := writeParts(out, separator, renderSessionHeader(event, pass, deny, opts), "\n\n"); err != nil {
```
xai: **Attribute interleaved events using their session field:**  
Session headers count every event until the next marker rather than events belonging to the named session. Concurrent sessions of one project/tool share a log, and `--all-running` also interleaves independent logs. For the valid sequence `start(A)`, `start(B)`, `deny(session=A)`, the renderer prints A with no counts and B with `"0 pass 1 deny"`, with A's request beneath B's header. Rows omit the session field, so users cannot identify the actual source. Compute counts by session and make row ownership explicit for interleaved streams (or group by session), with a regression test containing events from A after B starts.

xai (changes requested):
Really nice feature 👍 👍 

Looks pretty good, two small problems, see the inline comments of `render.go`.

internal/netlogview/render.go
planger: 👍 Counts are now attributed by the event's session field, with the next marker of the same session ending the span (session names recur across runs), plus a regression test with events from A arriving after B started.

Row ownership I left out on purpose: making it explicit means either a session column or grouping rows under their header, and grouping would undo the deliberate time-ordered merge `--all-running` does. I've deferred that for now, as it is probably a decision we need to make deliberately.

internal/netlogview/render.go
planger: 👍 Confirmed and fixed: `NormalizeHost` does not reject control bytes, so a crafted SNI reached the terminal verbatim. The terminal form now maps non-printable runes (control bytes, C1, format chars, and invalid UTF-8) to `?` for every event-derived cell, in rows, session headers, and summary domains, which also keeps the padding honest. Regression test asserts no `\x1b`/`\a` survives, for both the row and the summary path. `--json` was already safe, since the escaping there is the JSON encoder's.

xai:
Thanks for the quick fixes @planger. Can be approved after a rebase.

xai (approved):
LGTM, works great 👍

PR #53, opened 2026-08-14
fix: resolve tool extension config assets from the user extension tree

planger:
## Problem

A tool extension installed under `~/.config/enclave/extensions/tools/<tool>/` could not use `sandbox.settingsFile`. Session start aborted with:

```
prepare config source for <tool>: built-in settings template missing at
~/.cache/enclave/assets/<hash>/extensions/tools/<tool>/templates/settings.json
```

The runtime resolved tool extension assets against `paths.ToolsDir` only. That path points into the content-hashed assets tree extracted from the embedded FS, whereas user extensions are merged into a temporary image build context and never staged there. Spec loading and validation already searched both trees, so `enclave validate-extensions` reported such a tool as valid while the runtime refused to start a session for it.

The gateway allowlist had the same defect and failed silently: a user extension's `gateway-allowlist.conf` was never found, so the session fell back to `base.conf`, which allows more domains than the extension declared.

## Fix

Resolve these through the existing `config.ResolveToolFile` and `ResolveToolDirs` helpers — user tree first for single files, built-in before user for directories — matching the per-file precedence the image build context already applies:

- the settings template
- `config-base/`, both in the overlay and in the mount decision
- `gateway-allowlist.conf`

Settings template resolution moves into `config.ResolveToolSettingsTemplate`, shared by the runtime and by `validate-extensions`, so the two cannot drift apart again.

`settings_file` is spec data joined onto a `templates/` directory both on the host and inside the container, so `validateAndNormalizeProfile` now enforces its naming contract at spec load: the value is trimmed, and an empty template name or an embedded path separator is rejected. Every consumer therefore sees a value it can join safely, including the container-side template path handed to the entrypoint.

The two-tree walk for extension subdirectories is now `config.ResolveToolSubdirs` / `ResolveFeatureSubdirs`; the tool config-base, tool skills, and feature skills lookups all go through it instead of hand-rolling the same `os.Stat`/`IsDir` loop. `config-base` detection requires a directory instead of any node.

`applyBuiltInConfigLayer` / `applyBuiltInSkillsLayer` are renamed to `applyExtensionConfigLayer` / `applyExtensionSkillsLayer`, since both now compose the built-in and the user extension tree.

## Tests

New coverage for user-tree resolution and precedence of the settings template, `config-base` overlay from both trees, the config-source mount decision, allowlist resolution (user tree, precedence, base fallback), and rejection of an invalid or traversing `settings_file` at profile load.

`make build`, `make test`, `make lint`, and `make check-license-headers` pass.

### DIFF
--- a/docs/extensions/README.md
+++ b/docs/extensions/README.md
@@ -236,7 +236,7 @@ additional files beyond `spec.yaml`.
 | File | Purpose |
 |------|---------|
 | `spec.yaml` | Extension spec: `kind: sandbox`, `sandbox.*` metadata, `network`, `credentials`, `providers`, `ports` |
-| `gateway-allowlist.conf` | dnsmasq config for network isolation |
+| `gateway-allowlist.conf` | dnsmasq config for network isolation. Resolved from the user extension tree ahead of the built-in one; a tool that ships none falls back to the broader `base.conf` |
 | `install.sh` | Installation script run during Docker build |
 
 ### Optional Files
@@ -244,6 +244,7 @@ additional files beyond `spec.yaml`.
 | File | Purpose |
 |------|---------|
 | `templates/` | Settings templates copied to `/usr/local/share/enclave/templates/` |
+| `config-base/` | Seed files for the tool's config store, overlaid into the generated config source before the host and per-project layers |
 | `check-update.sh` | Optional prebuild hook that returns a stable upstream fingerprint for automatic update probes |
 | `entrypoint.d/*.sh` | Scripts sourced at container startup (only for matching tool) |
 | `go/` | Go code for custom hooks/handlers (compiled into binary) |
@@ -306,9 +307,15 @@ tool metadata. `sandbox.entrypoint.run` is the shared sbx-style command to
 launch the tool.
 
 When using `templates/`, set `sandbox.configDir`, `sandbox.settingsFile`
-(aggregated name like `<tool>-settings.json`), and `sandbox.settingsTarget`
-(path under `configDir`). Runtime composes the built-in template into the
-generated tool config source before container startup.
+(aggregated name like `<tool>-settings.json`: the `<tool>-` prefix followed by
+the template's bare filename, path separators are rejected), and `sandbox.settingsTarget`
+(path under `configDir`). Runtime composes the template into the generated tool
+config source before container startup. The template and the optional
+`config-base/` directory are resolved from the built-in tree and from
+`~/.config/enclave/extensions/tools/<tool>/`, with the user extension tree
+winning per file, the same precedence the image build context applies. A
+user-global tool extension can therefore use `templates/` without being
+upstreamed.
 
 If a tool supports host config passthrough, declare a narrow reviewed
 `sandbox.passthroughPaths` allow-list. Host passthrough is fail closed: only
@@ -686,7 +693,7 @@ Hooks run in a fixed order during runtime auth preparation:
 - `internal/model/types.go`: `Extension` struct with `IsMixin()` and `IsSandbox()` methods
 - `internal/config/extension.go`: Extension loading and filtering functions
 - `internal/config/profile.go`: `ListProfiles()` returns only tool extensions
-- `internal/runtime/network_manager.go`: Loads DNS allowlists from `extensions/tools/{tool}/gateway-allowlist.conf`
+- `internal/runtime/network_manager.go`: Loads DNS allowlists from `extensions/tools/{tool}/gateway-allowlist.conf`, resolved from the user extension tree ahead of the built-in one
 
 ## Adding a New Tool Extension
 
--- a/internal/config/extension.go
+++ b/internal/config/extension.go
@@ -12,6 +12,7 @@ import (
 	"os"
 	"path/filepath"
 	"sort"
+	"strings"
 
 	"enclave/internal/model"
 	"enclave/internal/util"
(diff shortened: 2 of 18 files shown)


internal/config/extension.go
```diff
@@ -12,26 +12,64 @@ import (
...
+		return "", fmt.Errorf("settings_file %q must start with %q", settingsFile, prefix)
+	}
+	templateName := strings.TrimPrefix(settingsFile, prefix)
+	if templateName == "" {
+		return "", fmt.Errorf("settings_file %q is missing a template name after %q", settingsFile, prefix)
+	}
+	// The template name is a bare filename joined onto templates/. Rejecting
+	// separators keeps a spec from escaping the extension tree and copying an
+	// arbitrary host file into the container's config source.
+	if strings.ContainsAny(templateName, `/\`) {
```
EclipseSourceAI, line 59: The separator guard only fires when the config source is actually composed. `runtime.go` still builds the container template path by concatenating the raw `SettingsFile` ([L1569-L1573](https://github.com/eclipse-enclave/enclave/blob/eb9261a8bc2c306986bff30678ddf1812e39c2d1/internal/runtime/runtime.go#L1569-L1573)), which `entrypoint.sh` then copies from, so a traversing name still escapes `templates/` on that path. `validateAndNormalizeProfile` ([L47-L50](https://github.com/eclipse-enclave/enclave/blob/eb9261a8bc2c306986bff30678ddf1812e39c2d1/internal/config/profile.go#L47-L50)) is where profile fields get normalized at load and would cover every consumer.

internal/config/extension.go
```diff
@@ -12,26 +12,64 @@ import (
...
-	candidate := filepath.Join(paths.ToolsDir, toolName, fileName)
-	if util.FileExists(candidate) {
-		return candidate, true
-	}
 	return "", false
 }
 
+// ToolFileCandidates lists the paths ResolveToolFile searches, in its
+// precedence order: the user extension tree ahead of the built-in one.
+func ToolFileCandidates(paths model.Paths, toolName string, fileName string) []string {
```
EclipseSourceAI: `ToolFileCandidates` is only used inside `config` (by `ResolveToolFile` and the error message below). Keep it unexported until something outside the package needs it.

internal/runtime/tool_config_source.go
```diff
@@ -175,34 +174,38 @@ func (r *Runtime) applyBuiltInConfigLayer(targetDir string) error {
...
-	if !strings.HasPrefix(r.profile.SettingsFile, prefix) {
-		return "", fmt.Errorf("settings_file %q must start with %q", r.profile.SettingsFile, prefix)
-	}
-	templateName := strings.TrimPrefix(r.profile.SettingsFile, prefix)
-	hostTemplatePath := filepath.Join(r.paths.ToolsDir, r.profile.Name, model.TemplatesDir, templateName)
-	if !util.PathExists(hostTemplatePath) {
-		return "", fmt.Errorf("built-in settings template missing at %s", hostTemplatePath)
+// extensionConfigBaseDirs returns the existing config-base directories of the
+// tool extension, built-in tree first so the user extension tree overlays it.
+func (r *Runtime) extensionConfigBaseDirs() []string {
```
EclipseSourceAI, line 193: `extensionSkillSourceDirs` already walks the same two trees for `skills/` ([L86-L98](https://github.com/eclipse-enclave/enclave/blob/eb9261a8bc2c306986bff30678ddf1812e39c2d1/internal/runtime/skills_mount.go#L86-L98)), including the hand-rolled `os.Stat`/`IsDir` this PR just extracted into `util.IsDir`. Folding both into one helper parameterized by the subdirectory name would keep the two from drifting.

internal/util/util.go
```diff
@@ -308,6 +308,11 @@ func FileExists(path string) bool {
 	return err == nil && !info.IsDir()
 }
 
+func IsDir(path string) bool {
```
EclipseSourceAI, line 311: `config.dirExists` ([L43-L46](https://github.com/eclipse-enclave/enclave/blob/eb9261a8bc2c306986bff30678ddf1812e39c2d1/internal/config/project_marker.go#L43-L46)) is byte-identical to this and sits in the same package the PR just removed `isDir` from. Drop it and call `util.IsDir` there too.

docs/extensions/README.md
```diff
@@ -306,9 +307,15 @@ tool metadata. `sandbox.entrypoint.run` is the shared sbx-style command to
...
-generated tool config source before container startup.
+(aggregated name like `<tool>-settings.json`: the `<tool>-` prefix followed by
+the template's bare filename, path separators are rejected), and `sandbox.settingsTarget`
+(path under `configDir`). Runtime composes the template into the generated tool
+config source before container startup. The template and the optional
+`config-base/` directory are resolved from the built-in tree and from
+`~/.config/enclave/extensions/tools/<tool>/`, with the user extension tree
+winning per file, the same precedence the image build context applies. A
+user-global tool extension can therefore use `templates/` without being
+upstreamed.
```
EclipseSourceAI, line 318: The allowlist half of the fix is undocumented. Further down this file ([L688](https://github.com/eclipse-enclave/enclave/blob/eb9261a8bc2c306986bff30678ddf1812e39c2d1/docs/extensions/README.md#L688-L688)) and the resolution order in `docs/networking.md` ([L64-L69](https://github.com/eclipse-enclave/enclave/blob/eb9261a8bc2c306986bff30678ddf1812e39c2d1/docs/networking.md#L64-L69)) still describe the built-in tree only, and this is the security-relevant behavior change since a user extension previously fell through to the broader `base.conf`.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

## What this PR does

Tool extensions installed under `~/.config/enclave/extensions/tools/<tool>/` were resolved against `paths.ToolsDir` only, which points into the content-hashed assets tree extracted from the embedded FS. User extensions never land there, so `sandbox.settingsFile` aborted session start and `gateway-allowlist.conf` silently fell back to the broader `base.conf`. The fix routes the settings template, `config-base/`, and the gateway allowlist through the existing `config.ResolveToolFile` / `ResolveToolDirs` helpers, moves template resolution into a shared `config.ResolveToolSettingsTemplate` used by both the runtime and `validate-extensions`, and renames `applyBuiltInConfigLayer` / `applyBuiltInSkillsLayer` to `applyExtension*`.

The approach fits the codebase: it reuses the resolution helpers rather than adding a parallel mechanism, and `internal/network` already depended on `internal/config`, so the `ResolveToolAllowlist` signature change introduces no new coupling. I verified the gateway build context handles an allowlist outside `AppRoot` (`prepareGatewayContext` stages it), so the user-tree allowlist path works end to end. Build, vet, and the touched package tests pass locally on the merged state.

- The new `settings_file` path-separator guard lives in the config-source composition path, but `runtime.go` still builds the container-side template path from the raw value, so the hardening is partial (low severity, container-internal).
- `extensionConfigBaseDirs` re-implements the two-tree walk that `extensionSkillSourceDirs` already does.
- The gateway allowlist behavior change (the security-relevant half) is not reflected in `docs/extensions/README.md` or `docs/networking.md`.

Test coverage for the new resolution paths and precedence looks good.

internal/config/extension.go
planger: Correct, the guard was in the wrong place. Moved it to `validateAndNormalizeProfile`: `settings_file` is trimmed and checked at spec load (`<tool>-` prefix, non-empty template name, no path separator), so every consumer — including `addToolSettingsTemplate`, which builds the container-side `/usr/local/share/enclave/templates/` path — sees a value that already satisfies the contract. `ResolveToolSettingsTemplate` and the load-time normalizer share `toolSettingsTemplateName`, so the rule is stated once. Covered by `TestValidateAndNormalizeProfileValidatesSettingsFile`.

internal/config/extension.go
planger: Unexported to `toolFileCandidates`.

internal/runtime/tool_config_source.go
planger: Folded both into `config.ResolveToolSubdirs` / `config.ResolveFeatureSubdirs`, layered on the existing `ResolveToolDirs` / `ResolveFeatureDirs`. `extensionConfigBaseDirs`, `extensionSkillSourceDirs`, and `featureSkillSourceDirs` are now one-liners over those, so the two-tree walk and the `IsDir` check exist in one place instead of three. Also added `model.ConfigBaseDirName` so the subdirectory name is a constant rather than a literal.

internal/util/util.go
planger: Dropped `config.dirExists`; `WriteProjectMarkers` calls `util.IsDir`.

docs/extensions/README.md
planger: Documented. The `gateway-allowlist.conf` row in the required-files table and the framework-integration line both state the user-tree-first resolution and the `base.conf` fallback, and `docs/networking.md` gained a paragraph under "Overriding the Main Allowlist" covering the resolution order and the fact that the fallback is broader than a tool-specific allowlist.

xai (approved):
Looks good and tested OK! Thanks!

PR #54, opened 2026-08-14
Add version command with build metadata

xai:
#### What it does

To better debug reported issues, I want to add a version command that includes the build's commit hash.

- Adds `enclave version` with the version, source commit, and commit date.
- Stamps standalone and packaged builds with release metadata.

#### How to test

```sh
make build
./bin/enclave version
make test
make lint
```

#### Follow-ups

None.

#### Breaking changes

#### Review checklist


internal/buildinfo/buildinfo.go
```diff
@@ -0,0 +1,121 @@
...
+	Version string
+	Commit  string
+	Date    string
+}
+
+func Read() Info {
+	goInfo, ok := debug.ReadBuildInfo()
+	if !ok {
+		goInfo = nil
+	}
```
EclipseSourceAI: `debug.ReadBuildInfo` already returns nil when `ok` is false, so this branch is a no-op. `goInfo, _ := debug.ReadBuildInfo()` does the same thing.

internal/buildinfo/buildinfo.go
```diff
@@ -0,0 +1,121 @@
...
+)
+
+const unknown = "unknown"
+
+// These values are populated with -ldflags by release and package builds.
+var (
+	Version string
+	Commit  string
+	Date    string
+)
```
EclipseSourceAI, line 24: `-ldflags -X` works on unexported symbols too, so these can be `version`/`commit`/`date`. Exported they invite callers to read the raw value (empty on a plain `go build`) instead of going through `Read()`.

internal/buildinfo/buildinfo.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+	return resolve(Info{Version: Version, Commit: Commit, Date: Date}, goInfo)
+}
+
+func (i Info) String() string {
+	return fmt.Sprintf("%s (%s, %s)", i.Version, i.Commit, i.Date)
+}
+
+func resolve(info Info, goInfo *debug.BuildInfo) Info {
+	commitFromBuildInfo := isUnknown(info.Commit)
```
EclipseSourceAI, line 44: Why `vcs.modified` is only honored when the commit came from Go's build info is the subtlest part of this file (the deb job rewrites `debian/changelog` before building, so an explicitly stamped commit must stay clean). One line of comment here would save the next reader the reconstruction.

internal/app/app.go
```diff
@@ -52,6 +39,25 @@ func Run(args []string) int {
 	if parsed.Action == "completion" {
 		return 0
 	}
+	if parsed.Action == "version" {
+		fmt.Printf("Enclave: %s\n", buildinfo.Read())
```
EclipseSourceAI: `Enclave` is hardcoded while the rest of the CLI derives the program name from `model.AppName`, e.g. [info.go](https://github.com/eclipse-enclave/enclave/blob/eb9261a8bc2c306986bff30678ddf1812e39c2d1/internal/app/info.go#L71). Use that so the casing matches every other command.

.github/workflows/ci.yml
```diff
@@ -25,7 +25,7 @@ jobs:
           go-version-file: go.mod
 
       - name: Build CLI
-        run: CGO_ENABLED=0 go build ./cmd/enclave
+        run: CGO_ENABLED=0 make build BIN_DIR=.
```
EclipseSourceAI, line 28: The linker silently ignores an `-X` flag whose symbol path does not resolve, so renaming or moving `internal/buildinfo` turns every build into `unknown (unknown, unknown)` with all tests still green. Asserting in [smoke-standalone-binary.sh](https://github.com/eclipse-enclave/enclave/blob/eb9261a8bc2c306986bff30678ddf1812e39c2d1/scripts/smoke-standalone-binary.sh#L40-L47) that `enclave version` contains no `unknown` would catch that. The macOS `Build CLI` step further down in this file also still uses plain `go build`, so it stays unstamped.

Makefile
```diff
@@ -93,7 +100,9 @@ deb:
 	mkdir -p dist
 	@tmpdir=$$(mktemp -d) && \
 	cp -r . "$$tmpdir/enclave" && \
-	cd "$$tmpdir/enclave" && dpkg-buildpackage -us -uc -b && \
+	cd "$$tmpdir/enclave" && \
+	BUILD_VERSION='$(BUILD_VERSION)' BUILD_COMMIT='$(BUILD_COMMIT)' BUILD_DATE='$(BUILD_DATE)' \
```
EclipseSourceAI: Exporting `BUILD_VERSION` here overrides the `dpkg-parsechangelog` default that `debian/rules` now has, so a local `make deb` stamps the Makefile `VERSION` even when `debian/changelog` says something else. That is also why the CI job has to re-derive `VERSION` from the changelog. Dropping `BUILD_VERSION` from the two deb targets makes both paths agree on the changelog as the source of truth.

Makefile
```diff
@@ -14,6 +14,13 @@ UNAME_S := $(shell uname -s)
 CMD_DIR := ./cmd/enclave
 COMPLETIONS_DIR ?= ./completions
 VERSION ?= 0.1.0
+GIT_COMMIT := $(shell git rev-parse --short HEAD 2>/dev/null)
+GIT_COMMIT_DATE := $(shell git show -s --format=%cs HEAD 2>/dev/null)
+GIT_DIRTY := $(shell if [ -n "$$(git status --porcelain 2>/dev/null)" ]; then printf '%s' '-dirty'; fi)
+BUILD_VERSION ?= $(VERSION)
+BUILD_COMMIT ?= $(if $(strip $(GIT_COMMIT)),$(GIT_COMMIT)$(GIT_DIRTY),unknown)
+BUILD_DATE ?= $(if $(strip $(GIT_COMMIT_DATE)),$(GIT_COMMIT_DATE),unknown)
+BUILD_LDFLAGS := -X enclave/internal/buildinfo.Version=$(BUILD_VERSION) -X enclave/internal/buildinfo.Commit=$(BUILD_COMMIT) -X enclave/internal/buildinfo.Date=$(BUILD_DATE)
```
EclipseSourceAI: These are immediate (`:=` for `BUILD_LDFLAGS` forces expansion of the `?=` vars), so every `make test`, `make lint`, and `make fmt` shells out to git three times, including `git status --porcelain` over the whole tree. Making `GIT_*` and `BUILD_LDFLAGS` recursive (`=`) confines that to the build recipes.

.github/workflows/rolling.yml
```diff
@@ -70,7 +70,9 @@ jobs:
           go-version-file: go.mod
 
       - name: Build standalone binary
-        run: CGO_ENABLED=0 GOOS=${{ matrix.goos }} GOARCH=${{ matrix.goarch }} go build -o ${{ matrix.output }} ./cmd/enclave
+        run: |
+          VERSION=$(./scripts/rolling-package-version.sh)
+          CGO_ENABLED=0 GOOS=${{ matrix.goos }} GOARCH=${{ matrix.goarch }} make build BIN_DIR=. BINARY=${{ matrix.output }} VERSION="$VERSION"
```
EclipseSourceAI: This makes the standalone binary report the Debian package version, so `enclave version` prints something like `0.1.1~git.20260814.42.abcdef12 (abcdef1, 2026-08-14)` where the commit and date are already duplicated in the version field. Packaging syntax like `~git.` earns its keep in a .deb/.rpm, less so in a plain binary.

internal/cli/parse.go
```diff
@@ -84,6 +84,7 @@ directly: "enclave --tool codex" is equivalent to "enclave run --tool codex".`,
 		shellCommand(&res),
 		authCommand(&res),
 		infoCommand(&res),
+		simpleCommand("version", "Show binary version and source commit", &res),
```
EclipseSourceAI: Cobra hands you `--version` for free if you set `rootCmd.Version`, and that is what most people type first. A `--json` would also match `ps`/`status`/`config` ([ps_command.go](https://github.com/eclipse-enclave/enclave/blob/eb9261a8bc2c306986bff30678ddf1812e39c2d1/internal/cli/ps_command.go#L26)) and the structured-output rule in [AGENTS.md](https://github.com/eclipse-enclave/enclave/blob/eb9261a8bc2c306986bff30678ddf1812e39c2d1/AGENTS.md#L22), since integrators will want to read the version as well.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Adds `enclave version` backed by a new `internal/buildinfo` package, plus `-ldflags -X` stamping wired through `make build`, `debian/rules`, and the RPM spec so standalone and packaged builds carry version, short commit, and commit date. `buildinfo.resolve` prefers linked values and falls back to Go's own `debug.ReadBuildInfo` VCS stamps, with `unknown` as the floor. `internal/app.Run` now parses args before resolving the project dir and loading config, so `version` works even from a directory that no longer exists (covered by a new test).

Verified locally: `make build`, `./bin/enclave version`, and tests for the touched packages pass, gofmt and the license header check are clean.

The scope is tight and the layering is right. Where a maintainer should focus:

- Nothing verifies that the stamps actually land. A stale `-X` symbol path is silently ignored by the linker, so this can degrade to `unknown (unknown, unknown)` with a green build. The macOS CI build step also still uses plain `go build`.
- The deb path has two sources of truth for the version: the Makefile exports `BUILD_VERSION` from `VERSION` and thereby defeats the `dpkg-parsechangelog` default just added to `debian/rules`, which is why the CI job re-derives it from the changelog.
- Rolling standalone binaries now report the Debian-style package version, so the commit and date show up twice in one line.

Smaller items on the new package and the Makefile variable block are inline.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Follow-up on the six commits since my last review (dcddd6d through c23b874).

Both remaining open threads from my earlier review are addressed: `Version`/`Commit`/`Date` are now unexported (`version`/`commit`/`date`), and `resolve()` in `internal/buildinfo/buildinfo.go` has a comment explaining why `vcs.modified` only gates the dirty suffix when the commit came from Go's build info. The other threads (macOS CI build, deb version double-source-of-truth, rolling package version format, Makefile immediate vs. recursive vars, `Enclave` hardcoding, cobra `--version`/`--json`) were already marked resolved and I checked the corresponding code is in fact fixed.

The new `--version`/`-v` root flag and `version --json` are wired cleanly: `hasVersionFlag` bypasses `normalizeArgs` only for the bare top-level case, which is the only case that needed it, and the smoke-test/CI changes now assert the binary actually gets stamped (`enclave version` output has no `unknown`). Build, vet, and tests for the touched packages pass locally.

- [`buildinfo.Version`/`Commit`/`Date` made unexported so callers can't bypass `Read()`](https://github.com/eclipse-enclave/enclave/pull/54#discussion_r3786128046)
- [added a comment explaining why `vcs.modified` is only honored for build-info-derived commits](https://github.com/eclipse-enclave/enclave/pull/54#discussion_r3786128125)

xai:
rebased against main

planger (approved):
Thank you, looks great!

PR #55, opened 2026-08-24
fix: serve the website from the enclave.eclipse.dev root

planger:
#### What it does

The website is now hosted at https://enclave.eclipse.dev, so it is served from a
domain root rather than the planned `eclipse.dev/enclave` subpath. Production
builds still injected a `/enclave/docs/` base path, which broke the deployed
docs: pages referenced `/enclave/docs/assets/...` (404 today) while the files
themselves sit at `/docs/assets/...`, so no CSS or JS loaded.

- Production `DOCS_BASE_URL` and the `workflow_dispatch` default drop the
  `/enclave` prefix, so the docs build with the default `/docs/`.
- `docusaurus.config.js` sets `url: https://enclave.eclipse.dev` for correct
  sitemap and canonical URLs.
- The publish workflow now re-creates the `CNAME` file on every deploy. This is
  a latent trap unrelated to the base path: `actions-gh-pages` replaces the
  published tree, so the next deploy would have deleted the file the Pages
  settings UI created and unset the custom domain. Otterdog cannot manage this,
  as its repository resource has no custom-domain property.
- `index.html` gains `og:url`, a canonical link, and an absolute `og:image`. The
  relative `og:image` never worked for social crawlers; a stable domain is what
  was missing. These are the only place the production domain is hardcoded.
- README links the site under the title and in the project resources line.
- `package-lock.json` picks up a two-line rename: the lockfile `name` was still
  `website-docs` while `package.json` says `eclipse-enclave-docs`.

The preview workflow keeps its `/enclave-website-previews/pr-previews/pr-N/docs/`
base path, which is still correct for that host.

#### How to test

```bash
OUTPUT_DIR=/tmp/enclave-site ./website/build.sh
cd /tmp/enclave-site && python3 -m http.server 8000
```

Built pages should reference `/docs/assets/...`, and `sitemap.xml` should use
`https://enclave.eclipse.dev/docs/...`. I verified 200s on `/`, `/docs/`, all
three doc pages, the docs CSS bundle, and the marketing assets. `website/preview.sh`
covers the same ground. The only build warnings are the four intentional `../`
links from docs back to the marketing home, which is why `onBrokenLinks` is `warn`.

To confirm the current breakage before the fix lands:

```bash
curl -s https://enclave.eclipse.dev/docs/ | grep -oE 'href=/[^ >]*\.css'
# href=/enclave/docs/assets/css/styles.430d70dc.css  -> 404
```

#### Follow-ups

- The otterdog config in `.eclipsefdn` sets `homepage: https://www.eclipse.dev/enclave`
  for all three repos. That URL is now wrong and needs a separate PR there. -> https://github.com/eclipse-enclave/.eclipsefdn/pull/4

#### Breaking changes

#### Review checklist

### DIFF
--- a/.github/workflows/website-publish.yml
+++ b/.github/workflows/website-publish.yml
@@ -18,7 +18,7 @@ on:
       docs_base_url:
         description: "Docs base path (use /enclave-website/docs/ to test on the raw GitHub Pages URL)"
         type: string
-        default: "/enclave/docs/"
+        default: "/docs/"
 
 permissions:
   contents: read
@@ -44,7 +44,7 @@ jobs:
           node-version: "22.x"
       - name: Build site
         env:
-          DOCS_BASE_URL: ${{ inputs.docs_base_url || '/enclave/docs/' }}
+          DOCS_BASE_URL: ${{ inputs.docs_base_url || '/docs/' }}
         run: ./website/build.sh
       - name: Publish to enclave-website (gh-pages)
         uses: peaceiris/actions-gh-pages@4f9cc6602d3f66b9c108549d475ec49e8ef4d45e # v4.0.0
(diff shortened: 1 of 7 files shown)


tortmayr (approved):
LGTM!

PR #57, opened 2026-08-25
feat: mark session-owned terminals with a background tint

### LINKED ISSUE #2: Show a visible indicator when a terminal is running inside Enclave
It is easy to lose track of whether a shell is on the host or inside the sandbox, and a default cue would help. 
For example i could envision a lock or enclave symbol in the terminal title.

sgraband:
#### What it does
A sandboxed session is indistinguishable from an ordinary shell once the agent paints over the scrollback, so it is easy to run host commands in a terminal that belongs to a session, or the reverse. The new session_tint config option takes an #rrggbb value; run and attach set the terminal background via OSC 11 on start and reset it via OSC 111 on exit. tmux 3.3 or later consumes both, so inside tmux only the session's pane is tinted.

The reset also runs on the interrupt signals, because a stale tint with no session behind it is a worse marker than none; the signal keeps its default effect afterwards. SIGKILL still leaks a tint, which the docs cover.

Only the config key enables the tint. NO_COLOR and ENCLAVE_COLOR=never suppress it, but ENCLAVE_COLOR=always does not turn it on, since a terminal background is not log text and should not follow log-color auto-detection. resolveColorEnabled grows a ColorSuppressedByEnv split for that reason.

Fixes #2

#### How to test
Scratch config root, so the real config stays untouched:

```bash
mkdir -p /tmp/tint-test/config/enclave
echo '{"session_tint": "#2a0f12"}' > /tmp/tint-test/config/enclave/config.json
export XDG_CONFIG_HOME=/tmp/tint-test/config
./bin/enclave config --view source --json | grep -A3 session_tint
# -> "effective": "#2a0f12", "source": "global"
```

A scratch config root means an empty auth store, so a real session will ask for
a fresh tool login. For the session cases below it is less friction to put the
key in the real `~/.config/enclave/config.json` and remove it afterwards.

| # | Steps | Expected |
|---|---|---|
| 1 | `enclave shell`, then `exit` | tinted on start, restored on exit |
| 2 | `enclave -d`, then `enclave attach <name>`, detach with `ctrl-\` | launching terminal untinted; tinted while attached, restored on detach |
| 3 | `enclave shell`, then `kill -TERM $(pgrep -f 'enclave shell')` from another terminal | tint restored *and* enclave dies from the signal |
| 4 | `NO_COLOR=1 enclave shell` / `ENCLAVE_COLOR=never enclave shell` | no tint |
| 5 | config key removed, `ENCLAVE_COLOR=always enclave shell` | no tint — the config key is the only switch |
| 6 | `"session_tint": "red"` | `Ignoring invalid session_tint …` warning, no tint, session runs normally |

`SIGKILL` leaves the tint behind by design; clear it with `printf '\e]111\a'`.

#### Breaking changes

#### Review checklist

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -65,6 +65,7 @@ supported under `tool_overrides.<tool>`.
 | `no_history` | Disable shell history |
 | `no_memory` | Disable per-project agent memory |
 | `session_monitor` | Run agents under the managed tmux session (enables `status` snapshots) |
+| `session_tint` | Terminal background color marking a session-owned terminal, as `#rrggbb` (unset: no tint) |
 | `base_image` | Docker base image override |
 | `devcontainer` | Derive base image from devcontainer.json |
 | `slim` | Build without features (tools only) |
@@ -99,6 +100,22 @@ on the host. Project config may strengthen the inherited mode from `follow` to
 the inherited mode are ignored. For a regular repository whose `.git`
 directory sits inside the project, the project mount mode governs and this
 option has no effect.
+
+`session_tint` marks the terminal that owns a session, so a sandboxed session
+is visually distinct from an ordinary shell. When set to an `#rrggbb` value,
+`run` (including `shell`, `continue`, and `resume`) and `attach` set the
+terminal background on start and reset it on exit. Inside tmux 3.3 or later only
+the session's pane is tinted; elsewhere the whole window is. Because only the
+background changes, pick a color that keeps your foreground text readable —
+enclave does not adjust it. Invalid values are reported and ignored.
+
+The tint is skipped when stdout is not a terminal and when `NO_COLOR` or
+`ENCLAVE_COLOR=never` is set; `ENCLAVE_COLOR=always` does not enable it, as the
+config key is the only switch. Because per-tool and per-project config layers
+apply, `tool_overrides.<tool>.session_tint` gives each tool its own color and a
+project config gives each project one. Agents that paint their own background
+can cover the tint. If enclave is killed with `SIGKILL` the reset never runs;
+clear a leftover tint with `printf '\e]111\a'`.
 
 **Example:**
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -103,19 +103,26 @@ option has no effect.
 
 `session_tint` marks the terminal that owns a session, so a sandboxed session
 is visually distinct from an ordinary shell. When set to an `#rrggbb` value,
-`run` (including `shell`, `continue`, and `resume`) and `attach` set the
-terminal background on start and reset it on exit. Inside tmux 3.3 or later only
-the session's pane is tinted; elsewhere the whole window is. Because only the
-background changes, pick a color that keeps your foreground text readable —
-enclave does not adjust it. Invalid values are reported and ignored.
+`run` (including `shell`, `continue`, and `resume`), `exec`, and `attach` set
+the terminal background on start and reset it on exit. Inside tmux 3.3 or later
+only the session's pane is tinted; elsewhere the whole window is. Because only
+the background changes, pick a color that keeps your foreground text readable —
+enclave does not adjust it. Invalid values are reported when the config file is
+read and ignored.
 
 The tint is skipped when stdout is not a terminal and when `NO_COLOR` or
 `ENCLAVE_COLOR=never` is set; `ENCLAVE_COLOR=always` does not enable it, as the
 config key is the only switch. Because per-tool and per-project config layers
 apply, `tool_overrides.<tool>.session_tint` gives each tool its own color and a
-project config gives each project one. Agents that paint their own background
-can cover the tint. If enclave is killed with `SIGKILL` the reset never runs;
-clear a leftover tint with `printf '\e]111\a'`.
+project config gives each project one; `attach` resolves the color for the tool
+and project of the session it attaches to, not for the current directory.
+Agents that paint their own background can cover the tint.
+
+The reset restores the terminal's configured default background, not whatever
+background was in effect before the session, so a color set at runtime (theme
+switchers, base16-style scripts) is dropped when the session ends. If enclave is
+killed with `SIGKILL` the reset never runs; clear a leftover tint with
+`printf '\e]111\a'`.
 
 **Example:**
 
--- a/internal/app/command_attach.go
+++ b/internal/app/command_attach.go
@@ -12,6 +12,7 @@ import (
 	"strings"
 
 	"enclave/internal/backend"
+	"enclave/internal/config"
 	"enclave/internal/logx"
 	"enclave/internal/model"
 	"enclave/internal/termtint"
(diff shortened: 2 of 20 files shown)


sgraband:
@EclipseSourceAI

internal/runtime/runtime.go
```diff
@@ -753,6 +754,8 @@ func (r *Runtime) runContainer(ctx *ExecutionContext) error {
 	if be == nil {
 		return fmt.Errorf("runtime backend is not configured")
 	}
+	restoreTint := termtint.Begin(r.run.SessionTint)
```
EclipseSourceAI, line 764: `enclave exec` also hands the terminal to a running session with an interactive TTY ([exec.go](https://github.com/eclipse-enclave/enclave/blob/64d00f4a61b34727260d48ed12c3c22190f258a0/internal/runtime/exec.go#L18-L47)) and stays untinted. That is the same "looks like an ordinary shell" case the PR is about, so either tint it too or say in the docs why `exec` is out of scope.

internal/app/command_attach.go
```diff
@@ -31,6 +32,8 @@ func runAttach(run model.RunOptions) int {
 		logx.Errorf("%v", err)
 		return 1
 	}
+	restoreTint := termtint.Begin(run.SessionTint)
```
EclipseSourceAI: attach resolves options with an empty target tool, so the tint comes from the default tool's `tool_overrides` and from the cwd project, not from the session being attached ([config.go](https://github.com/eclipse-enclave/enclave/blob/64d00f4a61b34727260d48ed12c3c22190f258a0/internal/config/config.go#L133-L146)). Attaching a codex session while `tool` resolves to claude paints the claude color, which contradicts the per-tool color promise in the new docs section. The container name carries both tool and project hash, so the right layer is resolvable here.

internal/termtint/termtint.go
```diff
@@ -0,0 +1,94 @@
...
+// Callers should defer the returned function; it is safe to call more than
+// once. Begin is a no-op when color is empty, is not an #rrggbb value, when the
+// environment suppresses color, or when stdout is not a terminal.
+func Begin(color string) func() {
+	color = strings.TrimSpace(color)
+	if color == "" {
+		return func() {}
+	}
+	if !colorPattern.MatchString(color) {
+		logx.Warnf("Ignoring invalid session_tint %q: expected an #rrggbb color", color)
```
EclipseSourceAI: Config values are usually validated where the layer is read, with a warning that names the file and the tool override (see `validatePassEnvNames`, [config.go](https://github.com/eclipse-enclave/enclave/blob/64d00f4a61b34727260d48ed12c3c22190f258a0/internal/config/config.go#L252-L258)). Warning there instead would name the offending config file and surface it before the session output takes over the screen, with `Begin` keeping the format check as a guard.

internal/termtint/termtint_test.go
```diff
@@ -0,0 +1,105 @@
...
+			if got := buf.String(); got != "" {
+				t.Fatalf("expected no output, got %q", got)
+			}
+		})
+	}
+}
+
+// ENCLAVE_COLOR=always may only veto a tint, never force one on: the config key
+// stays the single on/off switch.
+func TestColorAlwaysDoesNotForceTint(t *testing.T) {
```
EclipseSourceAI, line 84: With `terminal` false this passes on the tty check, so it would still pass if `always` did force a tint, and it duplicates the "not a terminal" case above. Pass `terminal` true with an empty color to actually cover the invariant the name claims.

docs/configuration.md
```diff
@@ -83,6 +84,22 @@ the inherited mode are ignored. For a regular repository whose `.git`
 directory sits inside the project, the project mount mode governs and this
 option has no effect.
 
+`session_tint` marks the terminal that owns a session, so a sandboxed session
+is visually distinct from an ordinary shell. When set to an `#rrggbb` value,
+`run` (including `shell`, `continue`, and `resume`) and `attach` set the
+terminal background on start and reset it on exit. Inside tmux 3.3 or later only
```
EclipseSourceAI: OSC 111 resets to the terminal's configured default background, not to the background that was in effect before the session, so a color set at runtime (theme switchers, base16-style scripts) is dropped on exit. Worth one sentence here next to the SIGKILL caveat.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Adds a `session_tint` config option that paints the terminal background via OSC 11 while a session owns the terminal and resets it with OSC 111 on exit, plus a signal handler so an interrupted enclave does not leave a stale tint behind. The option plumbing follows the existing generated option registry (defs, sources, golden rows all regenerated and consistent), the new `internal/termtint` package is small and self-contained, and the `logx.ColorSuppressedByEnv` split is a clean way to honor `NO_COLOR`/`ENCLAVE_COLOR=never` without inheriting the stderr auto-detection. `go build ./...`, `go generate ./internal/config/...` and the touched package tests are clean here.

Main points for a maintainer:

- Coverage gap: `enclave exec` also occupies the terminal with an interactive TTY into a running session and gets no tint.
- `attach` resolves the tint against the ambient tool and cwd project rather than the session it attaches to, so the documented per-tool colors can be wrong there.
- Validation of the `#rrggbb` value happens at emit time instead of where config layers are read, which is where the existing warning pattern lives.
- The `TestColorAlwaysDoesNotForceTint` test does not exercise the invariant its name describes.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Follow-up on the `48fa675` fix commit. It addresses every point from my earlier review: `exec` now tints via `termtint.Begin` in `internal/runtime/exec.go`, `attach` resolves the tint through the attached session's own tool and project dir (`attachSessionTint` in `internal/app/command_attach.go`, backed by a new `Inspect` call and covered by new tests), `session_tint` validation moved into `validateSessionTint` in `internal/config/config.go` so the warning names the config file at read time (with `termtint.Begin` keeping a format check as a guard), the `TestColorAlwaysDoesNotForceTint` test now actually exercises the invariant, and the docs got a line about OSC 111 restoring the terminal's default background rather than a prior runtime color.

I checked the resolution logic against `config.ResolveOptionsForTool`/`LoadDefaults` usage elsewhere (`internal/app/app.go`) and it follows the same pattern with the session's tool as target. Verified `exec`'s target always shares the ambient tool via `baseContainerName`, so it doesn't need the same per-session resolution `attach` does. `go build ./...` and the touched package tests are clean. No new issues found in the pushed changes.

- [exec now tints the terminal too](https://github.com/eclipse-enclave/enclave/pull/57#discussion_r3882804800)
- [attach now resolves the tint from the attached session's own tool/project](https://github.com/eclipse-enclave/enclave/pull/57#discussion_r3882804981)
- [session_tint is now validated where the config layer is read](https://github.com/eclipse-enclave/enclave/pull/57#discussion_r3882805224)
- [the ENCLAVE_COLOR=always test now covers the invariant it claims](https://github.com/eclipse-enclave/enclave/pull/57#discussion_r3882805449)
- [docs now cover the OSC 111 default-background reset caveat](https://github.com/eclipse-enclave/enclave/pull/57#discussion_r3882805555)

xai:
Thanks for the nice pr @sgraband, looks good to me and could be approved if rebased! Careful, there will probably be a semantic conflict with containerName which was removed in main iirc.

sgraband:
@xai thanks for the hint! Conflicts are resolved.

xai (approved):
Thanks!

tortmayr:
👍🏼 Great feature @sgraband 
Nice and simple way to improve UX. 
I can confirm that this also works out-of-the-box with herdr.

PR #58, opened 2026-08-25
fix: resolve sessions by name in attach, stop, and theia

planger:
## Why

Containers are named `enclave-<tool>-<project-hash>-<session>`, but `attach`, `stop <name>`, and `theia` passed their positional argument straight to Docker as an exact container name. The flow documented on the website — `enclave --background --name my-task` then `enclave attach my-task` — could therefore never work.

Global uniqueness of session names is deliberately *not* introduced: container names are already unique via the project hash, and reusing `--name review` across several repos is a normal parallel-agent workflow.

## What

- `attach`, `stop <name>`, and `theia`/`theia-next` share one resolver (`internal/app/session_target.go`). A full container name still matches verbatim; anything else is matched as a session name, preferring sessions from the same worktree, then the same project, then the whole host. When a name matches several containers, the candidates are listed instead of one being guessed. `--tool` narrows candidates when given on the CLI.
- `enclave attach` takes an optional argument: with a single running session in the project, bare `attach` picks it (matching `theia`).
- `stop <name>` also considers stopped containers, so a leftover can be cleared by name.
- Session-name normalization moved to `model.SanitizeSessionName`. The `enclave.session` label now stores the sanitized form, so it equals the trailing container-name segment; `ps --name` / `status --name` sanitize their filter input, which fixes `--name "My Task"` not matching `--name my-task`.
- `enclave ps` gained a `SESSION` column (`-` for a project's default container) so the addressable name is discoverable.
- The duplicate-session error now suggests `enclave stop my-task` rather than the full container name.
- Docs updated: `docs/cli-reference.md`, `docs/persistence.md`, `website/docs/docs/cli.md`, and the `attach`/`stop`/`theia` help text.

### Note for consumers

`ps --json` `sessionName` now reports the sanitized value (`my-task`, not `My Task`) for names that needed sanitizing. Field and shape are unchanged. Sessions started by an older build keep their raw label and still resolve, since matching sanitizes both sides.

## How to test

Unit coverage: `internal/app/session_target_test.go`, `internal/model/session_name_test.go`, `internal/runtime/session_name_test.go`, `internal/app/command_ps_test.go`. `make build`, `make test`, `make lint`, and `make check-license-headers` pass; the resolver has not been exercised against live containers.

Manual:

```bash
enclave --background --name my-task
enclave ps                 # NAME shows the container, SESSION shows my-task
enclave attach my-task     # attaches; Ctrl-\ to detach
enclave attach             # single running session in the project is picked
enclave stop my-task
```

Ambiguity and scoping:

```bash
# same name in two different projects
(cd ~/repo-a && enclave --background --name review)
(cd ~/repo-b && enclave --background --name review)
cd ~/repo-a && enclave attach review   # resolves to repo-a's session
cd /tmp && enclave attach review       # lists both candidate container names
```

Sanitization: `enclave --background --name "My Task"` is addressable as `enclave attach my-task`, and `enclave ps --name "My Task"` and `--name my-task` list the same row.

### DIFF
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -44,15 +44,29 @@ started as `--name "My Task"` is reachable as either `my-task` or `"My Task"`.
 The same holds for the `--name` filter of `ps`, `status`, and `stop`.
 
 Session names are resolved against the current project first (same worktree,
-then same project), then across all projects; they are not required to be
-unique, so when one matches several containers the candidates are listed and a
-container name must be passed instead. `--tool` narrows the candidates. Passing
-a container name or ID always resolves verbatim, independently of the working
-directory and `--tool`.
+then same project); they are not required to be unique, so when one matches
+several containers the candidates are listed and a container name must be passed
+instead. `--tool` narrows the candidates. Passing a container name or ID always
+resolves verbatim, independently of the working directory and `--tool`; an
+ambiguous container-ID prefix is reported like an ambiguous name.
+
+`attach` and `theia` widen the search to all projects when the current one has
+no match. `stop <name>` does not: removal is destructive and the auto-assigned
+names `1`, `2`, … collide across projects by construction, so another project's
+session has to be named by its container name or ID.
+
+`stop <name>` and `stop --name <name>` are different commands. The positional
+form removes the single session it resolves, whatever its tool and whether it is
+running or already stopped. `--name` filters the batch form instead: it removes
+every *background* session of the selected tool (`--tool`, default `claude`)
+whose session name matches. A `--name` that is blank or sanitizes to nothing
+matches no session rather than all of them.
 
 With no argument, `attach` picks the single detached session of the current
 project (a foreground session is entered with `exec` instead), and `theia` picks
-its single running container.
+the single running container of the current project. An argument that is present
+but blank (`enclave stop "$SESSION"` with an unset variable) is rejected instead
+of being treated as "no argument".
 
 ### Inspect
 
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -29,22 +29,30 @@ unchanged. See [windows.md](windows.md).
 | `enclave shell --admin` | Shell with limited sudo |
 | `enclave stop [name]` | Stop background containers, or one session by name |
 
-`enclave ps` prints one row per container with its `NAME` (container name) and `SESSION` (the `--name` session name, `-` for the project's default container); either can be passed to `attach`, `stop`, and `theia`. Flags: `--all` (include stopped containers, not just running ones), `--json` (emit a JSON array instead of the table). The flags compose (`ps --all --json`). Each JSON object has the fields `name`, `tool`, `projectDir` (absolute, resolved project path), `projectHash`, `worktree`, `status`, `createdAt` (RFC 3339, empty if unknown), `sessionName`, `background`, and `ports` (array of `{containerPort, hostPort, hostIP, protocol}` bindings).
+`enclave ps` prints one row per container with its `NAME` (container name) and `SESSION` (the `--name` session name, or the auto-assigned `1`, `2`, … for extra sessions; `-` for the project's default container); either can be passed to `attach`, `stop`, and `theia`. Flags: `--all` (include stopped containers, not just running ones), `--json` (emit a JSON array instead of the table). The flags compose (`ps --all --json`). Each JSON object has the fields `name`, `tool`, `projectDir` (absolute, resolved project path), `projectHash`, `worktree`, `status`, `createdAt` (RFC 3339, empty if unknown), `sessionName`, `background`, and `ports` (array of `{containerPort, hostPort, hostIP, protocol}` bindings).
 
 `status` reports sessions of the current project (like `exec`); `--all` widens it to every project. Flags: `--tool` and `--name` filter sessions; `--json` emits one machine-readable snapshot object per session (screen text and OSC title for external state detection). Each snapshot captures the trailing 24 screen rows. See [Session status snapshots](session-status.md).
 
-`enclave attach` flags: `--detach-keys <sequence>` overrides the key sequence for detaching from the session (default `ctrl-\`).
+`enclave attach` flags: `--detach-keys <sequence>` overrides the key sequence for detaching from the session (default `ctrl-\`); `--tool` disambiguates a session name used by more than one tool.
 
-`attach`, `stop <name>`, and `theia`/`theia-next` take either the container name
-from `enclave ps` or the session name passed to `--name` (`enclave --background
---name my-task` → `enclave attach my-task`). Session names are matched in their
-sanitized form (lowercased, non-alphanumerics collapsed to `-`, truncated to 32
-characters), so `--name "My Task"` is addressable as `my-task`. Names are
-resolved against the current project first, then against the whole host;
-`--tool` narrows the candidates. Session names are not required to be unique
-across projects — when one matches several containers, the candidates are listed
-and a container name must be passed instead. With no argument, `attach` and
-`theia` select the single running session of the current project.
+`attach`, `stop <name>`, and `theia`/`theia-next` accept the container name from
+`enclave ps`, a container ID, or the session name passed to `--name`
+(`enclave --background --name my-task` → `enclave attach my-task`). Session
+names are matched in sanitized form on both sides (lowercased,
+non-alphanumerics collapsed to `-`, truncated to 32 characters), so a session
+started as `--name "My Task"` is reachable as either `my-task` or `"My Task"`.
+The same holds for the `--name` filter of `ps`, `status`, and `stop`.
+
+Session names are resolved against the current project first (same worktree,
+then same project), then across all projects; they are not required to be
+unique, so when one matches several containers the candidates are listed and a
+container name must be passed instead. `--tool` narrows the candidates. Passing
+a container name or ID always resolves verbatim, independently of the working
+directory and `--tool`.
+
+With no argument, `attach` picks the single detached session of the current
+project (a foreground session is entered with `exec` instead), and `theia` picks
+its single running container.
 
 ### Inspect
 
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -41,7 +41,9 @@ unchanged. See [windows.md](windows.md).
 names are matched in sanitized form on both sides (lowercased,
 non-alphanumerics collapsed to `-`, truncated to 32 characters), so a session
 started as `--name "My Task"` is reachable as either `my-task` or `"My Task"`.
-The same holds for the `--name` filter of `ps`, `status`, and `stop`.
+The same holds for the `--name` filter of `ps`, `status`, and `stop`; a `--name`
+that is blank or sanitizes to nothing matches no session rather than all of
+them.
 
 Session names are resolved against the current project first (same worktree,
 then same project); they are not required to be unique, so when one matches
@@ -51,16 +53,17 @@ resolves verbatim, independently of the working directory and `--tool`; an
 ambiguous container-ID prefix is reported like an ambiguous name.
 
 `attach` and `theia` widen the search to all projects when the current one has
-no match. `stop <name>` does not: removal is destructive and the auto-assigned
-names `1`, `2`, … collide across projects by construction, so another project's
-session has to be named by its container name or ID.
+no match. `stop` never does, neither for `stop <name>` nor for `stop --name`:
+removal is destructive and the auto-assigned names `1`, `2`, … collide across
+projects by construction, so another project's session has to be named by its
+container name or ID.
 
-`stop <name>` and `stop --name <name>` are different commands. The positional
-form removes the single session it resolves, whatever its tool and whether it is
-running or already stopped. `--name` filters the batch form instead: it removes
-every *background* session of the selected tool (`--tool`, default `claude`)
-whose session name matches. A `--name` that is blank or sanitizes to nothing
-matches no session rather than all of them.
+`stop <name>` and `stop --name <name>` still select differently. The positional
+form removes exactly the one session it resolves, of any tool unless `--tool` is
+passed. `--name` filters the batch form instead: it removes every *background*
+session of the current project whose name matches, for the single tool that the
+options resolve to (so a profile or config `tool` applies as well). Both include
+containers that have already exited.
 
 With no argument, `attach` picks the single detached session of the current
 project (a foreground session is entered with `exec` instead), and `theia` picks
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -22,18 +22,29 @@ unchanged. See [windows.md](windows.md).
 | `enclave resume` | Session picker/list when supported (falls back to `continue`) |
 | `enclave ps` | List enclave containers (`--all` includes stopped, `--json` emits structured output) |
 | `enclave status` | Show terminal snapshots of running sessions |
-| `enclave attach <container>` | Attach to a named background session |
+| `enclave attach [name]` | Attach to a background session by session name or container name |
 | `enclave exec` | Attach to a running container |
 | `enclave exec --admin` | Attach with limited sudo (apt/dpkg) |
 | `enclave shell` | Open an interactive shell in the container |
 | `enclave shell --admin` | Shell with limited sudo |
-| `enclave stop` | Stop background containers |
+| `enclave stop [name]` | Stop background containers, or one session by name |
 
-`enclave ps` flags: `--all` (include stopped containers, not just running ones), `--json` (emit a JSON array instead of the table). The flags compose (`ps --all --json`). Each JSON object has the fields `name`, `tool`, `projectDir` (absolute, resolved project path), `projectHash`, `worktree`, `status`, `createdAt` (RFC 3339, empty if unknown), `sessionName`, `background`, and `ports` (array of `{containerPort, hostPort, hostIP, protocol}` bindings).
+`enclave ps` prints one row per container with its `NAME` (container name) and `SESSION` (the `--name` session name, `-` for the project's default container); either can be passed to `attach`, `stop`, and `theia`. Flags: `--all` (include stopped containers, not just running ones), `--json` (emit a JSON array instead of the table). The flags compose (`ps --all --json`). Each JSON object has the fields `name`, `tool`, `projectDir` (absolute, resolved project path), `projectHash`, `worktree`, `status`, `createdAt` (RFC 3339, empty if unknown), `sessionName`, `background`, and `ports` (array of `{containerPort, hostPort, hostIP, protocol}` bindings).
 
 `status` reports sessions of the current project (like `exec`); `--all` widens it to every project. Flags: `--tool` and `--name` filter sessions; `--json` emits one machine-readable snapshot object per session (screen text and OSC title for external state detection). Each snapshot captures the trailing 24 screen rows. See [Session status snapshots](session-status.md).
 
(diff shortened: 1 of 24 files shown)


planger:
Follow-up commit a37ee43 addresses a subagent review of the first commit. The substantive findings:

**Dangerous regression (introduced by the first commit):** `stop --name "???"` sanitized the query to `""`, and the docker label filter is only applied when non-empty, so the command stopped *and force-removed* every background container of the tool. `--name` filtering for `ps`, `status`, and `stop` now happens in the app layer with both sides sanitized; a name that sanitizes to nothing matches nothing.

**Dropped the session-label change.** The label keeps the user-typed name, since sanitized-both-sides matching makes the stored form irrelevant. This removes the `ps --json` `sessionName` / `status --json` `session_name` value change flagged in the description above (no contract change anymore), and pre-upgrade containers whose label holds e.g. `feature/ABC-123` stay selectable by `--name`.

**Container IDs resolve again.** `attach <id>` / `stop <id>` used to reach Docker directly; the resolver only compared names. IDs (12-char, full, or a prefix of at least 8 hex chars) are matched after session names, so names stay authoritative.

**`--tool` is real now.** The docs claimed it narrowed candidates, but neither `attach` nor `theia` registered the flag — so one `--name` used by two tools in the same project was permanently ambiguous with no escape hatch other than the full container name. Both commands now accept `--tool`, and the tool filter is applied to the candidates instead of to the listing, so an exact container name or ID still resolves regardless of tool.

**No-argument form stays home.** Bare `attach`/`theia` fell back host-wide when the current project had no session, so `cd ~ && enclave attach` could land in another project's agent. Auto-selection is now hard-scoped to the current project, and `attach` restricts it to detached sessions — attaching a second TTY to a foreground session means two terminals fighting over its stdin. Explicit names still resolve host-wide when unambiguous.

Also dropped an unused parameter with a misleading comment, and fixed the doc claims about `--tool`, the `SESSION` column (it also shows auto-assigned `1`, `2`, …), and the stale `attach <container>` row in the website overview table.

Tests added for all of the above (including two that failed first and caught real bugs in the fix: the unsanitizable-name guard and the project-scope fallback). `make build`, `make test`, `make lint`, `make check-license-headers` pass. Still no live-container verification — Docker is unavailable in my environment.

Two review points I did **not** act on, for the record: `theia` still loads preferences for the working-directory project even when an explicit name selects another project's container (pre-existing, and `theia` was always host-wide), and `runAttach`/`runStop`/`runTheia` have no injection seam, so their wiring is covered only at the helper level.

internal/app/command_ps.go
```diff
@@ -110,14 +110,22 @@ func listPSSessions(ctx context.Context, opts model.Options) ([]backend.Session,
...
 }
 
 // psSessionFilterFor builds the backend listing filter from the parsed ps
 // options. By default only running containers are listed; --all includes
-// stopped ones as well.
+// stopped ones as well. The session name is not part of it: names are matched
+// in sanitized form, which the backend cannot express as a label filter.
 func psSessionFilterFor(opts model.Options) backend.SessionFilter {
-	filter := backend.SessionFilter{Tool: psToolFilter(opts), SessionName: psSessionFilter(opts)}
+	filter := backend.SessionFilter{Tool: psToolFilter(opts)}
```
EclipseSourceAI, line 128: With `ps`, `status`, and `stop` all moved to app-layer matching, `SessionFilter.SessionName` has no producers left. It is still declared ([types.go](https://github.com/eclipse-enclave/enclave/blob/64d00f4a61b34727260d48ed12c3c22190f258a0/internal/backend/types.go#L347-L356)) and still turned into a label filter ([docker.go](https://github.com/eclipse-enclave/enclave/blob/64d00f4a61b34727260d48ed12c3c22190f258a0/internal/backend/docker/docker.go#L414-L416)). Dropping it removes the empty-value-means-no-filter footgun that caused the `stop --name "???"` regression instead of leaving it around for the next caller.

internal/app/command_stop.go
```diff
@@ -31,23 +34,38 @@ func runStop(run model.RunOptions) int {
...
+	sessions, err := be.List(context.Background(), backend.SessionFilter{All: true, Background: &background, Tool: run.Tool})
 	if err != nil {
 		logx.Errorf("Failed to list background sessions: %v", err)
 		return 1
 	}
+	// Sanitized matching happens here rather than as a label filter, so that a
+	// name which sanitizes to nothing stops nothing instead of everything.
+	if name := strings.TrimSpace(run.SessionName); name != "" {
+		sessions = sessionsMatchingName(sessions, name)
+	}
```
EclipseSourceAI: `stop <name>` and `stop --name <name>` now behave quite differently: the positional path resolves across every tool, every project, and includes stopped and foreground containers, while this path stays pinned to `run.Tool` (defaults to `claude`) and `Background: true`. Users will read them as the same thing, so either align them or spell the difference out in `docs/cli-reference.md`.

internal/app/session_target.go
```diff
@@ -0,0 +1,276 @@
...
+		}
+	}
+	if len(matches) == 0 {
+		if session, ok := sessionByContainerID(sessions, requested); ok {
+			return session, nil
+		}
+		return backend.Session{}, fmt.Errorf("no enclave session named %q (use `%s ps` to list sessions)", requested, model.AppName)
+	}
+	scoped, _ := projectSessions(matches, q.Project)
+	return selectSingleSession(scoped, requested)
```
EclipseSourceAI: When the current project has no match, this falls through to the host-wide candidates, and a single match elsewhere is used silently. That is fine for `attach`, but `stop` force-removes the container, and auto-assigned session names are `1`, `2`, … per project, so `enclave stop 1` from a project without a session `1` can reach into an unrelated project. Restricting the fallback when `IncludeStopped` is set would be safer.

internal/app/session_target.go
```diff
@@ -0,0 +1,276 @@
...
+		return backend.Session{}, fmt.Errorf("multiple running enclave containers; pass one of these container names explicitly:\n  %s",
+			strings.Join(describeSessions(candidates), "\n  "))
+	}
+}
+
+// projectSessions narrows candidates to the current project: sessions started
+// from the very same worktree first, then any session of the project. ok
+// reports whether the project matched anything, so that callers can decide
+// between falling back to all candidates and reporting nothing.
+func projectSessions(sessions []backend.Session, project model.Project) (scoped []backend.Session, ok bool) {
```
EclipseSourceAI, line 174: `resolveExecTarget` already implements same-worktree-preferred selection with ambiguity reporting for `exec`/`shell` ([exec.go](https://github.com/eclipse-enclave/enclave/blob/64d00f4a61b34727260d48ed12c3c22190f258a0/internal/runtime/exec.go#L54-L127)). Two implementations with different error wording now coexist for the same job. Not asking for the refactor in this PR, but a maintainer should decide whether `exec` should move onto this resolver.

docs/cli-reference.md
```diff
@@ -22,18 +22,37 @@ unchanged. See [windows.md](windows.md).
...
+Session names are resolved against the current project first (same worktree,
+then same project), then across all projects; they are not required to be
+unique, so when one matches several containers the candidates are listed and a
+container name must be passed instead. `--tool` narrows the candidates. Passing
+a container name or ID always resolves verbatim, independently of the working
+directory and `--tool`.
+
+With no argument, `attach` picks the single detached session of the current
+project (a foreground session is entered with `exec` instead), and `theia` picks
+its single running container.
```
EclipseSourceAI: `theia` is project-scoped now too (`autoSelectSession` scopes by project whether or not `BackgroundOnly` is set), so "its single running container" reads as host-wide and is misleading. The `theia` `--help` long text already says "of the current project".

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Makes `attach`, `stop <name>`, and `theia`/`theia-next` resolve their positional argument as a session name (and container ID) instead of passing it verbatim to Docker, so the documented `--background --name my-task` / `attach my-task` flow works. All three share a new resolver in `internal/app/session_target.go`: exact container name first, then sanitized session-name matching scoped current-worktree → project → host, then container ID, with ambiguity reported rather than guessed. `sanitizeSessionName` moved from `runtime` to `model.SanitizeSessionName`, `--name` filtering for `ps`/`status`/`stop` moved into the app layer with both sides sanitized, and `ps` gained a `SESSION` column.

The approach is sound and the follow-up commit already fixed the worst of it (the `stop --name "???"` wipe, container IDs, `--tool` actually being registered). Build and tests pass locally.

- `SessionFilter.SessionName` is now dead in `internal/backend`, and it is exactly the field whose empty-means-no-filter semantics caused the `stop --name "???"` regression. Removing it closes that off.
- `stop <name>` and `stop --name <name>` now diverge on tool scope, background-only, and stopped inclusion. That is easy to trip over.
- The host-wide fallback for explicit names is more consequential for `stop` (which force-removes) than for `attach`, especially with auto-assigned `1`, `2`, … names that collide across projects by construction.
- `resolveExecTarget` in `internal/runtime/exec.go` already does worktree-preferred selection. Two resolvers with different error wording now coexist.

The resolver has not been exercised against live containers, per the PR description. The cross-project and stopped-container paths in particular deserve a manual pass.

internal/app/command_stop.go
```diff
@@ -31,23 +34,38 @@ func runStop(run model.RunOptions) int {
...
 	background := true
-	sessions, err := be.List(context.Background(), backend.SessionFilter{All: true, Background: &background, Tool: run.Tool, SessionName: run.SessionName})
+	sessions, err := be.List(context.Background(), backend.SessionFilter{All: true, Background: &background, Tool: run.Tool})
 	if err != nil {
 		logx.Errorf("Failed to list background sessions: %v", err)
 		return 1
 	}
+	// Sanitized matching happens here rather than as a label filter, so that a
+	// name which sanitizes to nothing stops nothing instead of everything.
+	if name := strings.TrimSpace(run.SessionName); name != "" {
```
xai: An explicitly supplied blank or whitespace-only `--name` skips this filter, leaving the session list unfiltered; the loop below then stops and removes every matching background session. Please distinguish omission from an explicitly blank value and reject it or match nothing. The positional whitespace case should also be rejected rather than treated as auto-selection. Please add command-level regression tests for both forms.

internal/app/session_target.go
```diff
@@ -0,0 +1,276 @@
...
+			matches = append(matches, session)
+		}
+	}
+	if len(matches) == 0 {
+		if session, ok := sessionByContainerID(sessions, requested); ok {
+			return session, nil
+		}
+		return backend.Session{}, fmt.Errorf("no enclave session named %q (use `%s ps` to list sessions)", requested, model.AppName)
+	}
+	scoped, _ := projectSessions(matches, q.Project)
```
xai: For `stop`, falling back to host-wide matches is destructive: names such as `1` collide across projects, so `enclave stop 1` can remove a session from another project. Please keep name-based stop resolution project-scoped; require an exact container name or ID for cross-project removal.

internal/app/session_target.go
```diff
@@ -0,0 +1,276 @@
...
+	}
+	for _, session := range sessions {
+		id := strings.TrimSpace(session.Ref.ID)
+		if id == "" {
+			continue
+		}
+		// Session IDs are truncated, so a full ID pasted from docker is a
+		// superstring of the one recorded here.
+		if strings.HasPrefix(id, requested) || strings.HasPrefix(requested, id) {
+			return session, true
```
xai: This returns the first matching ID prefix, so an ambiguous prefix selects a container based on list order. Please collect matching IDs and report ambiguity instead. Also avoid treating arbitrary extensions of a stored truncated ID as valid matches, and add a collision regression test.

xai (changes requested):
I think we should also address three safety issues related to session selection before merging. See inline comments for details.

planger:
Follow-up commits `f24bf91` and `37bcd0b` address the review.

**`stop` reaching into another project by name** (both reviewers). Fixed, and wider than reported: the positional form is now project-scoped via `sessionTargetQuery.ProjectScopedNames`, and `stop --name` — which listed background sessions of *every* project — is scoped the same way. Scoping only the positional form would have left the exact `enclave stop 1` hazard reachable through the flag. An unresolvable project matches nothing rather than everything. Exact container names and container IDs still resolve host-wide, so cross-project removal remains possible, just not by a colliding name. `attach`/`theia` keep the host-wide fallback: they don't destroy anything.

Note for consumers: `enclave stop --name <name>` no longer removes matching sessions of other projects. Bare `enclave stop` is unchanged (all background containers of the tool, all projects).

**Blank `--name` / blank positional.** `sessionNameFilter(opts)` returns `(name, given)`, so an explicitly blank `--name` matches nothing instead of skipping the filter; `ps`, `status`, and `stop` share it. `requestedSession` rejects a positional argument that is present but blank — that path is what made `enclave stop "$SESSION"` with an unset variable fall through to auto-selection. To keep "not given" distinguishable in `attach`, the CLI now puts the detach keys first in `CmdArgs` (a parse test pins the layout; longer term a `DetachKeys` field on `RunOptions` would remove that positional coupling). `runStop` was split so that `stopSessions(ctx, be, opts, projectDir)` takes a backend: the command-level table covers the positive `--name` case, a same-named session in another project, blank `--name`, unsanitizable `--name`, blank positional, and the unfiltered batch form.

**Ambiguous container-ID prefix.** `sessionsByContainerID` collects every match and reports ambiguity like an ambiguous name. Recorded IDs are truncated to 12 characters, so only a value of exactly the full ID length (64) may extend one — a 20-character "prefix" can no longer piggyback on a real 12-character ID. Tests cover the collision and the partial extension.

**`SessionFilter.SessionName` is dead.** Removed, along with its docker label filter — that empty-means-everything field is what caused the earlier regression.

**`stop <name>` vs `stop --name <name>`.** Not aligned; they really are two selectors, and `--name` is the batch filter shared with `ps`/`status`. Project scope now agrees between them, and the remaining differences are documented in `docs/cli-reference.md` and in the new `stop --help` text: the positional form removes exactly one session of any tool unless `--tool` is passed, while `--name` removes every background session of the *resolved* tool (a profile or config `tool` counts) whose name matches. Both include containers that have already exited.

**`resolveExecTarget` duplication.** Agreed, but not a mechanical merge: `exec`/`shell` synthesize a container name from `--name` rather than searching, list by name prefix, and don't accept container IDs. Only the worktree-preferred selection overlaps, so it is worth its own issue.

**`theia` doc wording.** Fixed; it now reads "the single running container of the current project".

`make build`, `make test`, `make lint`, and `make check-license-headers` pass. Still no live-container verification.

internal/app/command_ps.go
planger: Removed in `f24bf91`: the field and the docker label filter it produced are gone. `ps`, `status`, and `stop` all match names in the app layer now, so nothing was left to feed it.

internal/app/command_stop.go
planger: Kept as two selectors, but the project-scope axis is gone: `37bcd0b` scopes `stop --name` to the current project as well (it previously listed background sessions of every project, which is the same collision hazard as the positional form). The remaining differences — one session vs. all matching, tool from an explicit `--tool` vs. the resolved tool, background-only — are now spelled out in `docs/cli-reference.md` and in a new `stop --help` long text.

internal/app/session_target.go
planger: Noted, not addressed here. Merging them is not mechanical: `exec`/`shell` synthesize the container name from `--name` instead of searching, list by name prefix, and do not accept container IDs — only the worktree-preferred selection overlaps. Worth its own issue.

docs/cli-reference.md
planger: Fixed in `f24bf91`: the sentence now reads "the single running container of the current project".

internal/app/session_target.go
planger: Fixed in `f24bf91`, and extended in `37bcd0b`. Name resolution for `stop` no longer falls back host-wide (`sessionTargetQuery.ProjectScopedNames`); a name that matches nothing in the current project is rejected with a pointer to the container name or ID, which still resolve from anywhere. The same scoping now applies to `stop --name`, which listed background sessions of every project.

internal/app/session_target.go
planger: Done in `f24bf91`: name-based `stop` resolution is project-scoped, and cross-project removal needs an exact container name or ID. `37bcd0b` closes the same hole in `stop --name`, which filtered background sessions of every project — scoping only the positional form would have left `enclave stop --name 1` able to remove another project's session.

internal/app/command_stop.go
planger: Fixed in `f24bf91`. `sessionNameFilter(opts)` returns `(name, given)`, so an explicitly blank `--name` is a filter that matches nothing instead of no filter at all; `ps`, `status`, and `stop` share it. A positional argument that is present but blank is rejected by `requestedSession` rather than treated as auto-selection, and `attach`'s CLI now puts the detach keys first in `CmdArgs` so that an omitted session argument stays distinguishable from an empty one. `runStop` was split so that `stopSessions(ctx, be, opts, projectDir)` takes a backend; the table test in `internal/app/command_stop_test.go` covers blank `--name`, unsanitizable `--name`, the blank positional form, a matching `--name` (positive), a same-named session in another project, and the unfiltered batch form.

internal/app/session_target.go
planger: Fixed in `f24bf91`/`37bcd0b`. `sessionsByContainerID` collects all matches and reports ambiguity through the same path as an ambiguous name, so listing order no longer decides. Recorded IDs are truncated to 12 characters, so only a value of exactly the full ID length (64) may extend one — an arbitrary 13-to-63-character hex string is no longer accepted. Tests cover both the prefix collision and the partial extension.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Follow-up on `f24bf91` and `37bcd0b`. Checked these against the current code and tests, all hold up:

- `stop`'s positional and `--name` forms are now both project-scoped (`ProjectScopedNames`, `projectHashSessions`), closing the cross-project removal hazard for auto-assigned names like `1`.
- Blank `--name` and blank positional arguments are rejected/matched-to-nothing instead of falling back to "no filter" or "auto-select" (`sessionNameFilter`, `requestedSession`), which also fixes the underlying cause of the earlier `stop --name "???"` force-remove regression.
- Ambiguous container-ID prefixes are now collected and reported instead of picked by list order (`sessionsByContainerID` + `selectSingleSession`), and a value can only extend a stored truncated ID when it's exactly full length.
- `SessionFilter.SessionName` and its docker label filter are removed.
- `docs/cli-reference.md`, `docs/persistence.md`, and the new `stop --help` text now spell out how `stop <name>` and `stop --name` differ, and the `theia` project-scoping wording is fixed.

Build and full test suite pass locally. No new issues in the commits since my last pass; the `exec`/`resolveExecTarget` duplication is acknowledged as out of scope with a reasonable explanation (different selection semantics, worth its own issue).

- [dead `SessionFilter.SessionName` field and its label filter](https://github.com/eclipse-enclave/enclave/pull/58#discussion_r3882797460)
- [`stop <name>` vs `stop --name <name>` scope divergence, now documented and aligned on project scope](https://github.com/eclipse-enclave/enclave/pull/58#discussion_r3882797656)
- [host-wide fallback for `stop` being destructive with colliding auto-assigned names](https://github.com/eclipse-enclave/enclave/pull/58#discussion_r3882797784)
- [`theia` doc wording implying host-wide single-container selection](https://github.com/eclipse-enclave/enclave/pull/58#discussion_r3882798046)

xai (approved):
Thanks for the fixes!

The changes look good, and all parts of the test plan (+ some additional testing) execute as expected.

PR #59, opened 2026-08-26
fix: require a serviceAuth entry for every serviceDomains host

planger:
#### What it does

Closes a fail-open gap in extension spec validation.

`network.serviceDomains` maps a host to a service id; `network.serviceAuth.<id>` defines how the gateway injects that credential. `buildSecrets` attaches an HTTP release rule only to ids present in **both** `credentials.sources` and `network.serviceAuth`, and `EffectiveResolver` unions the release hosts into the effective allowlist. A `serviceDomains` entry whose id had no `serviceAuth` entry was therefore inert in two security-relevant ways at once:

- its hosts dropped out of the effective allowlist, and
- the credential was injected as a raw env value instead of a proxy-swapped placeholder,

while `enclave validate-extensions` still reported `ok`. Deleting a single `serviceAuth` line from a spec was enough to trigger both, with no signal.

`validateServiceAuthMappings` now rejects that at load and names the remedy (add the `serviceAuth` entry, or list the hosts under `network.allowedDomains` if they only need to be reachable). The reverse direction — a `serviceAuth` entry with no hosts from either source — was already rejected by `normalizeHosts`, so the `serviceDomains` ↔ `serviceAuth` mapping is now required in both directions.

One related fix: `ListFeatures` swallowed every spec load error, so for a mixin the new load failure would have degraded to "feature silently missing from the built image". It now warns (on stderr, so shell completion stays intact) before skipping.

All built-in specs already satisfy the rule; no extension changes were needed.

#### How to test

`make build && make test && make lint`. The regression test is `TestValidateServiceAuthMappings/serviceDomains_id_without_serviceAuth_fails`.

End-to-end, with a user-authored feature spec:

```bash
TMP=$(mktemp -d)
mkdir -p $TMP/config/enclave/extensions/features/broken-feat
sed '/github-enterprise-token: { headerName/d' extensions/features/github-cli/spec.yaml \
  | sed 's/^name: github-cli/name: broken-feat/' \
  > $TMP/config/enclave/extensions/features/broken-feat/spec.yaml
XDG_CONFIG_HOME=$TMP/config ./bin/enclave validate-extensions   # before: "ok"; now: exits 1 naming the host and service
XDG_CONFIG_HOME=$TMP/config ./bin/enclave features              # now warns instead of silently omitting the feature
```

#### Follow-ups

- The `ListFeatures` skip-warning has no unit test: `specWarn` is called directly rather than injected, and threading a sink through an exported function with nine call sites seemed like more churn than a log line justifies. Verified manually as above.
- `ListFeatures` runs several times per `enclave run`, so the warning repeats. `warnReservedFields` already behaves that way, so this is the existing sink's behavior rather than something new here.

#### Breaking changes

A spec that was previously accepted and silently inert is now a load error. No built-in spec is affected; a third-party sbx kit that uses `serviceDomains` purely for host scoping would need those hosts moved to `network.allowedDomains`.

#### Review checklist

### DIFF
--- a/internal/config/extension.go
+++ b/internal/config/extension.go
@@ -8,6 +8,7 @@
 package config
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -118,7 +119,12 @@ func ListFeatures(paths model.Paths) ([]model.Extension, error) {
 	for _, name := range names {
 		ext, err := LoadFeatureExtension(paths, name)
 		if err != nil {
-			continue // Skip invalid extensions
+			// A feature dropped here is silently missing from the built image,
+			// so surface why. A directory without a spec is not an error.
+			if !errors.Is(err, os.ErrNotExist) {
+				specWarn(fmt.Sprintf("feature %q: %v; skipping", name, err))
+			}
+			continue
 		}
 		features = append(features, ext)
 	}
--- a/internal/config/extension.go
+++ b/internal/config/extension.go
@@ -8,7 +8,6 @@
 package config
 
 import (
-	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -92,25 +91,41 @@ func LoadFeatureExtension(paths model.Paths, name string) (model.Extension, erro
 
 // ListTools returns all tool extension names from both built-in and user extension roots.
 func ListTools(paths model.Paths) ([]string, error) {
-	names, err := listExtensionNames(paths.ToolsDir, paths.UserToolsDir)
+	return listSpecNames(paths, KindSandbox)
+}
+
+// listSpecNames returns the sorted names of the given kind's extensions, from
+// both the built-in and the user root, that carry a spec document. Unlike
+// ListFeatures it never loads those specs, so a name it returns may still fail
+// to load.
+func listSpecNames(paths model.Paths, kind string) ([]string, error) {
+	var builtinDir, userDir string
+	switch kind {
+	case KindSandbox:
+		builtinDir, userDir = paths.ToolsDir, paths.UserToolsDir
+	case KindMixin:
+		builtinDir, userDir = paths.FeaturesDir, paths.UserFeaturesDir
+	default:
+		return nil, fmt.Errorf("unknown extension kind %q", kind)
+	}
+
+	names, err := listExtensionNames(builtinDir, userDir)
 	if err != nil {
 		return nil, err
 	}
 
-	var tools []string
+	var withSpec []string
 	for _, name := range names {
-		if hasSpecFile(paths, name, KindSandbox) {
-			tools = append(tools, name)
+		if hasSpecFile(paths, name, kind) {
+			withSpec = append(withSpec, name)
 		}
 	}
-
-	sort.Strings(tools)
-	return tools, nil
+	return withSpec, nil
 }
 
 // ListFeatures returns all feature extensions from both built-in and user roots, sorted by priority.
 func ListFeatures(paths model.Paths) ([]model.Extension, error) {
-	names, err := listExtensionNames(paths.FeaturesDir, paths.UserFeaturesDir)
+	names, err := listSpecNames(paths, KindMixin)
 	if err != nil {
 		return nil, err
 	}
@@ -120,10 +135,8 @@ func ListFeatures(paths model.Paths) ([]model.Extension, error) {
 		ext, err := LoadFeatureExtension(paths, name)
 		if err != nil {
 			// A feature dropped here is silently missing from the built image,
-			// so surface why. A directory without a spec is not an error.
-			if !errors.Is(err, os.ErrNotExist) {
-				specWarn(fmt.Sprintf("feature %q: %v; skipping", name, err))
-			}
+			// so surface why. Spec-less directories are already filtered out.
+			specWarn(fmt.Sprintf("feature %q: %v; skipping", name, err))
 			continue
 		}
 		features = append(features, ext)
--- a/internal/config/spec_map.go
+++ b/internal/config/spec_map.go
@@ -17,13 +17,18 @@ import (
 
 // validateServiceAuthMappings fails loudly when a network.serviceAuth or
 // network.serviceDomains service id does not map to a declared
-// credentials.sources id, or when a serviceDomains id has no serviceAuth entry.
+// credentials.sources id, and when the serviceDomains ↔ serviceAuth pairing is
+// incomplete in either direction.
+//
 // buildSecrets only builds an HTTP release rule for ids present in both
-// credentials.sources and network.serviceAuth, so either mismatch (a typo, or a
-// dropped serviceAuth line) is otherwise silently inert: the token is injected
-// as a raw env value instead of a proxy-swapped placeholder — a secret-leak
-// risk — and the serviceDomains hosts drop out of the release hosts unioned
-// into the effective allowlist.
+// credentials.sources and network.serviceAuth, so a serviceDomains id with no
+// serviceAuth entry (a typo, or a dropped serviceAuth line) is silently inert:
+// the token is injected as a raw env value instead of a proxy-swapped
+// placeholder — a secret-leak risk — and the serviceDomains hosts drop out of
+// the release hosts unioned into the effective allowlist. The reverse, a
+// serviceAuth entry with no hosts from either source, would otherwise be
+// rejected downstream by normalizeHosts, but with a message that never names
+// serviceDomains; catching it here points both directions at the same remedy.
 func validateServiceAuthMappings(doc specDocument, specPath string) error {
 	if doc.Network == nil {
 		return nil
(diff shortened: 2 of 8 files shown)


internal/config/extension.go
```diff
@@ -118,7 +119,12 @@ func ListFeatures(paths model.Paths) ([]model.Extension, error) {
 	for _, name := range names {
 		ext, err := LoadFeatureExtension(paths, name)
 		if err != nil {
-			continue // Skip invalid extensions
+			// A feature dropped here is silently missing from the built image,
+			// so surface why. A directory without a spec is not an error.
+			if !errors.Is(err, os.ErrNotExist) {
+				specWarn(fmt.Sprintf("feature %q: %v; skipping", name, err))
```
EclipseSourceAI: Warning is good, but `make test` still can't catch this for a built-in feature. I deleted the `github-enterprise-token` line from `extensions/features/github-cli/spec.yaml` and `TestExtensionSurfaceGolden` stayed green: the feature just drops out of `ListFeatures` and nothing asserts its golden ([features leg](https://github.com/eclipse-enclave/enclave/blob/64d00f4a61b34727260d48ed12c3c22190f258a0/internal/config/conformance_test.go#L58-L66)). Doing the same edit to `extensions/tools/codex/spec.yaml` fails loudly, because the tools leg enumerates names and loads each one ([tools leg](https://github.com/eclipse-enclave/enclave/blob/64d00f4a61b34727260d48ed12c3c22190f258a0/internal/config/conformance_test.go#L38-L54)). Making the features leg iterate `listExtensionNames` + `LoadFeatureExtension` would close that.

internal/config/spec_map.go
```diff
@@ -40,6 +43,9 @@ func validateServiceAuthMappings(doc specDocument, specPath string) error {
 		if _, ok := sources[id]; !ok {
 			return fmt.Errorf("%s: network.serviceDomains[%q] references service %q with no matching credentials.sources entry", specPath, host, id)
 		}
+		if _, ok := doc.Network.ServiceAuth[id]; !ok {
+			return fmt.Errorf("%s: network.serviceDomains[%q] references service %q with no matching network.serviceAuth entry (add one, or list the hosts under network.allowedDomains instead)", specPath, host, id)
```
EclipseSourceAI, line 67: This flips a previously accepted (if inert) sbx kit spec into a hard load error, and for a `kind: sandbox` kit that means the tool stops running entirely, not just losing the header injection. AGENTS.md asks for breaking changes to external-consumer contracts to be coordinated and the breaking-changes box in the description is unchecked, so a maintainer should confirm hard fail over a load warning here.

internal/config/spec_map.go
```diff
@@ -40,6 +43,9 @@ func validateServiceAuthMappings(doc specDocument, specPath string) error {
 		if _, ok := sources[id]; !ok {
 			return fmt.Errorf("%s: network.serviceDomains[%q] references service %q with no matching credentials.sources entry", specPath, host, id)
 		}
+		if _, ok := doc.Network.ServiceAuth[id]; !ok {
```
EclipseSourceAI, line 66: The reverse direction is required but not nearly as legible. A `serviceAuth` entry with no `hosts` and no `serviceDomains` reference bottoms out in `secrets["tok"].release.http: hosts must contain at least one domain pattern` ([here](https://github.com/eclipse-enclave/enclave/blob/64d00f4a61b34727260d48ed12c3c22190f258a0/internal/config/secrets.go#L264-L267)), which never mentions `serviceDomains`. Catching that case here too would make both directions point at the same remedy.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Adds one validation rule to `validateServiceAuthMappings`: a `network.serviceDomains` host whose service id has no `network.serviceAuth` entry is now a load error. The reasoning checks out against the code. `buildSecrets` only attaches an HTTP release rule for ids present in both `credentials.sources` and `network.serviceAuth`, and `serviceDomains` has no other consumer, so such an entry really is inert in both directions the description claims (hosts never reach the effective allowlist, credential goes into the container as a raw env value). All built-in specs already satisfy the rule, and the docs updates in `docs/extensions/README.md` and `adding-a-tool.md` match the implemented behavior.

The `ListFeatures` change is a sensible companion: without it a mixin hitting the new error would have degraded to "feature silently missing from the image".

Points for a maintainer:

- The golden conformance test does not cover the feature side of this. A built-in feature spec that fails the new rule is skipped by `ListFeatures` and `make test` stays green, while the equivalent break in a tool spec fails loudly. Verified both locally.
- Whether the hard load error is the right call for third-party sbx kits that use `serviceDomains` for host scoping only. The description documents the break but leaves the breaking-changes checkbox unchecked.
- The reverse direction is enforced only indirectly, via a generic `hosts must contain at least one domain pattern` error that never names `serviceDomains`.

Scope is tight, no drive-by changes, and the new test sits with the existing subtests.

internal/config/extension.go
planger: Fixed in 35152b5. The features leg of `TestExtensionSurfaceGolden` now enumerates spec names and loads each one, so a spec that fails to load fails the test instead of dropping out of the snapshot set. I pulled the name-plus-`hasSpecFile` filter out of `ListTools` into `listSpecNames` and used it on both legs; `ListFeatures` uses it too, which filters spec-less directories before the load and lets the `os.ErrNotExist` carve-out around the new warning go away.

Verified with your reproduction. Deleting the `github-enterprise-token` line from `extensions/features/github-cli/spec.yaml` now gives:

```
--- FAIL: TestExtensionSurfaceGolden
    conformance_test.go:71: LoadFeatureExtension(github-cli): .../extensions/features/github-cli/spec.yaml: network.serviceDomains["*.ghe.com"] references service "github-enterprise-token" with no matching network.serviceAuth entry (add one, or list the hosts under network.allowedDomains instead)
```

internal/config/spec_map.go
planger: Fixed in 35152b5. `validateServiceAuthMappings` now also rejects a `serviceAuth` entry that gets no hosts from either source:

```
network.serviceAuth["github-enterprise-token"] has no hosts to release the credential to (add a hosts list, or map hosts to this service under network.serviceDomains)
```

Two follow-on adjustments. The check runs ahead of `normalizeHosts`, so `TestLoadProfileRejectsSecretReleaseWithEmptyHosts` now asserts the new message, and `TestValidateAndNormalizeSecretConfigsEmptyReleaseHosts` covers the `normalizeHosts` guard directly to keep it from going untested. The unknown-id checks for both maps also run before both pairing checks now, otherwise a typo like `serviceDomains: { ghe.com: github-tokn }` would report the hostless `serviceAuth` entry it happens to create rather than the typo itself.

internal/config/spec_map.go
planger: Keeping the hard fail, deliberately.

A load warning would preserve exactly the behaviour this PR exists to remove. With no `serviceAuth` entry the secret gets no release rule, and `auth_manager.go` then puts the real token into the container environment as a raw value instead of a proxy-swapped placeholder, while the hosts stay off the effective allowlist. So the spec was already not doing what it asked for, and it was leaking the credential while doing it. A warning printed during `enclave run` would scroll past.

It is also the treatment the loader already gives this class of authoring mistake: unknown keys under `UnmarshalStrict`, `validateProxyManaged`, `validateEntrypointArgv`, and root `commands.startup` all fail at load rather than warn.

On blast radius: no built-in spec is affected, and for features `ListFeatures` warns and skips, so a broken kit degrades rather than blocking the session. A `kind: sandbox` kit does stop running, but its spec is authored by whoever invokes the tool, the remedy is one line, and the error names it. I am fine with that trade as maintainer.

Note that the reverse-direction check added in 35152b5 does not widen this: it only re-messages specs `normalizeHosts` already rejected.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Follow-up review. The only functional change since my last pass is 35152b5, which addresses both open points from my first review:

- The reverse-direction gap (a `serviceAuth` entry with no hosts from either source) is now caught in `validateServiceAuthMappings` with a message naming `serviceDomains`, instead of bottoming out in the generic `normalizeHosts` error. Unknown-id checks run first so a typo'd id is reported at the right spot, verified with the new `spec_map_test.go` subtests.
- The features leg of `TestExtensionSurfaceGolden` now enumerates spec names via the extracted `listSpecNames` and loads each one directly, so a built-in feature spec that fails to load now fails the test instead of silently dropping out of `ListFeatures`. Confirmed this actually catches the repro from my earlier comment.

I re-ran `go build ./...` and `go test ./internal/config/...` locally, both pass, and read through the new validation ordering and tests; no issues found in the pushed changes.

The remaining open thread on `spec_map.go` (hard load error vs. warning for `kind: sandbox` kits) got a detailed rebuttal from the author defending the hard fail. That reasoning holds up against the code, so I'm not re-raising it; it's a call for a maintainer to sign off on, not a defect.

- [reverse-direction serviceAuth-without-hosts now names serviceDomains as the remedy](https://github.com/eclipse-enclave/enclave/pull/59#discussion_r3882784918)
- [feature golden test now loads each spec instead of silently skipping broken ones](https://github.com/eclipse-enclave/enclave/pull/59#discussion_r3882784604)

xai (approved):
Changes are clean and work as described! Thank you @planger!

PR #60, opened 2026-08-28
feat: install tools and features from git repositories

planger:
#### What it does

Adds an installer so tools and features can come from a git repository instead of
only shipping in the box.

`enclave tools|features add <source>` fetches a repository with a blobless sparse
checkout, stages it in a temp directory, validates and sanitizes the tree, and
swaps it into `~/.config/enclave/extensions/`. `update` re-fetches from the
recorded source, `remove` deletes, and `list` now reports built-in and installed
extensions side by side with where each one came from.

Because an extension runs code at image build and container start, `add` and
`update` print a capability summary distilled from the staged content before
writing anything: root install steps, install and startup scripts, allowlist
changes, declared credentials, the approval-bypass flag, skills, host config
passthrough, and files seeded into the project. `update` shows that summary as a
diff against what is installed, so a newly granted capability is visible before
you accept it. `--yes` skips the prompt, never the summary.

Each install records provenance in `.enclave-source.json` (remote, ref, commit,
tree hash, installer version). The sidecar is excluded from the build context and
the image identity hash, so re-pinning to a new commit with identical content does
not force a rebuild. The tree hash is what detects hand edits, and directories
without a sidecar are treated as unmanaged.

Docs are in `docs/extensions/installing.md`, with a shorter version on the website
under CLI commands.

Three commits, by layer: spec summaries and path resolution, then the
`internal/extinstall` package, then the CLI and app wiring.

#### How to test

The example repo is https://github.com/eclipse-enclave/enclave-extensions. It has
two `kind: sandbox` tools, `openclaw` and `dsh`, and no features. Its content is
still in review as eclipse-enclave/enclave-extensions#1, so until that merges the
commands below need `--ref initial-extensions`.

Build with `make build` and use `./bin/enclave`.

List, and note the source column:

```bash
enclave tools list
enclave features list
```

Install both tools. Without `--all` this lists the two candidates and asks which
one you meant:

```bash
enclave tools add eclipse-enclave/enclave-extensions
enclave tools add eclipse-enclave/enclave-extensions --all
```

Read the capability summary before confirming. `openclaw` and `dsh` both run an
install script and widen the allowlist, so there should be something to see. Then
check `enclave tools list` and `cat ~/.config/enclave/extensions/tools/openclaw/.enclave-source.json`.

Remove and add a single one back:

```bash
enclave tools remove dsh
enclave tools add eclipse-enclave/enclave-extensions --name dsh
```

Wrong kind, which should point you at the other verb rather than say "nothing
found":

```bash
enclave features add eclipse-enclave/enclave-extensions
```

For `update`, install from a local clone so you can create a change to pull:

```bash
git clone -b initial-extensions https://github.com/eclipse-enclave/enclave-extensions /tmp/ext
enclave tools remove openclaw dsh
enclave tools add /tmp/ext --all

# widen a capability, e.g. append a conf-file= line to
# /tmp/ext/tools/dsh/gateway-allowlist.conf, then commit it
git -C /tmp/ext commit -am "test: widen allowlist"

enclave tools update dsh
```

The summary should show that line as an added allowlist directive against what is
installed. `enclave tools update` with no name updates everything.

Worth poking at as well: `--dry-run` writes nothing, editing an installed file by
hand makes `update` refuse until `--force`, and `--json` requires `--yes`.

Finally, start a session with an installed tool (`enclave --tool dsh`) to confirm
the install triggers a rebuild and the extension actually works in the container.

#### Follow-ups

- The `owner/repo` shorthand is two segments, so GitLab subgroup repos need a full
  clone URL plus `--path`. Documented, not fixed.
- No version constraints or lockfile. You pin a ref, and `update` follows it.
- The website page covers the basics only. It links nothing yet for authoring an
  extension to install.
- #69: a sidecar that cannot be read blocks `add --force`, so recovery means
  `remove --force` first.

#### Breaking changes

`tools list` and `features list` gain fields under `--json`, but the envelope stays
`schemaVersion: "1"` and existing fields are unchanged.

#### Review checklist

### DIFF
--- a/internal/config/spec_summary.go
+++ b/internal/config/spec_summary.go
@@ -1,0 +1,228 @@
+// Copyright (C) 2026 EclipseSource GmbH and others.
+//
+// This program and the accompanying materials are made available under the
+// terms of the MIT License, which is available in the project root.
+//
+// SPDX-License-Identifier: MIT
+
+package config
+
+import (
+	"fmt"
+	"os"
+	"sort"
+	"strings"
+
+	"enclave/internal/model"
+)
+
+// SpecInitFileSummary describes one commands.initFiles entry.
+type SpecInitFileSummary struct {
+	Path          string
+	OnlyIfMissing bool
+}
+
+// SpecCredentialSource describes one credentials.sources entry: which env
+// aliases it fills, which host file (if any) it can also be read from, and —
+// the most consequential grant in the schema — whether it is released as an
+// HTTP header to a set of hosts via network.serviceDomains/serviceAuth.
+type SpecCredentialSource struct {
+	ID            string
+	Env           []string
+	FilePath      string
+	FileParser    string
+	ReleaseHeader string
+	ReleaseHosts  []string
+}
+
+// SpecProviderSummary projects one providers[] entry: the auth files and
+// session detection it declares, the OAuth callback ports it opens, and the
+// env var (if any) it uses to locate its credential-storage directory.
+type SpecProviderSummary struct {
+	Name                string
+	AuthFiles           []string
+	AuthSession         bool
+	AuthSessionMode     string
+	OAuthPorts          []string
+	SecurestorageDirEnv string
+}
+
+// SpecHostExposure collects the fields that expose host filesystem paths or
+// host credential files to the container: sandbox.* host-passthrough fields,
+// and the mixin-native top-level configDir/authFiles pair used by
+// enclave-native auth features (e.g. github-cli).
+type SpecHostExposure struct {
+	PassthroughPaths    []string
+	HostConfigDir       string
+	HostCredentialsFile string
+	HostOAuthJSON       string
+	MixinConfigDir      string
+	MixinAuthFiles      []string
+}
+
+// SpecSummary projects the capability-relevant fields of a spec document. It
+// exists so callers outside this package (the extension installer) can report
+// what an extension will be able to do without duplicating the on-disk schema,
+// which specDocument owns.
+type SpecSummary struct {
+	Kind                  string
+	Name                  string
+	DisplayName           string
+	Description           string
+	NeedsRoot             bool
+	AptPackages           []string
+	InstallCommands       int
+	InstallCommandsAsRoot int
+	StartupCommands       []string
+	InitFiles             []SpecInitFileSummary
+	AllowedDomains        []string
+	DeniedDomains         []string
+	Ports                 []int
+	CredentialEnv         []string
+	CredentialSources     []SpecCredentialSource
+	ProxyManaged          []string
+	EnvironmentVars       []string
+	EntrypointOverride    string
+	Providers             []SpecProviderSummary
+	HostExposure          SpecHostExposure
+	YoloFlag              string
+	YoloEnabled           bool
+	DefaultEnabled        *bool
+	DefaultIncluded       *bool
+}
+
+// SummarizeSpecDir reads dir's spec.yaml (falling back to spec.json) and
+// projects it onto a SpecSummary. It returns os.ErrNotExist when neither file
+// is present, and a parse error for a malformed, unknown-field, or
+// schemaVersion-mismatched document.
+func SummarizeSpecDir(dir string) (SpecSummary, error) {
+	specPath, ok := ownSpecFile(dir)
+	if !ok {
+		return SpecSummary{}, fmt.Errorf("no %s in %s: %w", SpecFilename, dir, os.ErrNotExist)
+	}
+
+	doc, err := parseSpecDocument(specPath)
+	if err != nil {
+		return SpecSummary{}, err
+	}
+	return summarizeSpecDocument(doc), nil
+}
+
+func summarizeSpecDocument(doc specDocument) SpecSummary {
+	summary := SpecSummary{
+		Kind:            strings.TrimSpace(doc.Kind),
+		Name:            strings.TrimSpace(doc.Name),
+		DisplayName:     strings.TrimSpace(doc.DisplayName),
+		Description:     strings.TrimSpace(doc.Description),
+		NeedsRoot:       doc.NeedsRoot,
+		AptPackages:     doc.AptPackages,
+		DefaultEnabled:  doc.DefaultEnabled,
+		DefaultIncluded: doc.DefaultIncluded,
+	}
+	if doc.Commands != nil {
+		summary.InstallCommands = len(doc.Commands.Install)
+		for _, cmd := range doc.Commands.Install {
+			if IsRootInstallUser(cmd.User) {
+				summary.InstallCommandsAsRoot++
+			}
+		}
+		for _, cmd := range doc.Commands.Startup {
+			summary.StartupCommands = append(summary.StartupCommands, describeSpecCommand(cmd))
+		}
+		for _, file := range doc.Commands.InitFiles {
+			summary.InitFiles = append(summary.InitFiles, SpecInitFileSummary{
+				Path:          strings.TrimSpace(file.Path),
+				OnlyIfMissing: file.OnlyIfMissing,
+			})
+		}
+	}
+	if doc.Network != nil {
+		summary.AllowedDomains = doc.Network.AllowedDomains
+		summary.DeniedDomains = doc.Network.DeniedDomains
+	}
+	if doc.Environment != nil {
+		summary.ProxyManaged = doc.Environment.ProxyManaged
+		for name := range doc.Environment.Variables {
+			summary.EnvironmentVars = append(summary.EnvironmentVars, name)
+		}
+		sort.Strings(summary.EnvironmentVars)
+	}
+	if doc.Credentials != nil {
+		hostsByService := serviceHostsByID(doc)
+		ids := make([]string, 0, len(doc.Credentials.Sources))
+		for id := range doc.Credentials.Sources {
+			ids = append(ids, id)
+		}
+		sort.Strings(ids)
+		var aliases []string
+		for _, id := range ids {
+			source := doc.Credentials.Sources[id]
+			cs := SpecCredentialSource{ID: id, Env: append([]string(nil), source.Env...)}
+			if source.File != nil {
+				cs.FilePath = source.File.Path
+				cs.FileParser = source.File.Parser
+			}
+			if doc.Network != nil {
+				if auth, ok := doc.Network.ServiceAuth[id]; ok {
+					cs.ReleaseHeader = auth.HeaderName
+					cs.ReleaseHosts = hostsByService[id]
+				}
+			}
+			summary.CredentialSources = append(summary.CredentialSources, cs)
+			aliases = append(aliases, source.Env...)
+		}
+		summary.CredentialEnv = sortDedupeStrings(aliases)
+	}
+	for _, port := range doc.Ports {
+		summary.Ports = append(summary.Ports, port.Container)
+	}
+	for _, p := range doc.Providers {
+		ps := SpecProviderSummary{
+			Name:                strings.TrimSpace(p.Name),
+			AuthFiles:           append([]string(nil), p.AuthFiles...),
+			SecurestorageDirEnv: p.SecurestorageDirEnv,
+		}
+		if p.AuthSession != nil {
+			ps.AuthSession = true
+			ps.AuthSessionMode = p.AuthSession.Mode
+		}
+		for _, op := range p.OAuthPorts {
+			ps.OAuthPorts = append(ps.OAuthPorts, op.Port)
+		}
+		summary.Providers = append(summary.Providers, ps)
+	}
+	if doc.Sandbox != nil {
+		if doc.Sandbox.Entrypoint != nil {
+			summary.EntrypointOverride = entrypointCommand(doc.Sandbox.Entrypoint)
+		}
+		summary.HostExposure.PassthroughPaths = doc.Sandbox.PassthroughPaths
+		summary.HostExposure.HostConfigDir = doc.Sandbox.HostConfigDir
+		summary.HostExposure.HostCredentialsFile = doc.Sandbox.HostCredentials
+		summary.HostExposure.HostOAuthJSON = doc.Sandbox.HostOAuthJSON
+		summary.YoloFlag = strings.TrimSpace(doc.Sandbox.YoloFlag)
+		summary.YoloEnabled = model.YoloEnabledValue(doc.Sandbox.YoloEnabled)
+	}
+	summary.HostExposure.MixinConfigDir = doc.ConfigDir
+	summary.HostExposure.MixinAuthFiles = doc.AuthFiles
+	return summary
+}
+
+// describeSpecCommand renders a startup command for display. specCommand.Command
+// is string or []string depending on the sbx stage, so both shapes are handled.
+func describeSpecCommand(cmd specCommand) string {
+	if description := strings.TrimSpace(cmd.Description); description != "" {
+		return description
+	}
+	switch value := cmd.Command.(type) {
+	case string:
+		return value
+	case []interface{}:
+		parts := make([]string, 0, len(value))
+		for _, item := range value {
+			parts = append(parts, fmt.Sprint(item))
+		}
+		return strings.Join(parts, " ")
+	default:
+		return "(command)"
+	}
+}
--- a/internal/config/spec_summary.go
+++ b/internal/config/spec_summary.go
@@ -8,18 +8,34 @@
 package config
 
 import (
+	"crypto/sha256"
+	"encoding/hex"
 	"fmt"
+	"maps"
 	"os"
-	"sort"
+	"slices"
 	"strings"
 
 	"enclave/internal/model"
 )
 
-// SpecInitFileSummary describes one commands.initFiles entry.
+// SpecInitFileSummary describes one commands.initFiles entry. kit-init.sh
+// writes the entry's content at the entry's mode on every container start, so
+// both belong here: ContentDigest stands in for the content itself, which is
+// arbitrarily long and is a file to inspect rather than a capability to read.
 type SpecInitFileSummary struct {
 	Path          string
+	Mode          string
 	OnlyIfMissing bool
+	ContentDigest string
+}
+
+// SpecPortSummary describes one ports[] entry. Publish is carried separately
+// from the port number because it, not the number, decides whether the port is
+// bound on the host.
+type SpecPortSummary struct {
+	Container int
+	Publish   bool
 }
 
 // SpecCredentialSource describes one credentials.sources entry: which env
@@ -64,6 +80,12 @@ type SpecHostExposure struct {
 // exists so callers outside this package (the extension installer) can report
 // what an extension will be able to do without duplicating the on-disk schema,
 // which specDocument owns.
+//
+// The command-bearing fields carry the command text rather than a count, so a
+// caller diffing two summaries notices a step that was rewritten in place.
+// InstallCommands additionally prefixes "(root) " to a step that runs as root,
+// which makes a step changing user visible even when the root count does not
+// move.
 type SpecSummary struct {
 	Kind                  string
 	Name                  string
@@ -71,18 +93,21 @@ type SpecSummary struct {
 	Description           string
 	NeedsRoot             bool
 	AptPackages           []string
-	InstallCommands       int
+	InstallCommands       []string
 	InstallCommandsAsRoot int
 	StartupCommands       []string
 	InitFiles             []SpecInitFileSummary
 	AllowedDomains        []string
 	DeniedDomains         []string
-	Ports                 []int
+	Ports                 []SpecPortSummary
 	CredentialEnv         []string
 	CredentialSources     []SpecCredentialSource
 	ProxyManaged          []string
 	EnvironmentVars       []string
 	EntrypointOverride    string
+	ContinueArgs          string
+	ResumeArgs            string
+	PostStartOpenIDE      string
 	Providers             []SpecProviderSummary
 	HostExposure          SpecHostExposure
 	YoloFlag              string
@@ -120,19 +145,23 @@ func summarizeSpecDocument(doc specDocument) SpecSummary {
 		DefaultIncluded: doc.DefaultIncluded,
 	}
 	if doc.Commands != nil {
-		summary.InstallCommands = len(doc.Commands.Install)
 		for _, cmd := range doc.Commands.Install {
+			text := specCommandText(cmd)
 			if IsRootInstallUser(cmd.User) {
 				summary.InstallCommandsAsRoot++
+				text = "(root) " + text
 			}
+			summary.InstallCommands = append(summary.InstallCommands, text)
 		}
 		for _, cmd := range doc.Commands.Startup {
-			summary.StartupCommands = append(summary.StartupCommands, describeSpecCommand(cmd))
+			summary.StartupCommands = append(summary.StartupCommands, specCommandText(cmd))
 		}
 		for _, file := range doc.Commands.InitFiles {
 			summary.InitFiles = append(summary.InitFiles, SpecInitFileSummary{
 				Path:          strings.TrimSpace(file.Path),
+				Mode:          strings.TrimSpace(file.Mode),
 				OnlyIfMissing: file.OnlyIfMissing,
+				ContentDigest: shortContentDigest(file.Content),
 			})
 		}
 	}
@@ -142,20 +171,17 @@ func summarizeSpecDocument(doc specDocument) SpecSummary {
 	}
 	if doc.Environment != nil {
 		summary.ProxyManaged = doc.Environment.ProxyManaged
-		for name := range doc.Environment.Variables {
-			summary.EnvironmentVars = append(summary.EnvironmentVars, name)
+		// Reported as NAME=value: the value is the capability. Setting
+		// NODE_TLS_REJECT_UNAUTHORIZED to 0 is not the same grant as setting it
+		// to 1, and a summary keyed on names alone cannot tell them apart.
+		for _, name := range slices.Sorted(maps.Keys(doc.Environment.Variables)) {
+			summary.EnvironmentVars = append(summary.EnvironmentVars, name+"="+doc.Environment.Variables[name])
 		}
-		sort.Strings(summary.EnvironmentVars)
 	}
 	if doc.Credentials != nil {
 		hostsByService := serviceHostsByID(doc)
-		ids := make([]string, 0, len(doc.Credentials.Sources))
-		for id := range doc.Credentials.Sources {
-			ids = append(ids, id)
-		}
-		sort.Strings(ids)
 		var aliases []string
-		for _, id := range ids {
+		for _, id := range slices.Sorted(maps.Keys(doc.Credentials.Sources)) {
 			source := doc.Credentials.Sources[id]
 			cs := SpecCredentialSource{ID: id, Env: append([]string(nil), source.Env...)}
 			if source.File != nil {
@@ -174,7 +200,7 @@ func summarizeSpecDocument(doc specDocument) SpecSummary {
 		summary.CredentialEnv = sortDedupeStrings(aliases)
 	}
 	for _, port := range doc.Ports {
-		summary.Ports = append(summary.Ports, port.Container)
+		summary.Ports = append(summary.Ports, SpecPortSummary{Container: port.Container, Publish: port.Publish})
 	}
 	for _, p := range doc.Providers {
 		ps := SpecProviderSummary{
@@ -191,10 +217,18 @@ func summarizeSpecDocument(doc specDocument) SpecSummary {
 		}
 		summary.Providers = append(summary.Providers, ps)
 	}
+	if doc.PostStart != nil {
+		summary.PostStartOpenIDE = strings.TrimSpace(doc.PostStart.OpenIDE)
+	}
 	if doc.Sandbox != nil {
 		if doc.Sandbox.Entrypoint != nil {
 			summary.EntrypointOverride = entrypointCommand(doc.Sandbox.Entrypoint)
 		}
+		// continueArgs/resumeArgs are appended to the agent's argv for
+		// `--continue`/`--resume`, so they grant whatever those flags grant —
+		// including approval bypass — without going through sandbox.yoloFlag.
+		summary.ContinueArgs = strings.Join(doc.Sandbox.ContinueArgs, " ")
+		summary.ResumeArgs = strings.Join(doc.Sandbox.ResumeArgs, " ")
 		summary.HostExposure.PassthroughPaths = doc.Sandbox.PassthroughPaths
 		summary.HostExposure.HostConfigDir = doc.Sandbox.HostConfigDir
 		summary.HostExposure.HostCredentialsFile = doc.Sandbox.HostCredentials
@@ -207,12 +241,23 @@ func summarizeSpecDocument(doc specDocument) SpecSummary {
 	return summary
 }
 
-// describeSpecCommand renders a startup command for display. specCommand.Command
-// is string or []string depending on the sbx stage, so both shapes are handled.
-func describeSpecCommand(cmd specCommand) string {
-	if description := strings.TrimSpace(cmd.Description); description != "" {
-		return description
+// shortContentDigest identifies a body of text by a truncated SHA-256, so a
+// summary can report that it changed without carrying it. An empty body has no
+// digest: there is nothing to identify.
+func shortContentDigest(content string) string {
+	if content == "" {
+		return ""
 	}
+	sum := sha256.Sum256([]byte(content))
+	return hex.EncodeToString(sum[:])[:12]
+}
+
+// specCommandText renders a command's argv as one string. specCommand.Command
+// is string or []string depending on the sbx stage, so both shapes are handled.
+// The author's description is deliberately not used in its place: an update
+// diff has to react to the shell that actually runs, and a description can stay
+// identical while the command underneath it is rewritten.
+func specCommandText(cmd specCommand) string {
 	switch value := cmd.Command.(type) {
 	case string:
 		return value
--- a/internal/config/spec_summary.go
+++ b/internal/config/spec_summary.go
@@ -119,14 +119,17 @@ type SpecSummary struct {
 // SummarizeSpecDir reads dir's spec.yaml (falling back to spec.json) and
 // projects it onto a SpecSummary. It returns os.ErrNotExist when neither file
 // is present, and a parse error for a malformed, unknown-field, or
-// schemaVersion-mismatched document.
+// schemaVersion-mismatched document. dir may be extension content that nothing
+// has sanitized yet — the installer classifies a fetched repository's
+// candidates before it copies anything — so the document is read under the
+// untrusted-source rules.
 func SummarizeSpecDir(dir string) (SpecSummary, error) {
 	specPath, ok := ownSpecFile(dir)
 	if !ok {
 		return SpecSummary{}, fmt.Errorf("no %s in %s: %w", SpecFilename, dir, os.ErrNotExist)
 	}
 
-	doc, err := parseSpecDocument(specPath)
+	doc, err := parseUntrustedSpecDocument(specPath)
 	if err != nil {
 		return SpecSummary{}, err
 	}
--- a/internal/config/validate_extensions.go
+++ b/internal/config/validate_extensions.go
@@ -40,20 +40,34 @@ func ValidateExtensions(paths model.Paths) (ExtensionValidation, error) {
 	if err := validateBuiltinFeatureExtensions(paths, &result); err != nil {
 		return result, err
 	}
-	if err := validateUserToolExtensions(paths, &result); err != nil {
+	if err := validateUserExtensions(paths, model.KindTool, &result); err != nil {
 		return result, err
 	}
-	if err := validateUserFeatureExtensions(paths, &result); err != nil {
+	if err := validateUserExtensions(paths, model.KindFeature, &result); err != nil {
 		return result, err
 	}
 
 	return result, nil
 }
 
+// ValidateExtensionDir validates the single user extension directory dir as an
+// extension of kind. dir's parent stands in for the kind's user extension root,
+// so built-ins still resolve from paths while nothing else is scanned: that is
+// what lets the extension installer check a staged copy, before the swap,
+// exactly as it would be checked once installed. No built-in pass runs beside
+// it, so spec content is validated even when dir shadows a built-in.
+func ValidateExtensionDir(dir string, kind model.ExtensionKind, paths model.Paths) ExtensionValidation {
+	result := ExtensionValidation{}
+	scoped := withUserExtensionRoot(paths, kind, filepath.Dir(dir))
+	validateUserExtension(scoped, kind, filepath.Base(dir), true, &result)
+	return result
+}
+
 // hasOwnSpecFile reports whether extDir itself (no user-override resolution)
 // contains a spec.yaml or spec.json.
 func hasOwnSpecFile(extDir string) bool {
-	return util.PathExists(filepath.Join(extDir, SpecFilename)) || util.PathExists(filepath.Join(extDir, SpecFilenameJSON))
+	_, ok := ownSpecFile(extDir)
+	return ok
 }
 
 func validateBuiltinToolExtensions(paths model.Paths, result *ExtensionValidation) error {
@@ -63,7 +77,7 @@ func validateBuiltinToolExtensions(paths model.Paths, result *ExtensionValidatio
 	}
 
 	for _, entry := range entries {
-		if !entry.IsDir() {
+		if !isExtensionDir(entry) {
 			continue
 		}
 		name := entry.Name()
@@ -89,11 +103,14 @@ func validateBuiltinToolExtensions(paths model.Paths, result *ExtensionValidatio
 	return nil
 }
 
-func validateUserToolExtensions(paths model.Paths, result *ExtensionValidation) error {
-	if strings.TrimSpace(paths.UserToolsDir) == "" {
+// validateUserExtensions validates every extension in the user extension root
+// of kind, one directory at a time.
+func validateUserExtensions(paths model.Paths, kind model.ExtensionKind, result *ExtensionValidation) error {
+	_, userRoot := ExtensionRoots(paths, kind)
+	if strings.TrimSpace(userRoot) == "" {
 		return nil
 	}
-	entries, err := os.ReadDir(paths.UserToolsDir)
+	entries, err := os.ReadDir(userRoot)
 	if err != nil {
 		if os.IsNotExist(err) {
 			return nil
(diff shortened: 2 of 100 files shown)


internal/extinstall/remove.go
```diff
@@ -0,0 +1,79 @@
...
+	inventory, err := Inventory(env.Paths, req.Kind, nil, InventoryProvenance)
+	if err != nil {
+		return nil, err
+	}
+
+	results := make([]ActionResult, 0, len(req.Names))
+	for _, name := range req.Names {
+		entry, ok := inventory[name]
+		if !ok {
+			return nil, fmt.Errorf("%s %q is not installed", req.Kind.Label(), name)
```
EclipseSourceAI, line 34: Bailing out here throws away the results of the extensions already removed in this loop, so the run reports a failure for work it actually did. Verified: `enclave tools remove beta nope --yes --json` deletes `beta` from disk but emits only `{"name":"","action":"failed","error":"tool \"nope\" is not installed"}` - a consumer of the result envelope has no way to learn `beta` is gone. `updateTargets` validates every name up front before touching anything; `Remove` should do the same, or record a failed `ActionResult` and continue.

internal/extinstall/fetch.go
```diff
@@ -0,0 +1,493 @@
...
+			result.Commit = fields[0]
+		}
+	}
+	if result.Commit == "" {
+		return RemoteRef{}, fmt.Errorf("%s has no HEAD", RedactRemote(remote))
+	}
+	if result.Ref == "" {
+		// A remote HEAD that is not symbolic (rare) still gives a usable commit.
+		result.Ref = "HEAD"
+		return result, nil
```
EclipseSourceAI: A remote whose HEAD is not symbolic records `ref: "HEAD"` with `refType: "branch"`, and `update` can never resolve that again: `ResolveRef` sees a non-empty, non-SHA ref and runs `ls-remote refs/heads/HEAD refs/tags/HEAD`, which matches nothing. Reproduced with a local clone in detached HEAD state (a common way to install from a locally checked-out tag): `enclave tools update demo` fails with `/tmp/extsrc has no branch or tag "HEAD"` forever, with no recovery short of re-adding. Since the commit is already known here, recording `RefTypeCommit` with the resolved SHA would make update a correct no-op instead.

internal/extinstall/install.go
```diff
@@ -0,0 +1,445 @@
...
+	}
+	if len(match) == 0 {
+		if len(other) > 0 {
+			return nil, fmt.Errorf("%s contains %d %s extension(s) but no %s; use `enclave %s add`",
+				src.Display(), len(other), req.Kind.Other().Label(), req.Kind.Label(), req.Kind.Other().Verb())
+		}
+		return nil, fmt.Errorf("%s contains no %s extension", src.Display(), req.Kind.Label())
+	}
+
+	if err := detectNameCollisions(req.Kind, match); err != nil {
```
EclipseSourceAI: This runs over the whole `match` set before `selectExtensions` narrows it, so a duplicate-name pair anywhere in the repository blocks installing an unrelated extension by name. Verified: a repo with `tools/alpha` + `nested/alpha` and a third `tools/beta` rejects `add <src> --name beta` with `multiple tool extensions named "alpha" found`. Moving the check after selection (or restricting it to the selected set) would keep the error where it matters.

internal/util/util.go
```diff
@@ -283,6 +283,17 @@ func RedactSecret(value string) string {
 	return "****"
 }
 
+// NonEmptyLines splits text into trimmed lines, dropping blank ones.
+func NonEmptyLines(text string) []string {
```
EclipseSourceAI: `NonEmptyLines` is a straight move of `internal/docker`'s private `splitLines` ([main](https://github.com/eclipse-enclave/enclave/blob/64d00f4a61b34727260d48ed12c3c22190f258a0/internal/docker/container.go#L185-L193)) and its only four callers are still in `internal/docker`. Nothing in `extinstall` uses it, so this is an unrelated refactor riding along in an already large PR.

internal/extinstall/fetch.go
```diff
@@ -0,0 +1,493 @@
...
+	// fullSHAPattern, a ref name that passed validateRefName (rejecting a
+	// leading "-" and git's other unsafe ref characters), or a repository
+	// path that passed normalizeSubpath and a leading-"-" check in
+	// Materialize; no caller-controlled value reaches git unvalidated.
+	cmd := exec.CommandContext(ctx, f.git, full...)
+	cmd.Dir = dir
+	var stdoutBuf, stderrBuf bytes.Buffer
+	cmd.Stdout = &stdoutBuf
+	cmd.Stderr = &stderrBuf
+	cmd.Env = append(os.Environ(), "GIT_ADVICE=0")
```
EclipseSourceAI: Nothing sets `GIT_TERMINAL_PROMPT=0` here. Under `--yes --json` against a private remote git will still try to read credentials from `/dev/tty` and block, which is exactly the mode that is supposed to be non-interactive. Setting it when `Request.Interactive` is false turns a hang into a clean failure.

internal/extinstall/narrate.go
```diff
@@ -0,0 +1,235 @@
...
+	if !s.Color || color == "" || text == "" {
+		return text
+	}
+	return string(color) + text + string(logx.ColorReset)
+}
+
+// section opens a block for one extension: a rule carrying the name on the
+// left and a caption such as `tool @ 8f0bbc7` on the right. When the two do not
+// fit on one line the caption moves below the rule rather than being truncated.
+func (e Env) section(style Style, name string, caption string) {
```
EclipseSourceAI: `section`, `outcome`, `note` and `summarize` all take a `style Style` parameter although `Env` already carries `Style`, and every call site passes `env.Style`. `unchanged` just below reads `e.Style` directly, which is the shape the rest should have too.

internal/config/validate_extensions.go
```diff
@@ -102,48 +119,74 @@ func validateUserToolExtensions(paths model.Paths, result *ExtensionValidation)
...
+// of the same name exists, its spec content is validated only if
+// validateOverrideContent is set: the full-tree pass leaves it off because its
+// built-in pass already checked the same merged spec.
+func validateUserExtension(paths model.Paths, kind model.ExtensionKind, name string, validateOverrideContent bool, result *ExtensionValidation) {
+	_, userRoot := ExtensionRoots(paths, kind)
+	userDir := filepath.Join(userRoot, name)
+	builtinDir, _ := ResolveExtensionDirs(paths, kind, name)
+	hasBuiltin := builtinDir != ""
+
+	if util.DirExists(filepath.Join(userDir, "go")) {
```
EclipseSourceAI: This PR adds `model.ExtensionGoDir` and uses it in `extinstall.scanExtensionTree`, but the literal `"go"` stays here. Use the constant so the two places that decide what a `go/` directory means cannot drift.

internal/extinstall/capabilities_render.go
```diff
@@ -0,0 +1,132 @@
...
+	row("allowlist domains", joinWithPlus(c.AllowDomains))
+	row("allowlist directives", strings.Join(c.AllowlistDirectives, ", "))
+	row("denied domains", strings.Join(c.Spec.DeniedDomains, ", "))
+	row("ports", strings.Join(c.allPorts(), ", "))
+	envAliases := strings.Join(c.Spec.CredentialEnv, ", ")
+	if envAliases != "" {
+		envAliases += " (env alias)"
+	}
+	row("credentials", envAliases)
+	for _, source := range c.Spec.CredentialSources {
```
EclipseSourceAI: The loop variable `source` shadows the `source string` parameter rendered at the top of this function. Harmless today, but rename it (`credSource`) so a later edit that wants the original does not silently get the wrong one.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Adds `enclave tools|features add|update|remove`, backed by a new `internal/extinstall` package: blobless sparse git fetch, staged copy under strict limits (no symlinks, 1000 files / 10 MiB, name charset), a capability summary rendered before the swap, an atomic rename into `~/.config/enclave/extensions/`, and a `.enclave-source.json` provenance sidecar excluded from the build context and the image identity hash. `internal/config` is refactored so tool and feature paths share one kind-parameterised set of helpers, and `SpecSummary` is the new projection the installer reads capabilities from.

The design holds up well. Staging-then-swap with `.replaced-*` unwind, the tree hash as the hand-edit detector, redaction on read as well as on write, and validating remote-supplied ref names before they reach git as bare operands are all the right calls. I built the branch and exercised add/update/remove/list against local repos; the happy paths and the `--json` envelopes behave as documented.

Critical points for a maintainer:

- **`Remove` loses successful results** when a later name is unknown. Verified on the built binary: the extension is deleted from disk but the `--json` envelope reports only the failure. This is the one that breaks the documented integration contract.
- **Detached remote HEAD is recorded as `ref: "HEAD"`, `refType: "branch"`**, which no later `ls-remote` can resolve, so `update` fails permanently for that install. Reachable from a local clone sitting on a tag.
- **`detectNameCollisions` runs before selection**, so an unrelated duplicate pair in the repo blocks `--name <other>`.
- `GIT_TERMINAL_PROMPT` is not disabled, so a private remote under `--yes --json` can block on a credential prompt.

The rest are smaller: an unrelated `splitLines` -> `util.NonEmptyLines` move in `internal/docker` that no new code uses, a `"go"` literal that should be the constant this PR introduces, a redundant `style Style` parameter on the `Env` narration methods, and one shadowed variable. Docs (`docs/extensions/installing.md`) are unusually complete and match the implementation.

internal/extinstall/remove.go
planger: Good catch, fixed. `Remove` now checks every name against the inventory before it deletes anything, the way `updateTargets` does, so a run that fails on an unknown name has removed nothing.

I kept the other two refusals (a built-in, or a directory enclave did not install) as per-extension failures rather than hoisting them as well: those come back as a `failed` `ActionResult` in the envelope, so nothing is lost, and `remove a b c` should still remove what it can when `b` happens to be a built-in. Covered by `TestRemoveUnknownNameRemovesNothing`.

internal/extinstall/fetch.go
planger: Agreed, fixed. A HEAD that is not symbolic offers no branch to follow, so the install records the commit with `refType: "commit"` and `update` treats it as a pin instead of hunting for a branch called `HEAD`.

Worth knowing: `update` on such a source is now a quiet no-op where it used to fail loudly, and re-pinning takes `add --force`. Covered by `TestGitFetcherResolveRefFromDetachedClone`, and documented in `docs/extensions/installing.md`.

internal/extinstall/install.go
planger: Agreed, fixed. `detectNameCollisions` now considers only the names being installed. `--all` and the interactive path still check the whole set, and `--name foo` with two `foo` directories still errors, since the check stays ahead of `selectExtensions` and its last-wins `byName` map. Covered by `TestAddNamedExtensionIgnoresUnrelatedCollision`.

internal/util/util.go
planger: Fair, reverted. `splitLines` is private in `internal/docker` again and `NonEmptyLines` is gone; the only `internal/util` addition left on the branch is `DirExists`, which `internal/config` uses.

internal/extinstall/fetch.go
planger: Fixed, with one change to the condition. Gating on `Interactive` would also catch `--yes` at a terminal, where a human can still answer git, so the gate is `--json`: the mode whose whole report is an envelope nobody is watching.

`GIT_TERMINAL_PROMPT=0` alone also turned out to be too little. Git asks an askpass program first, and a VS Code terminal exports `GIT_ASKPASS`, so the dialog would open before git ever reached the terminal. An empty `GIT_ASKPASS` goes with it and ends that search. Credential helpers still answer; an ssh key passphrase is ssh's own prompt and can still appear. Both are documented, and `TestGitFetcherEnvGovernsTerminalPrompts` covers the environment.

internal/extinstall/narrate.go
planger: Agreed, dropped. All four read `e.Style` now. `capabilities.render` keeps its `Style` parameter since it is not an `Env` method.

internal/config/validate_extensions.go
planger: Fixed, it uses `model.ExtensionGoDir`.

internal/extinstall/capabilities_render.go
planger: Fixed, renamed to `cred`.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

All seven points from the earlier review are fixed on the current head (dcea80e..403581a).

Verified in code:
- `Remove` now validates every name against the inventory before deleting anything (`internal/extinstall/remove.go`), covered by `TestRemoveUnknownNameRemovesNothing`.
- A HEAD that names no branch now records `RefTypeCommit` instead of the unresolvable `"HEAD"` ref.
- `detectNameCollisions` only checks names actually being installed.
- The `internal/util.NonEmptyLines` move is reverted, `splitLines` is back in `internal/docker`.
- `GIT_TERMINAL_PROMPT=0` and an empty `GIT_ASKPASS` are set for `--json` runs (gated correctly on `--json`, not `Interactive`, so a `--yes` terminal run can still answer git).
- `Env`'s narration methods no longer take a redundant `Style` param.
- `validate_extensions.go` uses `model.ExtensionGoDir` instead of the `"go"` literal, and the shadowed `source` var in `capabilities_render.go` is renamed.

Also reviewed the one substantive change pushed since the earlier review that wasn't part of those threads: `93a2940` fixes `parseSymrefHead` matching `refs/remotes/origin/HEAD` by suffix instead of requiring an exact `HEAD` match, which mis-recorded the ref when installing from a local clone. The fix and its new tests (`TestParseSymrefHeadIgnoresRemoteTrackingHead`, `TestGitFetcherResolveRefFromLocalClone`) look correct. Build and `go test ./internal/extinstall/... ./internal/docker/... ./internal/util/... ./internal/config/...` pass. No new findings.

- [Remove no longer loses successful results when a later name is unknown](https://github.com/eclipse-enclave/enclave/pull/60#discussion_r3882811119)
- [Detached/non-symbolic remote HEAD now pins to the commit instead of recording an unresolvable "HEAD" ref](https://github.com/eclipse-enclave/enclave/pull/60#discussion_r3882811324)
- [detectNameCollisions now only checks the names being installed](https://github.com/eclipse-enclave/enclave/pull/60#discussion_r3882811507)
- [Unrelated splitLines/NonEmptyLines move reverted out of the PR](https://github.com/eclipse-enclave/enclave/pull/60#discussion_r3882811687)
- [GIT_TERMINAL_PROMPT/GIT_ASKPASS now suppress credential prompts under --json](https://github.com/eclipse-enclave/enclave/pull/60#discussion_r3882811845)
- [Redundant Style parameter dropped from Env narration methods](https://github.com/eclipse-enclave/enclave/pull/60#discussion_r3882811990)
- ["go" literal replaced with model.ExtensionGoDir](https://github.com/eclipse-enclave/enclave/pull/60#discussion_r3882812173)
- [Shadowed loop variable in capabilities_render.go renamed](https://github.com/eclipse-enclave/enclave/pull/60#discussion_r3882812317)

planger:
The prepared extensions repo is available: https://github.com/eclipse-enclave/enclave-extensions
I'll move everything there, before merging this PR.

internal/extinstall/fetch.go
```diff
@@ -0,0 +1,519 @@
...
+	var stdoutBuf, stderrBuf bytes.Buffer
+	cmd.Stdout = &stdoutBuf
+	cmd.Stderr = &stderrBuf
+	cmd.Env = f.env()
+	if runErr := cmd.Run(); runErr != nil {
+		detail := strings.TrimSpace(stderrBuf.String())
+		if detail == "" {
+			detail = runErr.Error()
+		}
+		return "", stderrBuf.String(), fmt.Errorf("git %s: %s", strings.Join(args, " "), RedactRemote(detail))
```
sgraband: `RedactRemote` covers `detail` but not `strings.Join(args, " ")`, and `args` carries the remote for `ls-remote` and `remote add origin`. So `enclave tools add "https://user:TOKEN@host/repo" --json --yes` puts the token straight into the envelope's `error` field. [The doc comment above](https://github.com/eclipse-enclave/enclave/blob/403581a6ac3216e5e556977f23a0233ab4107207/internal/extinstall/fetch.go#L452-L457) promises the opposite.

internal/extinstall/fetch.go
```diff
@@ -0,0 +1,519 @@
...
+		clean, err := normalizeSubpath(dir)
+		if err != nil {
+			return fmt.Errorf("materialize path %q: %w", dir, err)
+		}
+		if strings.HasPrefix(clean, "-") {
+			return fmt.Errorf("materialize path %q looks like a command-line option, refusing", dir)
+		}
+		normalized = append(normalized, clean)
+	}
+	if len(normalized) > 0 {
```
sgraband, line 443: `sparse-checkout disable` only runs inside this branch, but [`fetchCache`](https://github.com/eclipse-enclave/enclave/blob/403581a6ac3216e5e556977f23a0233ab4107207/internal/extinstall/fetch.go#L97-L109) shares one repo across every extension from the same remote and ref. A root extension updated after a subpath one checks out into the stale cone — the rest of the tree is dropped, `validateStaged` still passes because the root `spec.yaml` survives, and `TreeHash` records the truncated content as the new baseline. Hoisting the disable out of the `if` would cover it.

internal/config/spec_summary.go
```diff
@@ -0,0 +1,228 @@
...
+		for _, op := range p.OAuthPorts {
+			ps.OAuthPorts = append(ps.OAuthPorts, op.Port)
+		}
+		summary.Providers = append(summary.Providers, ps)
+	}
+	if doc.Sandbox != nil {
+		if doc.Sandbox.Entrypoint != nil {
+			summary.EntrypointOverride = entrypointCommand(doc.Sandbox.Entrypoint)
+		}
+		summary.HostExposure.PassthroughPaths = doc.Sandbox.PassthroughPaths
```
sgraband, line 235: `continueArgs` and `resumeArgs` aren't projected here, so they reach neither the summary nor the diff. They're [appended to the agent argv](https://github.com/eclipse-enclave/enclave/blob/403581a6ac3216e5e556977f23a0233ab4107207/internal/runtime/command_builder.go#L40-L48) one line below the `yoloFlag` path that `yoloActive` does model, so `continueArgs: ["--dangerously-skip-permissions"]` with no `yoloFlag` installs with the `skips approval` row empty. A flag in `sandbox.entrypoint.args` gets reported, this one doesn't.

internal/config/spec_summary.go
```diff
@@ -0,0 +1,228 @@
...
+		Name:            strings.TrimSpace(doc.Name),
+		DisplayName:     strings.TrimSpace(doc.DisplayName),
+		Description:     strings.TrimSpace(doc.Description),
+		NeedsRoot:       doc.NeedsRoot,
+		AptPackages:     doc.AptPackages,
+		DefaultEnabled:  doc.DefaultEnabled,
+		DefaultIncluded: doc.DefaultIncluded,
+	}
+	if doc.Commands != nil {
+		summary.InstallCommands = len(doc.Commands.Install)
```
sgraband: `InstallCommands` is a count, so an update that rewrites `commands.install[0].command` and keeps the entry count [diffs to nothing](https://github.com/eclipse-enclave/enclave/blob/403581a6ac3216e5e556977f23a0233ab4107207/internal/extinstall/capabilities_diff.go#L35-L37). Pad a comment to match the byte count and the update prints `no capability changes` for the field that runs shell at build time. Startup has the same hole through `describeSpecCommand` preferring the author's `description` over the command.

internal/config/spec_summary.go
```diff
@@ -0,0 +1,228 @@
...
+					cs.ReleaseHeader = auth.HeaderName
+					cs.ReleaseHosts = hostsByService[id]
+				}
+			}
+			summary.CredentialSources = append(summary.CredentialSources, cs)
+			aliases = append(aliases, source.Env...)
+		}
+		summary.CredentialEnv = sortDedupeStrings(aliases)
+	}
+	for _, port := range doc.Ports {
```
sgraband, line 205: `publish` is dropped here, so `diffList("port", ...)` compares bare numbers and a `publish: false → true` flip on update reports nothing — that's the field that decides whether it [binds on the host](https://github.com/eclipse-enclave/enclave/blob/403581a6ac3216e5e556977f23a0233ab4107207/internal/runtime/runtime.go#L594-L602). Worth carrying into the summary so a published port reads differently, imho.

internal/config/spec_summary.go
```diff
@@ -0,0 +1,228 @@
...
+			summary.EntrypointOverride = entrypointCommand(doc.Sandbox.Entrypoint)
+		}
+		summary.HostExposure.PassthroughPaths = doc.Sandbox.PassthroughPaths
+		summary.HostExposure.HostConfigDir = doc.Sandbox.HostConfigDir
+		summary.HostExposure.HostCredentialsFile = doc.Sandbox.HostCredentials
+		summary.HostExposure.HostOAuthJSON = doc.Sandbox.HostOAuthJSON
+		summary.YoloFlag = strings.TrimSpace(doc.Sandbox.YoloFlag)
+		summary.YoloEnabled = model.YoloEnabledValue(doc.Sandbox.YoloEnabled)
+	}
+	summary.HostExposure.MixinConfigDir = doc.ConfigDir
```
sgraband, line 242: `doc.PostStart` isn't projected. `openIDE` [launches a program on the host](https://github.com/eclipse-enclave/enclave/blob/403581a6ac3216e5e556977f23a0233ab4107207/internal/runtime/background.go#L74-L92) after start and forces the session detached, so it seems worth a line in the summary imho.

internal/extinstall/stage.go
```diff
@@ -0,0 +1,181 @@
...
+	if !util.PathStrictlyWithin(s.kindDir, final) {
+		return "", fmt.Errorf("refusing to install outside %s", s.kindDir)
+	}
+	staged := s.extDir(name)
+	replaced := filepath.Join(s.kindDir, fmt.Sprintf("%s%s-%d-%d", replacedPrefix, name, os.Getpid(), replacedSwapSeq.Add(1)))
+
+	err := util.WithFileLock(env.lockPath(), func() error {
+		hadExisting := false
+		if _, statErr := os.Stat(final); statErr == nil {
+			if renameErr := os.Rename(final, replaced); renameErr != nil {
```
sgraband, line 140: `os.Rename` keeps the directory's mtime, so `.replaced-*` carries the original install time and [`sweepStale`](https://github.com/eclipse-enclave/enclave/blob/403581a6ac3216e5e556977f23a0233ab4107207/internal/extinstall/stage.go#L86-L90) measures the 24h window from that. For anything installed more than a day ago the recovery copy is stale as soon as it's created, and the next `add`/`update` of that kind removes it before doing anything else. [`installing.md`](https://github.com/eclipse-enclave/enclave/blob/403581a6ac3216e5e556977f23a0233ab4107207/docs/extensions/installing.md?plain=1#L207-L212) says the window is 24 hours. `os.Chtimes` after the rename would fix it.

internal/extinstall/fetch.go
```diff
@@ -0,0 +1,519 @@
...
+	}
+	r.checkoutTarget = "FETCH_HEAD"
+
+	// Preferred path: commit and trees only. A server that rejects the filter
+	// outright fails this fetch (non-zero exit); one that merely does not
+	// support it (e.g. uploadpack.allowFilter off) instead answers with a
+	// full fetch anyway and only warns on stderr. Both mean the filter did
+	// not apply, so both are detected before trusting r.blobless.
+	_, filterStderr, err := r.fetcher.runCombined(ctx, r.dir, "fetch", "--depth", "1", "--no-tags", "--filter=blob:none", "origin", fetchRef)
+	if err == nil && !filterIgnoredByServer(filterStderr) {
```
sgraband: Both `err == nil` arms return the same call and only differ in setting `r.blobless`, which is never read outside `fetch_test.go`. Could merge them.

internal/extinstall/narrate.go
```diff
@@ -0,0 +1,236 @@
...
+// left and a caption such as `tool @ 8f0bbc7` on the right. When the two do not
+// fit on one line the caption moves below the rule rather than being truncated.
+func (e Env) section(name string, caption string) {
+	w := e.narrate()
+	style := e.Style
+	width := style.width()
+	nameCells := ruleLeadIn + 1 + displayWidth(name) + 1 // "━━ name "
+	painted := style.paint(name, logx.ColorBold)
+
+	if caption == "" {
```
sgraband: Both callers pass a non-empty caption (`install.go:312`, `remove.go:62`), so this branch is dead.

internal/extinstall/capabilities_diff.go
```diff
@@ -0,0 +1,147 @@
...
+		return []string{fmt.Sprintf("adds %d %s(s) (%d -> %d)", after-before, label, before, after)}
+	case after < before:
+		return []string{fmt.Sprintf("drops %d %s(s) (%d -> %d)", before-after, label, before, after)}
+	default:
+		return nil
+	}
+}
+
+func diffList(label string, before []string, after []string) []string {
+	beforeSet := map[string]struct{}{}
```
sgraband: Two maps for membership over a handful of domains — `slices.Contains` on each side would do it, imho.

internal/extinstall/capabilities.go
```diff
@@ -0,0 +1,376 @@
...
+		switch rel {
+		case model.InstallScriptFilename:
+			tree.InstallScript = true
+		case model.CheckUpdateScriptFilename:
+			tree.UpdateProbe = true
+		}
+		if name, ok := immediateChild(rel, skillsPrefix); ok {
+			tree.Skills = append(tree.Skills, name)
+		}
+		for _, sub := range entrypoints {
```
sgraband: `entrypointDirs` returns one element and its own comment says neither kind runs the other's directory, so the `map[string][]string` and the two loops over it could be a single string and one slice.

internal/extinstall/capabilities.go
```diff
@@ -0,0 +1,376 @@
...
+	}
+	defer func() { _ = file.Close() }()
+
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		line := strings.TrimSpace(scanner.Text())
+		if line == "" || strings.HasPrefix(line, "#") {
+			continue
+		}
+		matched := false
```
sgraband: A line can only start with one of the two prefixes, so the loop and the `matched` flag aren't buying anything. `strings.CutPrefix` twice, then `strings.Cut(rest, "/")` — `CutPrefix` is already used at line 189.

internal/extinstall/update.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+	match, _, skipped := selectKind(classify(repo.Dir(), candidates), req.Kind)
+	if len(match) == 0 {
+		reason := "no matching extension"
+		if len(skipped) > 0 {
+			reason = skipped[0].Skip
+		}
+		return ActionResult{}, fmt.Errorf("%s: %s", src.Display(), reason)
+	}
+	var selected *classified
```
sgraband: `slices.IndexFunc` over `match` would drop the pointer, which only exists to carry not-found.

internal/extinstall/result.go
```diff
@@ -0,0 +1,42 @@
...
+	Action   string   `json:"action"`
+	Commit   string   `json:"commit,omitempty"`
+	Path     string   `json:"path,omitempty"`
+	Warnings []string `json:"warnings,omitempty"`
+	Error    string   `json:"error,omitempty"`
+}
+
+// HasFailure reports whether any extension failed, so a caller can pick a
+// non-zero exit code while still keeping the successes.
+func HasFailure(results []ActionResult) bool {
```
sgraband, line 37: `slices.ContainsFunc(results, func(r ActionResult) bool { return r.Action == ActionFailed })`, one line.

internal/config/spec_summary.go
```diff
@@ -0,0 +1,228 @@
...
+			})
+		}
+	}
+	if doc.Network != nil {
+		summary.AllowedDomains = doc.Network.AllowedDomains
+		summary.DeniedDomains = doc.Network.DeniedDomains
+	}
+	if doc.Environment != nil {
+		summary.ProxyManaged = doc.Environment.ProxyManaged
+		for name := range doc.Environment.Variables {
```
sgraband: `slices.Sorted(maps.Keys(...))` for this and the credential sources below. `install.go` already uses that idiom.

internal/extinstall/filediff.go
```diff
@@ -0,0 +1,81 @@
...
+
+// printChangedFiles renders the file-level diff, capping the listed paths at
+// maxChangedFilesShown and printing a remainder count beyond that so a large
+// change stays readable.
+func printChangedFiles(w io.Writer, changes []changedFile) {
+	if len(changes) == 0 {
+		return
+	}
+	_, _ = fmt.Fprintf(w, "%s%d file(s) changed:\n", bodyIndent, len(changes))
+	shown := changes
```
sgraband: `shown := changes[:min(len(changes), maxChangedFilesShown)]` with the builtin `min`.

internal/model/extension_kind.go
```diff
@@ -0,0 +1,62 @@
...
+// DirName is the subdirectory of an extension root holding this kind.
+func (k ExtensionKind) DirName() string {
+	if k == KindTool {
+		return "tools"
+	}
+	return "features"
+}
+
+// Verb is the CLI parent command for this kind.
+func (k ExtensionKind) Verb() string { return k.DirName() }
```
sgraband: `Verb()` and `DirName()` have the same body. One accessor for one string is enough imho, the call sites read fine either way.

sgraband (changes requested):
I went over the installer package, the git driver and the capability summary separately.

The staging-and-swap design holds up — I walked every failure point of `commit` and the unwind is correct in all of them, including the cross-process races. Test coverage in `internal/extinstall` is good, the two structural guards on the capability fields especially.

Two things I'd want fixed first, both in the capability summary, because that's the part the whole feature rests on. `continueArgs` isn't projected, so an extension can skip per-action approval with the `skips approval` row empty. And `commands.install` is diffed as a count, so an update that rewrites the build-time shell prints `no capability changes`.

The third is `runCombined` putting the unredacted argv into the error, which puts a token from a remote URL into the `--json` envelope. One-liner.

The other comments are just minor cleanups.

internal/extinstall/fetch.go
planger: Fixed, and this was the one worth catching. Both halves of the error are redacted now, so `ls-remote` and `remote add origin` can no longer put a token from a remote URL into the message — or into the `--json` envelope the doc comment promises it stays out of.

`TestGitFetcherErrorRedactsRemoteInArgv` pins it against an unknown subcommand, which fails without a network round trip while still carrying the remote through as an operand.

internal/extinstall/fetch.go
planger: Agreed, fixed. `sparse-checkout disable` is hoisted out of the `if`, so the full checkout always starts from a repository that is not sparse — whether the cone came from a half-applied attempt just above, or from an earlier `Materialize` on the repository the fetch cache is sharing.

`TestMaterializeAfterSparseSelectionRestoresFullTree` reproduces the sequence you describe (a subpath extension, then a root one out of the same cached repo) and fails on the old code with exactly the truncated tree.

internal/config/spec_summary.go
planger: Agreed, fixed. `Ports` is `[]SpecPortSummary{Container, Publish}`, and a published port renders and diffs differently from a merely declared one:

```
  ports  8080 (published on the host), 9090
```

One consequence worth knowing: a published port and a provider `oauthPort` with the same number no longer collapse into one entry, since only one of them carries the marker. That seemed the right trade — they are not the same grant.

internal/config/spec_summary.go
planger: Agreed, fixed. `postStart.openIDE` gets a row and a diff entry, phrased around what it does rather than the field name:

```
  post-start  theia (launched on the host once the container is running;
              the session runs detached)
```

internal/extinstall/fetch.go
planger: Agreed, merged: `r.blobless = !filterIgnoredByServer(filterStderr)` and one return.

internal/extinstall/narrate.go
planger: Agreed, dropped. Both callers pass a non-empty caption.

internal/extinstall/capabilities.go
planger: Agreed. It is `entrypointDir(kind) string` now; the `map[string][]string` and both loops over it are gone, and the scripts land straight in `tree.StartupScripts` and get sorted once with the rest.

internal/extinstall/update.go
planger: Agreed, `slices.IndexFunc` with an index check.

internal/model/extension_kind.go
planger: Agreed, `Verb` is gone and `DirName`'s doc names both roles.

I had kept the alias as a seam in case the CLI verb and the directory name ever diverge, but that is a speculative future paying for a real duplicate today, and as you say the three call sites read fine either way.

internal/extinstall/fetch.go
planger: Fixed, and this was the one worth catching. Both halves of the error are redacted now, so `ls-remote` and `remote add origin` can no longer put a token from a remote URL into the message — or into the `--json` envelope the doc comment promises it stays out of.

`TestGitFetcherErrorRedactsRemoteInArgv` pins it against an unknown subcommand, which fails without a network round trip while still carrying the remote through as an operand.

internal/extinstall/fetch.go
planger: Agreed, fixed. `sparse-checkout disable` is hoisted out of the `if`, so the full checkout always starts from a repository that is not sparse — whether the cone came from a half-applied attempt just above, or from an earlier `Materialize` on the repository the fetch cache is sharing.

`TestMaterializeAfterSparseSelectionRestoresFullTree` reproduces the sequence you describe (a subpath extension, then a root one out of the same cached repo) and fails on the old code with exactly the truncated tree.

internal/config/spec_summary.go
planger: Agreed, fixed. Both are projected, with their own rows and diff entries. I put them next to `entrypoint override` rather than folding them into `skips approval`: they are argv appended to the agent, and what they grant depends on the flags they carry — approval bypass is one possibility, not the definition.

```
  entrypoint override  smoke
  continue args        --continue --dangerously-skip-permissions
  resume args          --resume
```

Covered by the `Spec.ContinueArgs` / `Spec.ResumeArgs` cases in `capabilityFieldCases` and by `TestSummarizeSpecDir`.

internal/config/spec_summary.go
planger: Agreed, and the hole was wider than the one you pointed at. `SpecSummary.InstallCommands` is `[]string` carrying the command text now, and `describeSpecCommand` became `specCommandText`, which no longer prefers the author's `description` — that closes the same hole on `commands.startup`. A step running as root is prefixed `(root) `, so a step *changing* user is visible even when the root count does not move.

The render is unchanged — still `commands.install (2, 1 as root)`, no command text — so the summary stays a summary and only the update diff goes down to the shell.

Your comment turned out to name a class rather than a single field, so I swept the rest of the projection for anything that can change what an extension does without changing the shape of the document. Two more:

- **`environment.variables` was compared by name only.** Flipping `NODE_TLS_REJECT_UNAUTHORIZED` from `1` to `0` printed `no capability changes`. It carries `NAME=value` now.
- **A `commands.initFiles` entry was compared by path only**, although `kit-init.sh` writes its `content` at its `mode` on every container start. The mode is carried through, and the content as a truncated SHA-256 — enough to report that the seeded file changed without pulling an arbitrarily long body into a summary that deliberately does not print command text.

An update that would previously have printed `no capability changes` now reads:

```
  adds install command (root) curl -fsSL https://elsewhere.example/i.sh | sh
  drops install command (root) curl -fsSL https://known.example/i.sh | sh
  adds environment var NODE_TLS_REJECT_UNAUTHORIZED=0
  drops environment var NODE_TLS_REJECT_UNAUTHORIZED=1
  adds init file ${WORKDIR}/.smoke.yaml mode=0600 writesProject content=59e3ba15c06e
  drops init file ${WORKDIR}/.smoke.yaml mode=0600 writesProject content=24d474e6fab0
```

`TestDiffCapabilitiesReportsRewrittenInstallCommand`, `TestDiffCapabilitiesReportsRewrittenStartupCommandUnderSameDescription`, and `TestDiffCapabilitiesReportsRewrittenEnvValueAndInitFile`.

What stays unprojected: `sandbox.image` (documented as a hint), `aiFilename`/`agentContext` (reserved), the container-side layout fields (`configDir`/`skillsDir`/`memoryDir`/`settingsFile`/`settingsTarget`), the QEMU sizing hints, and `priority`/`failOnInstallError` — none of which grant anything. Happy to be argued out of that reading on the layout fields.

internal/config/spec_summary.go
planger: Agreed, fixed. `Ports` is `[]SpecPortSummary{Container, Publish}`, and a published port renders and diffs differently from a merely declared one:

```
  ports  8080 (published on the host), 9090
```

One consequence worth knowing: a published port and a provider `oauthPort` with the same number no longer collapse into one entry, since only one of them carries the marker. That seemed the right trade — they are not the same grant.

internal/config/spec_summary.go
planger: Agreed, fixed. `postStart.openIDE` gets a row and a diff entry, phrased around what it does rather than the field name:

```
  post-start  theia (launched on the host once the container is running;
              the session runs detached)
```

internal/extinstall/stage.go
planger: Good catch, fixed with `os.Chtimes` right after the rename. You are right that the window was fiction for anything installed more than a day ago, and `installing.md` promises 24 hours.

`TestStagingCommitStartsReplacedRecoveryWindowAtSwap` stats the recovery copy from inside `stamp` — the only moment it exists on a successful commit — and fails on the old code.

internal/extinstall/fetch.go
planger: Agreed, merged: `r.blobless = !filterIgnoredByServer(filterStderr)` and one return.

internal/extinstall/narrate.go
planger: Agreed, dropped. Both callers pass a non-empty caption.

internal/extinstall/capabilities_diff.go
planger: Agreed, `slices.Contains` on each side. The lists are single-digit, so the maps were buying nothing but six lines.

internal/extinstall/capabilities.go
planger: Agreed. It is `entrypointDir(kind) string` now; the `map[string][]string` and both loops over it are gone, and the scripts land straight in `tree.StartupScripts` and get sorted once with the rest.

internal/extinstall/capabilities.go
planger: Agreed, rewritten as two `CutPrefix` calls and a `strings.Cut`, as suggested.

internal/extinstall/update.go
planger: Agreed, `slices.IndexFunc` with an index check.

internal/extinstall/result.go
planger: Agreed, one line with `slices.ContainsFunc`.

internal/config/spec_summary.go
planger: Agreed, both the environment variables and the credential source ids use it now.

internal/extinstall/filediff.go
planger: Agreed, `shown := changes[:min(len(changes), maxChangedFilesShown)]`.

internal/model/extension_kind.go
planger: Agreed, `Verb` is gone and `DirName`'s doc names both roles.

I had kept the alias as a seam in case the CLI verb and the directory name ever diverge, but that is a speculative future paying for a real duplicate today, and as you say the three call sites read fine either way.

sgraband (approved):
Thanks for the revisions, this is good to merge from my side. Very nice feature!

planger:
@sgraband Very coo, thank you for the review.
@xai Fine from your point of view too?

internal/extinstall/fetch.go
```diff
@@ -0,0 +1,522 @@
...
+	for _, line := range strings.Split(out, "\n") {
+		fields := strings.Fields(line)
+		if len(fields) != 2 {
+			continue
+		}
+		switch fields[1] {
+		case "refs/heads/" + ref:
+			branch = RemoteRef{Commit: fields[0], Ref: ref, RefType: RefTypeBranch}
+		case "refs/tags/" + ref:
+			tag = RemoteRef{Commit: fields[0], Ref: ref, RefType: RefTypeTag}
```
xai: An annotated tag is recorded as the tag object id, not a commit.

`parseLsRemote` matches only the `refs/tags/<ref>` line and ignores the peeled `refs/tags/<ref>^{}` line, so `RemoteRef.Commit` is the tag object's SHA for an annotated tag. `verifyCommit` (lines 376-392) then compares `rev-parse FETCH_HEAD`, which is the same tag object, so nothing notices.

Reproduced on this head: `enclave features add /tmp/kit --ref v1 --yes --json` against a repo with `git tag -a v1` returned `"commit": "35f295b3…"` (the tag object) while `git rev-parse v1^{commit}` is `8cb26617…`. The value reaches the sidecar, `list --json`'s `origin.commit`, the result envelope, and the section caption. `docs/extensions/installing.md` documents `commit` as "The exact commit installed", and the field is part of the `schemaVersion: "1"` contract, so it is hard to correct after release.

Suggested fix: prefer the `^{}` line for tags here (fall back to the unpeeled SHA for lightweight tags) and use `rev-parse <target>^{commit}` in `verifyCommit`; add a real-git test with `git tag -a`.

internal/extinstall/discover.go
```diff
@@ -0,0 +1,130 @@
...
+
+// classify parses each candidate's spec document from a materialized checkout.
+// A candidate whose spec cannot be read, or whose declared name disagrees with
+// its directory name, is marked with a Skip reason: the extension loader
+// resolves extensions by directory name, so such a directory could never load.
+func classify(repoDir string, candidates []candidate) []classified {
+	entries := make([]classified, 0, len(candidates))
+	for _, cand := range candidates {
+		entry := classified{candidate: cand}
+		summary, err := config.SummarizeSpecDir(filepath.Join(repoDir, filepath.FromSlash(cand.Dir)))
```
xai, line 90: Candidate specs are read through symlinks, unbounded, before the sanitizer runs.

`SummarizeSpecDir` ends in `os.ReadFile` (`parseSpecDocument` in `internal/config/spec_loader.go`), which follows symlinks and has no size cap. The symlink and size rules in `sanitize.go` only apply later in `applyPlan`, and unscoped discovery classifies every candidate, so `--name` does not narrow this.

Reproduced on this head: a source repo with `features/blocker/spec.yaml` committed as a symlink to a FIFO makes `enclave features add /tmp/kit --name other --yes` block until killed. A symlink to `/dev/zero` exhausts memory instead, and a symlink to a host file feeds that file to the YAML parser. `docs/extensions/installing.md` promises symlinks are rejected; the first read of repository content does not enforce that.

Suggested fix: `Lstat` the spec, require a regular file, and read through an `io.LimitReader` before parsing (or use `ls-tree -l` modes and sizes to drop non-blob or oversized specs before materializing).

xai (changes requested):
Great feature, I really love it!

My clanker found two more things that I could reproduce and that might be worth fixing before merging, see inline comments.

planger:
Both findings are fixed on the rebased branch. A few things I did beyond the comments themselves:

Peeling with `rev-parse <target>^{commit}` alone would have broken `--ref <annotated-tag-object-sha>`. `fullSHAPattern` accepts any object id, so such a pin takes the commit path and would then fail with "requested commit X but fetched Y". The pin is now checked against the object the fetch landed on, and only the recorded value gets peeled. `revParse` also passes `--verify`, so a tag that peels to a tree fails with one line instead of git's ambiguous-argument advice.

The hardened spec read is scoped to the installer: `SummarizeSpecDir` uses it, `LoadSpec` keeps plain `os.ReadFile`. A symlinked `spec.yaml` under `~/.config/enclave` is your own file and still works, while one committed in a fetched repository is reported as unusable. `spec.json` goes through the same check, and the 1 MiB cap is tested on both sides of the boundary.

`parseLsRemote` also got a table test next to the real-git one, mostly to pin that a branch still beats a same-named tag without picking up that tag's peeled line.

Worth knowing before this merges: an extension already installed from an annotated tag re-installs once on the first `update`, since the recorded tag id no longer matches the resolved commit. The content is identical, so no "rebuilds the image" hint appears, because that hint comes from the tree hash rather than the commit.

internal/extinstall/install.go
```diff
@@ -0,0 +1,450 @@
...
+			}
+			return &plan, nil
+		}
+		return nil, err
+	}
+
+	origin, err := readOrigin(target)
+	if err != nil {
+		return nil, err
+	}
```
xai, line 261: Corrupt sidecar blocks add, `--force` included:
planFor returns the readOrigin error unconditionally. Reproduced on dc8a64d: after writing `not json` into .enclave-source.json, `features update demo` says "its provenance file could not be read; re-add it to manage it", and then `features add <kit> --name demo --yes --force --json` fails with the JSON parse error. The only recovery is remove --force then add, which neither the hint nor docs/extensions/installing.md say. Mirror removeOne: without --force fail with "its provenance file could not be read; pass --force to replace it"; with --force treat the directory as unmanaged and proceed.

xai (approved):
Looks good, all blockers fixed. The new inline comment in install.go would be a follow-up.

planger:
Thanks for your reviews!

I've seeded the extensions repo: https://github.com/eclipse-enclave/enclave-extensions
And I created a follow-up issue: https://github.com/eclipse-enclave/enclave/issues/69

PR #61, opened 2026-08-29
Coordinate concurrent image builds and session startup

xai:
#### What it does

- Lock runtime image builds by image name. After waiting, recalculate the build plan and skip work another process completed unless `--rebuild` was requested.
- Lock session startup by tool and project until the backend reports the container running. This prevents concurrent starts from choosing the same container name without blocking later sessions for the lifetime of a foreground run.

Thanks @kalyi for initially reporting this!

#### How to test

Run `make build`.

Then in  parallel start >=2 sessions that trigger a build: `bin/enclave`.
If you already have an image built for this branch,  you can delete it using `docker
  rmi enclave-<tool>:fix-concurrent-image-build-lock-latest`.
One of the sessions should execute the image build; the others should print something like `info: Waiting for another enclave process to finish building enclave-<tool>:fix-concurrent-image-build-lock-latest.`

#### Follow-ups

None.

#### Breaking changes

#### Review checklist


internal/app/build.go
```diff
@@ -68,6 +68,29 @@ func (p runtimeImageBuildPlan) NeedsRebuild() bool {
...
+	lockPath := config.HostLockPath(home, lockName)
+	release, waited, err := util.AcquireFileLock(lockPath, func() {
+		logx.Infof("Waiting for another enclave process to finish building %s.", imageName)
+	})
+	if err != nil {
+		return err
+	}
+	defer release()
+	buildPlan := initialBuildPlan
+	if waited {
```
EclipseSourceAI: The re-resolve only runs when `waited` is true, so if another process finishes its build between `resolveRuntimeImageBuildPlan` in `ensureRuntimeImage` and `tryLockFile` here, this caller still builds from the stale plan. Dropping the `waited` guard and always resolving inside the lock closes that window for the cost of one extra plan resolution.

internal/backend/types.go
```diff
@@ -236,9 +236,11 @@ type AttachIO struct {
...
 	DetachKeys string
-	// OnStarted, if set, is invoked once the backend confirms the session is
-	// running (e.g. to announce published ports). Backends that cannot observe
-	// startup may skip it.
+	// OnStarted, if set, must be invoked exactly once as soon as the backend
+	// confirms the session is running. The runtime releases the session-start
+	// lock here (and announces published ports); a foreground Run that never
+	// calls it holds that lock until the session exits, blocking every other
+	// session start for the same tool and project.
 	OnStarted func()
```
EclipseSourceAI, line 247: The docker backend only forwards `OnStarted` on the interactive TTY path and drops it in the other three branches of `runForeground` ([link](https://github.com/eclipse-enclave/enclave/blob/64d00f4a61b34727260d48ed12c3c22190f258a0/internal/backend/docker/docker.go#L556-L575)). Nothing routes through those today, but with the lock release wired to this hook they would silently hold the session-start lock for the whole session, so they should forward it too.

internal/runtime/background.go
```diff
@@ -18,6 +18,12 @@ import (
 // ExecuteBackground starts a detached container that runs the tool command as
 // PID 1; it can be reattached later via `docker attach`.
 func (r *Runtime) ExecuteBackground() (string, error) {
+	releaseStartLock, err := r.acquireSessionStartLock()
+	if err != nil {
+		return "", err
+	}
+	defer releaseStartLock()
```
EclipseSourceAI, line 25: The lock stays held through `announcePublishedPorts` and `runPostStart` (container inspects plus an IDE launch), well past the point the backend reported the container running. Releasing right after `be.Start` succeeds would match what the PR description promises and what the foreground path does.

internal/runtime/runtime.go
```diff
@@ -186,7 +192,19 @@ func (r *Runtime) Execute() error {
 	}
 	// The backend syncs auth files from the config store to the shared auth
 	// store after the container exits, per the request's AuthSync intent.
-	return r.runContainer(ctx)
+	return r.runContainer(ctx, releaseStartLock)
+}
+
+func (r *Runtime) acquireSessionStartLock() (func(), error) {
+	lockName := "session-start-" + util.HashString(r.baseContainerName()) + ".lock"
```
EclipseSourceAI: Gateway image builds still race the same way: `gateway.Start` does an unsynchronized check-then-build ([link](https://github.com/eclipse-enclave/enclave/blob/64d00f4a61b34727260d48ed12c3c22190f258a0/internal/gateway/gateway.go#L348-L355)) and its image is keyed by profile only, so two projects on the same tool build `enclave-gateway-<profile>` concurrently even with this lock held. Worth a follow-up issue given the PR lists none.

internal/app/command_run_test.go
```diff
@@ -110,3 +113,91 @@ func TestEnsureExistingRuntimeImageWith(t *testing.T) {
...
+	var imageReady atomic.Bool
+	var builds atomic.Int32
+	var resolves atomic.Int32
+	firstBuildStarted := make(chan struct{})
+	releaseFirstBuild := make(chan struct{})
+	secondResolveStarted := make(chan struct{})
+
+	resolve := func() (runtimeImageBuildPlan, error) {
+		if resolves.Add(1) == 1 {
+			close(secondResolveStarted)
```
EclipseSourceAI, line 128: `secondResolveStarted` is closed on the first resolve call, not the second. It happens to be correct because only the contended caller resolves, but the name reads backwards.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Two host-level file locks to stop concurrent `enclave` processes from stepping on each other: one keyed by runtime image name around the build, one keyed by tool plus project hash around session startup. `util.WithFileLock` is refactored into `AcquireFileLock`, which adds a non-blocking probe so callers learn whether they actually waited and get an idempotent release function.

The approach fits the existing lock conventions (`config.HostLockPath` plus `util.HashString`, same naming shape as the hoststore and skills-mount locks), and builds/tests pass locally including `-race` and a `GOOS=windows` build.

Critical points for a maintainer:

- The image-build double check is guarded by `waited`, which leaves the exact race the PR is closing open when the other process releases the lock just before this one probes it.
- The session-start lock release is wired to `AttachIO.OnStarted`, and the doc comment now declares that hook mandatory. The docker backend still drops it in three of the four `runForeground` branches, so the new contract is only honored by luck of the current call path.
- The lock is deliberately released mid-`Run` for foreground sessions. Correctness there rests on `RunInteractiveWithStartHook` firing the hook after the container name is registered, and on `Execute`'s deferred release covering the paths where it never fires. Both hold today, but it is the part most worth a second pair of eyes.
- Gateway image builds are still unsynchronized and keyed by profile only, so the reported problem is not fully gone across projects.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Follow-up review after `c4a8214` ("fix: address concurrent startup review findings").

All five points from my previous review are addressed in that commit:
- `coordinateRuntimeImageBuild` now always re-resolves the build plan inside the lock, no more `waited` guard.
- `runForeground` forwards `attach.OnStarted` in all four branches via the new `RunWithStartHook` / `RunWithIOAndStartHook` helpers, so the session-start lock is released consistently regardless of TTY/IO wiring.
- `ExecuteBackground` now releases the start lock right after `be.Start` succeeds, before port announcement and post-start IDE launch.
- Gateway image builds are now serialized with a lock keyed by the image name (mirrors the runtime-image build lock), closing the check-then-build race.
- The `secondResolveStarted` test channel now closes on the second resolve call, matching its name.

Checked build and tests (`go build ./...`, `go test` on `internal/app`, `internal/gateway`, `internal/docker`, `internal/runtime`, `internal/backend/...`) locally, all green. No new issues found in the changes since my last review.

- [re-resolve build plan unconditionally inside the lock, not just when waited](https://github.com/eclipse-enclave/enclave/pull/61#discussion_r3886734321)
- [docker backend forward OnStarted in all runForeground branches](https://github.com/eclipse-enclave/enclave/pull/61#discussion_r3886734342)
- [release session-start lock right after be.Start succeeds in background path](https://github.com/eclipse-enclave/enclave/pull/61#discussion_r3886734375)
- [serialize gateway image builds with a lock keyed by image name](https://github.com/eclipse-enclave/enclave/pull/61#discussion_r3886734412)
- [fix secondResolveStarted to close on the second resolve call](https://github.com/eclipse-enclave/enclave/pull/61#discussion_r3886734457)

xai:
Rebased against main

internal/gateway/gateway.go
```diff
@@ -201,6 +201,27 @@ func buildGatewayImage(paths model.Paths, profile model.Profile, allowlistPath s
 	return nil
 }
 
+func coordinateGatewayImageBuild(home string, image string, forceRebuild bool, resolveBuildPlan func() (bool, string, error), executeBuild func(string) error) error {
```
tortmayr, line 204: This is `coordinateRuntimeImageBuild` from `internal/app/build.go` again, identical apart from the plan type: same `"image-build-" + util.HashString(image)` name, same wait message, same recheck shape. Two copies of the same lock protocol in two packages will drift. The lock-name derivation belongs next to [`HostLockPath`](https://github.com/eclipse-enclave/enclave/blob/7591f109f0cbdf34d26f090a27b65c85636a4d74/internal/config/host_paths.go#L24-L30) with a generic `AcquireImageBuildLock(home, image)`, leaving each package only its own resolve/execute pair.

internal/app/command_run.go
```diff
@@ -274,33 +274,39 @@ func ensureRuntimeImage(input *CommandInput, opts model.Options, buildCfg *build
...
 
-func buildOrReuseRuntimeImage(input *CommandInput, opts model.Options, host model.Host, resolved buildConfig, buildPlan runtimeImageBuildPlan) int {
-	reused := false
-	if !opts.ForceRebuild && buildPlan.StructuralRebuild && !buildPlan.AgentUpdates.NeedsRebuild {
-		ok, reuseErr := reuseRuntimeImageByContentHash(context.Background(), resolved.ImageName, buildPlan.CombinedHash)
-		if reuseErr != nil {
-			logx.Debugf("content-cache lookup failed: %v", reuseErr)
+func buildOrReuseRuntimeImage(input *CommandInput, opts model.Options, host model.Host, resolved buildConfig) int {
+	resolveBuildPlan := func() (runtimeImageBuildPlan, error) {
+		return resolveRuntimeImageBuildPlan(input.Ctx.Paths, resolved, opts.BuildOptions, opts.Tool, host.Home, opts.ForceRebuild, time.Now().UTC())
```
tortmayr: `resolveRuntimeImageBuildPlan` is not a pure computation: it walks into [`planAgentUpdatesForTools`](https://github.com/eclipse-enclave/enclave/blob/7591f109f0cbdf34d26f090a27b65c85636a4d74/internal/app/build.go#L826-L870) → [`probeToolUpdateFingerprint`](https://github.com/eclipse-enclave/enclave/blob/7591f109f0cbdf34d26f090a27b65c85636a4d74/internal/app/build.go#L1026-L1067), which `docker run`s the tool image and hits the network. The stamp file is only written after the build, so the in-lock resolve sees `Due` again and probes a second time: two container runs and two network fetches per auto-rebuild. Worse, if the second probe fails transiently, `Known` goes false and an agent-update-only rebuild is silently skipped even though the outer check decided one was due. Memoizing the probe result per process would fix both.

internal/runtime/runtime.go
```diff
@@ -192,7 +198,19 @@ func (r *Runtime) Execute() error {
...
```
tortmayr: The races this closes (`nextSessionName()` and the `runningConfigKeyExists` branch in `deriveConfigVolumeSuffix`) only happen when no explicit name was given. With `--name`, both the container name and the store suffix are deterministic, so two differently-named sessions in the same project now serialize their entire startup, gateway image build included, for no benefit. Skipping the lock when `r.run.SessionName != ""` keeps `--background` (which does allocate a name) covered.

internal/runtime/runtime.go
```diff
@@ -192,7 +198,19 @@ func (r *Runtime) Execute() error {
...
+	lockPath := config.HostLockPath(r.host.Home, lockName)
+	release, _, err := util.AcquireFileLock(lockPath, func() {
+		logx.Infof("Waiting for another enclave session to finish starting.")
```
tortmayr: The build-lock message names the image; this one gives the user nothing to act on. Include the tool and project, e.g. `Waiting for another enclave session to finish starting (%s).` with `r.baseContainerName()`.

internal/docker/run.go
```diff
@@ -44,22 +50,24 @@ func Run(ctx context.Context, config *ContainerConfig, hostConfig *HostConfig, n
...
 	cmd.Stderr = &stderr
-	return classifyRunError(args, cmd.Run(), stderr.String())
+	return classifyRunError(args, runCommandWithStartHook(ctx, name, cmd, onStarted), stderr.String())
 }
 
 // RunWithIO runs a container wired to the supplied streams (no TTY) and returns
 // an *ExitError when the container exits non-zero.
 func RunWithIO(ctx context.Context, config *ContainerConfig, hostConfig *HostConfig, name string, in io.Reader, out io.Writer, errOut io.Writer) error {
-	return runWithIO(ctx, config, hostConfig, name, in, out, errOut, false)
+	return RunWithIOAndStartHook(ctx, config, hostConfig, name, in, out, errOut, false, nil)
```
tortmayr: `RunWithIO` and `RunWithIOAndTTY` have no callers left. Their only call sites were the two `runForeground` branches this PR replaced ([main](https://github.com/eclipse-enclave/enclave/blob/7591f109f0cbdf34d26f090a27b65c85636a4d74/internal/backend/docker/docker.go#L569-L572)). Drop both.

internal/util/lock.go
```diff
@@ -10,25 +10,55 @@ package util
...
 	"path/filepath"
+	"sync"
 )
 
-// WithFileLock runs fn while holding an exclusive lock on path.
-func WithFileLock(path string, fn func() error) error {
+// AcquireFileLock acquires an exclusive lock on path and returns an idempotent
+// release function. onWait is called when another process currently holds the
+// lock, immediately before blocking for it.
+func AcquireFileLock(path string, onWait func()) (release func(), waited bool, err error) {
```
tortmayr: `waited` has no consumer any more: it existed for the `waited`-guarded recheck that `cfc39c6` removed, and all three production call sites now pass `_`. Either drop it from the signature or keep it and let `coordinate*ImageBuild` use it to skip the redundant resolve when uncontended.

internal/gateway/gateway_test.go
```diff
@@ -54,6 +57,65 @@ func TestEnsureExistingGatewayImageWith(t *testing.T) {
...
+		errs <- coordinateGatewayImageBuild(home, "enclave-gateway-codex:latest", false, resolve, execute)
+	}()
+	<-firstBuildStarted
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		errs <- coordinateGatewayImageBuild(home, "enclave-gateway-codex:latest", false, resolve, execute)
+	}()
+
+	select {
```
tortmayr, line 102: This one missed the hardening the app-side twin got in `cfc39c6`: there's no handshake confirming the second goroutine reached `AcquireFileLock`, and the `t.Fatal` fires before `close(releaseFirstBuild)`, so on failure the first goroutine blocks forever in `execute` holding the flock. Mirror `TestCoordinateRuntimeImageBuildRechecksAfterWaiting`: record a `serialized` bool, always close the channel, assert after `wg.Wait()`.

internal/runtime/runtime.go
tortmayr, line 1: Nothing in `docs/` mentions that sessions can now block on each other. `ARCHITECTURE.md` already documents the [embedded-asset lock](https://github.com/eclipse-enclave/enclave/blob/7591f109f0cbdf34d26f090a27b65c85636a4d74/docs/ARCHITECTURE.md#L202) under "Other host-side data", so the two new locks under `~/.local/state/enclave/locks/` and the resulting wait behavior fit right there (or in `docs/persistence.md`, which is where parallel `--name` sessions are described).

internal/gateway/gateway.go
xai: The lock protocol now lives in one place, so it cannot drift between the two packages.

internal/app/command_run.go
xai: Good catch, especially the additional network request is annoying. We now memoize the probe per process, as suggested.

internal/runtime/runtime.go
xai: Kept the lock in scope for `--name` sessions on purpose. Named starts still share state beyond name allocation:

- `startGateway` writes the per-project/per-tool gateway config bundle that every session of that tool mounts, so two named starts would write it concurrently.
- An explicit numeric name (`--name 1`) resolves to the same container an unnamed --background start allocates next, so the named side is only deterministic on its own.
- The already-running check in `prepareExecution` is check-then-act. Two `--name` task-a starts that both pass it reach `gateway.Start`, which force-removes the existing sidecar; the loser's cleanup then stops the winner's gateway.

The lock is released once the container is running, so the recurring cost is a few seconds of startup, not the full session. Documented in `docs/persistence.md`, with tests pinning the lock scope in `session_name_test.go`.

internal/docker/run.go
xai: Oh, perfect! Both gone now.

internal/util/lock.go
xai: Correct, removed!

internal/gateway/gateway_test.go
xai: Right. The two twins are now line-for-line the same apart from the function under test and the image name.

internal/runtime/runtime.go
xai: Fixed in both.

xai:
rebased against main

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Follow-up review after `d82ea8b` ("fix: memoize the agent update probe across the image-build recheck"), which responds to tortmayr's review round.

All six of tortmayr's points are addressed in that commit:
- Gateway and runtime image-build locks now share `config.AcquireImageBuildLock`/`HostImageBuildLockPath`, so the lock protocol can't drift between packages.
- The agent-update fingerprint probe is memoized per tool (`memoizeToolFingerprintProbe`), so the pre-lock and in-lock plan resolves only run the check-update container and its network fetch once, with tests covering both the memoization and the reused-probe-across-resolves case.
- The `--name` session-start lock scope is kept intentionally (justified by the shared gateway config bundle, numeric-name collisions, and the check-then-act already-running check), now backed by `sessionStartLockName` tests pinning the scope for named/background/unnamed starts.
- Dead `RunWithIO`/`RunWithIOAndTTY` helpers are removed (confirmed no remaining callers).
- The unused `waited` return from `AcquireFileLock` is dropped from the signature and all call sites.
- `gateway_test.go`'s `TestCoordinateGatewayImageBuildRechecksAfterWaiting` now mirrors the app-side twin exactly (handshake channel, `close` before `t.Fatal`, assert after `wg.Wait()`).
- Lock behavior is now documented in `docs/ARCHITECTURE.md` and `docs/persistence.md`.

Verified `go build ./...`, `go vet ./...`, and `go test` (including `-race`) across `internal/app`, `internal/gateway`, `internal/runtime`, `internal/util`, `internal/config`, `internal/docker`, `internal/backend/...`, all green. No new issues found in the changes since the last review round.

tortmayr (approved):
Thanks for addressing the issues.
LGTM! 👍🏼

PR #63, opened 2026-09-01
fix: confine the arch import walk to this module

planger:
## Problem

`TestDockerImportsConfinedToBackendAndCarveOuts` walks the whole repository root and skips only `.git`, `node_modules`, `frontend`, and `bin`. It therefore descends into nested checkouts that live inside the worktree, such as agent worktrees under `.claude/`, and reports their files as boundary violations:

```
--- FAIL: TestDockerImportsConfinedToBackendAndCarveOuts
    arch_test.go:83: files import enclave/internal/docker outside the Docker backend and approved integration paths:
          .claude/worktrees/<name>/internal/app/build.go
          .claude/worktrees/<name>/internal/backend/docker/docker.go
          ...
```

Those files are not part of this module. `go list ./...` and `go test ./...` both stop at nested module boundaries, so the toolchain never builds them; only this hand-rolled walk saw them. The paths also cannot match the allowlist, since the allowlist entries are module-relative prefixes like `internal/docker/`.

The result is that `make test` fails for anyone with a nested checkout under the repository, which is easy to hit because `.claude/` is gitignored and agent tooling puts worktrees there.

## Fix

Skip directories that carry their own `go.mod`, so the walk covers the same files the toolchain builds. Only the root `go.mod` is tracked in this repository, so no first-party code is excluded.

## Verification

- The failing test passes with a nested worktree present in the working tree.
- Reintroducing a genuine violation (a file in `internal/runtime` importing `internal/docker`) still fails the test, so the guard is not weakened.
- `make build`, `make test`, and `make check-license-headers` pass. `make lint` could not run locally because `shellcheck` is not installed; `go vet` and `golangci-lint` are clean on the changed package.

### DIFF
--- a/internal/arch/arch_test.go
+++ b/internal/arch/arch_test.go
@@ -14,6 +14,7 @@ import (
 	"go/parser"
 	"go/token"
 	"io/fs"
+	"os"
 	"path/filepath"
 	"strings"
 	"testing"
@@ -50,6 +51,9 @@ func TestDockerImportsConfinedToBackendAndCarveOuts(t *testing.T) {
 		if d.IsDir() {
 			name := d.Name()
 			if name == ".git" || name == "node_modules" || name == "frontend" || name == "bin" {
+				return filepath.SkipDir
+			}
+			if path != root && isModuleRoot(path) {
 				return filepath.SkipDir
 			}
 			return nil
@@ -127,6 +131,14 @@ func isAllowedDockerImporter(rel string) bool {
 		}
 	}
 	return false
+}
+
+// isModuleRoot reports whether dir carries its own go.mod. Nested modules are
+// separate checkouts (agent worktrees, scratch clones) that the Go toolchain
+// excludes from ./..., so the boundary walk must exclude them too.
+func isModuleRoot(dir string) bool {
+	_, err := os.Stat(filepath.Join(dir, "go.mod"))
+	return err == nil
 }
 
 func repoRoot(t *testing.T) string {


internal/arch/arch_test.go
```diff
@@ -52,6 +53,9 @@ func TestDockerImportsConfinedToBackendAndCarveOuts(t *testing.T) {
 			if name == ".git" || name == "node_modules" || name == "frontend" || name == "bin" {
 				return filepath.SkipDir
 			}
+			if path != root && isModuleRoot(path) {
+				return filepath.SkipDir
+			}
```
EclipseSourceAI, line 58: This only catches nested checkouts that carry a `go.mod`, but the failure mode is broader: I dropped `.claude/scratchtest/x.go` (importing `internal/docker`, no `go.mod` anywhere above it) into the worktree and the test still failed. Since the justification is matching what the toolchain builds, the go tool also ignores directories whose name starts with `.` or `_`, so extending the name check at [arch_test.go#L52-L53](https://github.com/eclipse-enclave/enclave/blob/1355fc55b27acf1a2586169416b6366688946166/internal/arch/arch_test.go#L52-L53) covers the whole `.claude/` tree and makes the explicit `.git` entry redundant. No tracked Go files live under a dot directory today, so nothing first-party gets dropped.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Test-only fix: `TestDockerImportsConfinedToBackendAndCarveOuts` walks the repo root by hand, and the walk descended into nested checkouts under `.claude/`, reporting their files as boundary violations. The PR skips any directory carrying its own `go.mod`, matching the module boundary the Go toolchain stops at.

The approach is sound and the scope is tight. I verified the claims: only the root `go.mod` is tracked, so no first-party code is excluded, and the test passes on the merged state. `isModuleRoot` does not duplicate anything else in the tree, and `TestDockerBackendDoesNotImportRuntime` is unaffected since it globs a fixed directory.

One point for the maintainer: the fix is narrower than the rationale it gives. Nested trees without a `go.mod` (a plain copy of sources under `.claude/`) still trip the walk, which I reproduced. Widening the existing directory-name filter to dot and underscore prefixes would cover the class the go tool itself ignores. Details inline.

xai (approved):
Approved! Comment by the bot is correct, but can be done as follow-up. Having the fix as-is is more important.

PR #64, opened 2026-09-02
fix: keep pnpm store in persistent cache

tortmayr:
#### What it does

 - Configure pnpm to use Enclave's persistent cache mount instead of creating
    `.pnpm-store` in the project workspace
  - Preserve `--no-cache` behavior and normalize the store path for devcontainer
    remote users
#### Breaking changes

#### Review checklist

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -418,6 +418,7 @@ normalize_devcontainer_paths() {
         COREPACK_ROOT
         VOLTA_HOME
         FNM_DIR
+        PNPM_CONFIG_STORE_DIR
         PNPM_STORE_DIR
         YARN_CACHE_FOLDER
         XDG_CACHE_HOME
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -419,13 +419,13 @@ normalize_devcontainer_paths() {
         VOLTA_HOME
         FNM_DIR
         PNPM_CONFIG_STORE_DIR
-        PNPM_STORE_DIR
         YARN_CACHE_FOLDER
         XDG_CACHE_HOME
         XDG_CONFIG_HOME
         XDG_DATA_HOME
         npm_config_prefix
         npm_config_cache
+        npm_config_store_dir
         npm_config_userconfig
     )
     local -a extra_rewrite_vars=()
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -425,7 +425,6 @@ normalize_devcontainer_paths() {
         XDG_DATA_HOME
         npm_config_prefix
         npm_config_cache
-        npm_config_store_dir
         npm_config_userconfig
     )
     local -a extra_rewrite_vars=()
@@ -510,6 +509,54 @@ normalize_devcontainer_paths() {
     if [ "$changed" = "1" ]; then
         echo "Normalized devcontainer paths from $remote_home to $current_home"
     fi
+}
+
+# Rewrites a config file in place with store-dir set to $2, preserving any other
+# entries. In place because ~/.npmrc is a bind-mounted file, where replacing the
+# inode via mktemp+mv fails.
+write_store_dir_config() {
+    local file="$1"
+    local store_dir="$2"
+    local existing=""
+
+    if [ -e "$file" ]; then
+        [ -w "$file" ] || return
+        # Drop a previously written entry so restarts cannot accumulate
+        # duplicates or keep a stale path from an earlier container home.
+        existing="$(grep -v '^[[:space:]]*store-dir[[:space:]]*=' "$file" || true)"
+    fi
+    {
+        if [ -n "$existing" ]; then
+            printf '%s\n' "$existing"
+        fi
+        if [ -n "$store_dir" ]; then
+            printf 'store-dir=%s\n' "$store_dir"
+        fi
+    } > "$file" 2>/dev/null || true
+}
+
+# pnpm 11 and later read PNPM_CONFIG_STORE_DIR from the environment, while pnpm 9
+# and 10 ignore it and take store-dir from pnpm's global config file (which 11 and
+# later in turn ignore). The file covers the older versions without depending on
+# PATH, so pnpm installed after startup via corepack, a global install, or an nvm
+# version switch is still pinned to the cache-backed store. It is pnpm's own file
+# rather than ~/.npmrc, where npm 11+ warns about store-dir as an unknown user
+# config on every npm invocation.
+configure_pnpm_store() {
+    local store_dir="${PNPM_CONFIG_STORE_DIR:-}"
+    if [ -z "$store_dir" ] || [ -z "${HOME:-}" ]; then
+        return
+    fi
+
+    # Earlier releases wrote the key to ~/.npmrc; strip it so the npm warning
+    # does not survive in the persisted file.
+    if [ -e "$HOME/.npmrc" ] && grep -q '^[[:space:]]*store-dir[[:space:]]*=' "$HOME/.npmrc" 2>/dev/null; then
+        write_store_dir_config "$HOME/.npmrc" ""
+    fi
+
+    local rc_dir="${XDG_CONFIG_HOME:-$HOME/.config}/pnpm"
+    mkdir -p "$rc_dir" 2>/dev/null || return
+    write_store_dir_config "$rc_dir/rc" "$store_dir"
 }
 
 cleanup_devcontainer_once_stamps() {
@@ -599,6 +646,11 @@ run_devcontainer_command() {
 
 if [ "${ENCLAVE_DEVCONTAINER:-}" = "1" ]; then
     normalize_devcontainer_paths
+fi
+
+configure_pnpm_store
+
+if [ "${ENCLAVE_DEVCONTAINER:-}" = "1" ]; then
     run_devcontainer_command "post-create" "${ENCLAVE_DEVCONTAINER_POST_CREATE:-}" "1"
     run_devcontainer_command "post-start" "${ENCLAVE_DEVCONTAINER_POST_START:-}" ""
 fi
--- a/internal/runtime/entrypoint_colorterm_test.go
+++ b/internal/runtime/entrypoint_colorterm_test.go
@@ -9,7 +9,6 @@ package runtime
 
 import (
 	"os"
-	"os/exec"
 	"path/filepath"
 	"testing"
 )
@@ -34,63 +33,10 @@ func TestEntrypoint_COLORTERMPreservesExplicitValue(t *testing.T) {
 	}
 }
 
-func TestEntrypointNormalizesPnpmStoreForDevcontainerUser(t *testing.T) {
-	home := t.TempDir()
-	projectDir := filepath.Join(home, "project")
-	if err := os.MkdirAll(projectDir, 0o755); err != nil {
-		t.Fatalf("mkdir project dir: %v", err)
-	}
-
-	entrypointPath := filepath.Join("..", "..", "entrypoint.sh")
-	captureEnv := `printf "%s" "${PNPM_CONFIG_STORE_DIR:-}" > "$HOME/pnpm-store-dir.out"`
-	cmd := exec.Command("bash", entrypointPath, "bash", "-c", captureEnv)
-	cmd.Env = []string{
-		"PATH=" + os.Getenv("PATH"),
-		"HOME=" + home,
-		"PROJECT_DIR=" + projectDir,
-		"TOOL=pi",
-		"ENCLAVE_DEVCONTAINER=1",
-		"ENCLAVE_DEVCONTAINER_REMOTE_USER=node",
-		"PNPM_CONFIG_STORE_DIR=/home/node/.local/share/pnpm/store",
-	}
-
-	output, err := cmd.CombinedOutput()
-	if err != nil {
-		t.Fatalf("entrypoint failed: %v\noutput:\n%s", err, string(output))
-	}
-	value, err := os.ReadFile(filepath.Join(home, "pnpm-store-dir.out"))
-	if err != nil {
-		t.Fatalf("read pnpm store output: %v", err)
-	}
-	want := filepath.Join(home, ".local", "share", "pnpm", "store")
-	if string(value) != want {
-		t.Fatalf("PNPM_CONFIG_STORE_DIR = %q, want %q", string(value), want)
-	}
-}
-
 func runEntrypointCaptureCOLORTERM(t *testing.T, extraEnv []string) (string, string, error) {
 	t.Helper()
 
-	home := t.TempDir()
-	projectDir := filepath.Join(home, "project")
-	if err := os.MkdirAll(projectDir, 0o755); err != nil {
-		t.Fatalf("mkdir project dir: %v", err)
-	}
-
-	entrypointPath := filepath.Join("..", "..", "entrypoint.sh")
-	cmd := exec.Command("bash", entrypointPath, "bash", "-lc", `printf "%s" "${COLORTERM:-}" > "$HOME/colorterm.out"`)
-	env := []string{
-		"PATH=" + os.Getenv("PATH"),
-		"HOME=" + home,
-		"PROJECT_DIR=" + projectDir,
-		"TOOL=pi",
-	}
-	if len(extraEnv) > 0 {
-		env = append(env, extraEnv...)
-	}
-	cmd.Env = env
-
-	out, err := cmd.CombinedOutput()
+	home, out, err := runEntrypointCommand(t, extraEnv, "bash", "-lc", `printf "%s" "${COLORTERM:-}" > "$HOME/colorterm.out"`)
 	valueBytes, readErr := os.ReadFile(filepath.Join(home, "colorterm.out"))
 	if readErr != nil {
 		t.Fatalf("read colorterm output: %v\nentrypoint output:\n%s", readErr, string(out))
--- a/internal/runtime/entrypoint_colorterm_test.go
+++ b/internal/runtime/entrypoint_colorterm_test.go
@@ -34,6 +34,40 @@ func TestEntrypoint_COLORTERMPreservesExplicitValue(t *testing.T) {
 	}
 }
 
+func TestEntrypointNormalizesPnpmStoreForDevcontainerUser(t *testing.T) {
+	home := t.TempDir()
+	projectDir := filepath.Join(home, "project")
+	if err := os.MkdirAll(projectDir, 0o755); err != nil {
+		t.Fatalf("mkdir project dir: %v", err)
+	}
+
+	entrypointPath := filepath.Join("..", "..", "entrypoint.sh")
+	captureEnv := `printf "%s" "${PNPM_CONFIG_STORE_DIR:-}" > "$HOME/pnpm-store-dir.out"`
+	cmd := exec.Command("bash", entrypointPath, "bash", "-c", captureEnv)
+	cmd.Env = []string{
+		"PATH=" + os.Getenv("PATH"),
+		"HOME=" + home,
+		"PROJECT_DIR=" + projectDir,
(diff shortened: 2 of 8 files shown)


tortmayr:
@EclipseSourceAI

tortmayr:
@EclipseSourceAI

internal/runtime/runtime.go
```diff
@@ -1120,6 +1120,7 @@ func (r *Runtime) addCacheMounts(mounts *mountAccumulator) {
 	mounts.AddMount(bindMount(filepath.Join(cacheDir, "cargo"), r.containerHome+"/.cargo", false))
 	// pnpm store
 	mounts.AddMount(bindMount(filepath.Join(cacheDir, "pnpm"), r.containerHome+"/.local/share/pnpm", false))
+	mounts.AddEnv("PNPM_CONFIG_STORE_DIR", r.containerHome+"/.local/share/pnpm/store")
 	// uv (Python) cache
 	mounts.AddMount(bindMount(filepath.Join(cacheDir, "uv"), r.containerHome+"/.cache/uv", false))
 	// Yarn cache
```
EclipseSourceAI, line 1128: `PNPM_CONFIG_STORE_DIR` is only honored from pnpm 11 on. I checked against the real CLIs: 9.15.9 and 10.34.5 ignore it and read `npm_config_store_dir` instead, 11.25.0 does the opposite. The image's bundled pnpm is latest so this works there, but devcontainer images and repos pinning `packageManager: pnpm@10.x` through corepack silently keep the workspace-local store, so both vars are probably needed (npm logs `Unknown env config "store-dir"` for the `npm_config_` one).

internal/runtime/runtime.go
```diff
@@ -1120,6 +1120,7 @@ func (r *Runtime) addCacheMounts(mounts *mountAccumulator) {
 	mounts.AddMount(bindMount(filepath.Join(cacheDir, "cargo"), r.containerHome+"/.cargo", false))
 	// pnpm store
 	mounts.AddMount(bindMount(filepath.Join(cacheDir, "pnpm"), r.containerHome+"/.local/share/pnpm", false))
+	mounts.AddEnv("PNPM_CONFIG_STORE_DIR", r.containerHome+"/.local/share/pnpm/store")
 	// uv (Python) cache
 	mounts.AddMount(bindMount(filepath.Join(cacheDir, "uv"), r.containerHome+"/.cache/uv", false))
```
EclipseSourceAI, line 1127: Forcing the path also bypasses pnpm's same-device check, which is what was picking `.pnpm-store` in the first place, so where the cache mount and the workspace sit on different filesystems installs now fall back to copying instead of hard-linking. It also overrides a repo's own `storeDir` from `pnpm-workspace.yaml` (env config wins), both worth a line in the docs since AGENTS.md asks for user-visible behavior changes to be documented.

internal/runtime/env_variables_test.go
```diff
@@ -80,6 +80,37 @@ func TestBaseMountsIncludesMixinEnvVariables(t *testing.T) {
...
+	mounts := newMountAccumulator(nil, nil)
+	r.addCacheMounts(mounts)
+
+	want := "/home/agent/.local/share/pnpm/store"
+	if !envSliceContainsKV(mounts.Env(), "PNPM_CONFIG_STORE_DIR", want) {
+		t.Fatalf("expected PNPM_CONFIG_STORE_DIR=%s, got %v", want, mounts.Env())
+	}
+}
+
+func TestAddCacheMountsOmitsPnpmStoreWithoutCache(t *testing.T) {
```
EclipseSourceAI: With `--no-cache` the var is dropped, so pnpm falls back to its default resolution and can recreate `.pnpm-store` in the workspace, which is exactly what this PR set out to fix. Pointing the store at a container-local path instead of omitting the var would keep the workspace clean in both modes.

internal/runtime/entrypoint_colorterm_test.go
```diff
@@ -34,6 +34,40 @@ func TestEntrypoint_COLORTERMPreservesExplicitValue(t *testing.T) {
 	}
 }
 
+func TestEntrypointNormalizesPnpmStoreForDevcontainerUser(t *testing.T) {
```
EclipseSourceAI: This is the first test for `normalize_devcontainer_paths` and it lands in the COLORTERM file with a hand-rolled copy of the harness sitting right below it ([runEntrypointCaptureCOLORTERM](https://github.com/eclipse-enclave/enclave/blob/8a89634b22280fa322c8037d8c46be7d32d72fbf/internal/runtime/entrypoint_colorterm_test.go#L37-L66)). Move it to its own `entrypoint_devcontainer_test.go` and generalize the runner instead of duplicating the setup.

entrypoint.sh
```diff
@@ -418,6 +418,7 @@ normalize_devcontainer_paths() {
         COREPACK_ROOT
         VOLTA_HOME
         FNM_DIR
+        PNPM_CONFIG_STORE_DIR
         PNPM_STORE_DIR
```
EclipseSourceAI: `PNPM_STORE_DIR` right below is not a var pnpm reads at all (v9/v10 take `npm_config_store_dir`, v11 takes `PNPM_CONFIG_STORE_DIR`), so it is dead weight in this list. Pre-existing, but you are editing the list anyway.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Small, focused fix: `addCacheMounts` now injects `PNPM_CONFIG_STORE_DIR` pointing at the existing `~/.local/share/pnpm` cache bind mount, so pnpm stops falling back to a workspace-local `.pnpm-store`. `PNPM_CONFIG_STORE_DIR` is added to the devcontainer path rewrite list in `entrypoint.sh`, plus one shell-level and two Go unit tests. The env var is skipped under `--no-cache`. All three new tests pass locally.

Main point for a maintainer: the chosen env var only takes effect on pnpm 11+. Checked against the real CLIs, pnpm 9.15.9 and 10.34.5 ignore `PNPM_CONFIG_STORE_DIR` and honor `npm_config_store_dir`, while 11.25.0 is the reverse. The bundled pnpm is latest so the default image is fine, but corepack-pinned repos and devcontainer images on pnpm 10 or older get no fix at all. Secondary points: forcing the path bypasses pnpm's same-device check (installs copy instead of hard-link when the cache mount and workspace are on different filesystems) and overrides a repo's own `storeDir`, neither of which is documented; `--no-cache` still leaves `.pnpm-store` in the workspace; and the new entrypoint test sits in the COLORTERM test file with a duplicated harness.

tortmayr:
@EclipseSourceAI

internal/runtime/runtime.go
```diff
@@ -1096,6 +1096,11 @@ func (r *Runtime) addSessionMonitorEnv(mounts *mountAccumulator) {
 }
 
 func (r *Runtime) addCacheMounts(mounts *mountAccumulator) {
+	pnpmStoreDir := r.containerHome + "/.local/share/pnpm/store"
+	// pnpm 11 uses PNPM_CONFIG_*; pnpm 10 and older use npm_config_*.
+	mounts.AddEnv("PNPM_CONFIG_STORE_DIR", pnpmStoreDir)
+	mounts.AddEnv("npm_config_store_dir", pnpmStoreDir)
+
 	if r.run.NoCache {
 		return
```
EclipseSourceAI, line 1103: Setting `npm_config_store_dir` unconditionally makes plain `npm` (not just pnpm) print `npm warn Unknown env config \"store-dir\". This will stop working in the next major version of npm.` on every invocation, since npm doesn't recognize that key. Verified locally with `npm_config_store_dir=/tmp/foo npm install`. This adds warning noise to every npm command run in the container, scoping the var to pnpm only or filtering the warning in the entrypoint would avoid that.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Follow-up on 8ed084e, which addresses all five points from my earlier review: both `PNPM_CONFIG_STORE_DIR` and `npm_config_store_dir` are now set so the fix works across pnpm 9-11, the vars are now set even under `--no-cache` (pointing at a container-local, ephemeral path instead of falling back to workspace `.pnpm-store`), the same-device/`storeDir` override tradeoffs are documented in `docs/tools.md`, the dead `PNPM_STORE_DIR` entry was dropped from the devcontainer rewrite list, and the new test moved out of the COLORTERM file into its own `entrypoint_devcontainer_test.go` with a shared harness.

One new issue from this commit: setting `npm_config_store_dir` unconditionally makes plain `npm` print an "Unknown env config" warning on every invocation inside the container, not just pnpm. Left an inline comment with a local repro.

- [PNPM_CONFIG_STORE_DIR only works on pnpm 11+, both env var formats now needed](https://github.com/eclipse-enclave/enclave/pull/64#discussion_r3925405805)
- [forcing the store path bypasses same-device hardlinking and overrides project storeDir undocumented](https://github.com/eclipse-enclave/enclave/pull/64#discussion_r3925405969)
- [--no-cache dropped the var entirely, letting .pnpm-store reappear in the workspace](https://github.com/eclipse-enclave/enclave/pull/64#discussion_r3925406100)
- [new devcontainer test duplicated the COLORTERM test harness instead of its own file](https://github.com/eclipse-enclave/enclave/pull/64#discussion_r3925406246)
- [PNPM_STORE_DIR in the rewrite list is a var pnpm never reads](https://github.com/eclipse-enclave/enclave/pull/64#discussion_r3925406470)

tortmayr:
@EclipseSourceAI

entrypoint.sh
```diff
@@ -511,6 +511,35 @@ normalize_devcontainer_paths() {
...
+    local shim="$shim_dir/pnpm"
+    mkdir -p "$shim_dir"
+    {
+        printf '%s\n' '#!/bin/sh'
+        printf '%s\n' '# Copyright (C) 2026 EclipseSource GmbH and others.'
+        printf '%s\n' '#'
+        printf '%s\n' '# This program and the accompanying materials are made available under the'
+        printf '%s\n' '# terms of the MIT License, which is available in the project root.'
+        printf '%s\n' '#'
+        printf '%s\n' '# SPDX-License-Identifier: MIT'
```
EclipseSourceAI: The full copyright/license header on this generated shim seems unnecessary. It's a throwaway script written to `$TMPDIR` at container start and never persisted or shipped, so it isn't a "source file" in the sense AGENTS.md's header rule targets. Would just drop it and keep the shim to the shebang + exec line.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Since my last review (8ed084e), the last open point (`npm_config_store_dir` triggering an "Unknown env config" warning on plain `npm`) is fixed in 6fdfc20: that env var is dropped entirely, and instead a generated `pnpm` shim on `PATH` passes `--store-dir` as a CLI flag to the real binary. That's a solid approach, it covers pnpm 9-11 uniformly since `--store-dir` is a stable CLI flag, and the new `entrypoint_pnpm_test.go` confirms the env var no longer leaks into either the pnpm child or the parent shell. `go build`, `go test ./internal/runtime/...`, and `shellcheck` all pass clean on the current state.

One small nit left inline: the generated shim carries a full copyright header for a throwaway file that's never persisted.

- [npm prints an Unknown env config warning because npm_config_store_dir is set unconditionally](https://github.com/eclipse-enclave/enclave/pull/64#discussion_r3933150075)

tortmayr:
@EclipseSourceAI

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Only change since my last review (6fdfc202) is dropping the copyright/license header from the generated pnpm shim in `entrypoint.sh` (commit 4503dab), which is exactly what I asked for. The subsequent merge commit (39900d0) only pulls in unrelated main-branch changes (zsh completions, session-name resolution refactor); it doesn't touch the pnpm store logic. No new issues found.

- [full copyright header on the generated pnpm shim is unnecessary for a throwaway file](https://github.com/eclipse-enclave/enclave/pull/64#discussion_r3933304090)

entrypoint.sh
```diff
@@ -511,6 +511,29 @@ normalize_devcontainer_paths() {
...
+        return
+    fi
+
+    local shim_dir="${TMPDIR:-/tmp}/enclave-pnpm-bin"
+    local shim="$shim_dir/pnpm"
+    mkdir -p "$shim_dir"
+    {
+        printf '%s\n' '#!/bin/sh'
+        # shellcheck disable=SC2016 # Expanded by the generated shim at runtime.
+        printf 'exec %q --store-dir="${PNPM_CONFIG_STORE_DIR:-}" "$@"\n' "$pnpm_bin"
```
xai: pnpm validates options per command and `store-dir` is only accepted by install-like commands and `store`. With this shim on PATH, `pnpm run <script>`, `pnpm test`, `pnpm exec ...`, `pnpm ls`, `pnpm why`, and `pnpm config ...` all fail with `ERROR Unknown option: 'store-dir'` and exit 1. Verified against pnpm 11.23.0: `pnpm --store-dir=/x run hello` fails, while `pnpm --config.store-dir=/x run hello`, `exec`, `store path`, `install --offline`, and `-v` all succeed. Change the shim to `exec %q --config.store-dir="${PNPM_CONFIG_STORE_DIR:-}" "$@"`, and extend the entrypoint test (or a real-pnpm smoke test) to cover at least one non-install subcommand, since the current fake-pnpm test only checks `install` and cannot detect this.

entrypoint.sh
```diff
@@ -511,6 +511,29 @@ normalize_devcontainer_paths() {
...
+setup_pnpm_store_shim() {
+    if [ -z "${PNPM_CONFIG_STORE_DIR:-}" ]; then
+        return
+    fi
+
+    local pnpm_bin=""
+    pnpm_bin="$(command -v pnpm || true)"
+    if [ -z "$pnpm_bin" ]; then
+        return
+    fi
```
xai: `command -v pnpm` runs once at entrypoint time and the shim hard-codes that binary path. pnpm made available later (`corepack enable`, `npm i -g pnpm`, or a node version from the persistent `~/.nvm/versions` cache activated by the login shell) gets no shim at all, so the mount-boundary fallback silently recreates `.pnpm-store` in the project, which is the bug this change fixes. Switching node versions via nvm after start keeps the shim pinned to the old pnpm. Prefer resolving lazily: always write the shim, and inside it look up the real pnpm at call time with the shim dir removed from PATH (`PATH="${PATH#"$shim_dir:"}" command -v pnpm`), exiting 127 when none is found. Note that pnpm 11 already honors PNPM_CONFIG_STORE_DIR natively (verified with `pnpm store path`), so a comment stating which pnpm versions the shim exists for would make the compatibility intent clear.

xai (commented):
Thanks for the PR, looks pretty good!

Fable found two things that might be worth fixing.

entrypoint.sh
tortmayr: Thanks, I also encountered this issue in practice and already pushed a fix for that.

tortmayr:
@EclipseSourceAI

entrypoint.sh
```diff
@@ -511,6 +511,38 @@ normalize_devcontainer_paths() {
...
+        existing="$(grep -v '^[[:space:]]*store-dir[[:space:]]*=' "$npmrc" || true)"
+    fi
+
+    # ~/.npmrc is a bind-mounted file, so rewrite it in place; replacing the
+    # inode via mktemp+mv fails on the mount.
+    {
+        if [ -n "$existing" ]; then
+            printf '%s\n' "$existing"
+        fi
+        printf 'store-dir=%s\n' "$store_dir"
```
EclipseSourceAI: This reintroduces the exact warning the earlier `npm_config_store_dir` env var caused, just via `.npmrc` instead. Verified locally: writing `store-dir=/tmp/foo` to `~/.npmrc` and running `npm run test` (or any npm command that loads project config) prints `npm warn Unknown user config "store-dir". This will stop working in the next major version of npm.` on every invocation, same as before. pnpm needs this key, but plain npm doesn't recognize `store-dir` any more than it recognized `npm_config_store_dir`.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Since my last review (4503dab), the maintainer's `a479f27` dropped the generated `pnpm` shim entirely and instead writes `store-dir=...` to `~/.npmrc` (kept in sync with `PNPM_CONFIG_STORE_DIR`, idempotent across restarts). That's a good change: it sidesteps both open points from xai's review in one move, since there's no longer a shim on `PATH` to break on non-install pnpm subcommands or to get pinned to a pnpm resolved before a later `corepack`/`nvm` switch.

One new issue: writing `store-dir` to `~/.npmrc` reintroduces the same "Unknown user config" warning that the earlier `npm_config_store_dir` env var caused, just through the config file instead of the environment. Verified locally, left inline.

All of my own earlier threads are already marked resolved by the maintainer, nothing further to add there.

entrypoint.sh
tortmayr: Good catch. Dropped the shim entirely instead of making it lazy.

pnpm 9/10 read store-dir from ~/.npmrc (pnpm 11 ignores it); pnpm 11 honors PNPM_CONFIG_STORE_DIR natively (9/10 ignore it). Setting both covers all three with no PATH dependency, so corepack, late global installs, and nvm use are all handled.

On the lazy-shim suggestion: ${PATH#"$shim_dir:"} only strips a leading match, and nvm use prepends ahead of the shim dir, so the shim would resolve to itself and exec in a loop. For the same reason the nvm case bypasses the shim rather than pinning it to the old pnpm.

Verified against pnpm 9.15.9 / 10.34.5 / 11.26.0 with pnpm added to PATH only after startup: all three hit the cache-backed store, no .pnpm-store in the project.

entrypoint.sh
tortmayr: Confirmed,. npm 10.9.4 stays quiet, but npm 12.0.2 does warn.

Fixed by writing the key to pnpm's own global config file (${XDG_CONFIG_HOME:-$HOME/.config}/pnpm/rc) instead of ~/.npmrc. npm never reads that file, so the warning is gone, while pnpm 9/10 still pick up store-dir there and 11/12 keep using PNPM_CONFIG_STORE_DIR.

Since ~/.npmrc is a persisted mount, the entrypoint also strips a store-dir line written by the previous revision so existing sandboxes stop warning after the upgrade.

Verified with pnpm 9.15.9 / 10.34.5 / 11.26.0 / 12.3.4: all four resolve to the cache-backed store, no .pnpm-store in the project, and npm ls under npm 12.0.2 prints no unknown-config warning.

tortmayr:
@EclipseSourceAI

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Since my last review (a479f27), the only change is `store-dir` moving from `~/.npmrc` to pnpm's own global config file (`${XDG_CONFIG_HOME:-$HOME/.config}/pnpm/rc`), with a cleanup step that strips any stale `store-dir` line left in `~/.npmrc` by the earlier revision. This is exactly the fix needed: `configure_pnpm_store` and `write_store_dir_config` in `entrypoint.sh` are straightforward, the new/renamed tests in `entrypoint_pnpm_test.go` cover writing to the new file, leaving `~/.npmrc` untouched, migrating a stale entry, and idempotency, and `docs/tools.md` was updated to match. `go build`, `go test ./internal/runtime/...`, and `shellcheck entrypoint.sh` all pass clean on the current state. The shim-related concerns from xai's review are moot now since the shim was dropped entirely a commit earlier. No new issues found.

- [writing store-dir to ~/.npmrc made npm print an Unknown user config warning on every invocation](https://github.com/eclipse-enclave/enclave/pull/64#discussion_r3957811930)

xai (approved):
Thanks for the quick fixes!

PR #65, opened 2026-09-02
docs: add a configuration page to the website

planger:
The website docs had nothing on configuration. This adds a Configuration page covering what a user actually reaches for in daily use:

- the three layers (CLI flag, project config, global config) and how to find both files with `enclave config`
- precedence, and the `--view source|effective|diff` modes for debugging a surprising value
- the options most users end up setting, grouped by agent/session, files, network, and auth
- what project config deliberately cannot do, and why
- features (`+`/`-` syntax), per-tool overrides, and reusing host agent config

Details are intentionally left out and cross-referenced to `docs/configuration.md`, `docs/cli-reference.md`, `docs/networking.md`, and `docs/auth.md`.

Every claim on the page was checked against the CLI (defaults, flag parsing, precedence layers, each guardrail warning, feature resolution, passthrough allow-list resolution) and against a real session for the mount promises.

While verifying, `docs/configuration.md` turned out to list only `allow_all_network` and `allow_domains` as guarded at project scope, while the loader also guards `tool`, `yolo`, `pass_env`, `host_config`, `host_config_paths`, `base_image`, `bridge_ports`, out-of-project `add_dirs`, `project_mount="writable"`, and `worktree_metadata` relaxations. That paragraph is corrected here so both pages agree.

The CLI page now also links `docs/cli-reference.md`, since it documents only the core commands and previously offered no way to the rest.

One change reaches beyond this page: Infima sets `display: block` on tables, so rows shrank to fit their content while the table's border box still spanned the full width, leaving header backgrounds short of the border. `.markdown table` is now laid out as a table at full width, which affects every table on the site. The trade-off is that wide tables wrap their cells instead of scrolling horizontally; worth a look on a narrow screen before merging.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -27,10 +27,14 @@ Security guardrail: project config cannot elevate guarded options. Each is
 ignored with a warning naming the file; use global config or CLI flags for
 explicit host opt-in. Guarded at project scope: `tool`, `yolo`,
 `allow_all_network=true`, `allow_domains`, `pass_env`, `host_config`,
-`host_config_paths` (including under `tool_overrides.<tool>`), `base_image`,
-`bridge_ports`, `add_dirs`/`add_readonly_dirs` entries outside the project
-subtree, `project_mount="writable"`, and any `worktree_metadata` value that
-relaxes the inherited mode.
+`tool_overrides.<tool>.host_config_paths`, `base_image`, `bridge_ports`,
+`add_dirs`/`add_readonly_dirs` entries outside the project subtree,
+`project_mount="writable"`, and any `worktree_metadata` value that relaxes the
+inherited mode.
+
+Top-level `host_config_paths` is not part of that set: it is ignored with a
+warning in any config file, global included, because passthrough paths are only
+supported under `tool_overrides.<tool>`.
 
 ## Config Keys
 
@@ -40,7 +44,7 @@ relaxes the inherited mode.
 | `backend` | Isolation backend (`docker`, default; experimental `qemu`) |
 | `host_config` | `none` (default) or `passthrough` |
 | `tool_overrides.<tool>.host_config_paths` | Per-tool passthrough path directives (`default`, `+path`, `-path`, or explicit list) |
-| `yolo` | Enable YOLO mode (default: `true`) |
+| `yolo` | Enable YOLO mode (default: `true`). Only consulted when the selected tool's profile leaves `yoloEnabled` unset; every bundled CLI agent sets it, so use `--yolo`/`--no-yolo` for those |
 | `ephemeral` | Run without persistent auth/env stores |
 | `auth_scope` | `shared` or `project` (default: `shared`) |
 | `auth_name` | Named per-tool shared auth identity slug; unset uses the default store |
@@ -95,7 +99,7 @@ option has no effect.
 ```json
 {
   "tool": "codex",
-  "yolo": false,
+  "project_mount": "readonly",
   "secrets_scope": "project",
   "pass_env": ["GITHUB_TOKEN"],
   "allow_all_network": false
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -23,7 +23,14 @@ substitute it for `~/.config/enclave/` in every path below.
 4. Global config (`~/.config/enclave/config.json`)
 5. Built-in defaults
 
-Security guardrail: project config cannot elevate guarded options such as `allow_all_network=true` or `allow_domains`. Those values are ignored with a warning; use global config or CLI flags for explicit host opt-in.
+Security guardrail: project config cannot elevate guarded options. Each is
+ignored with a warning naming the file; use global config or CLI flags for
+explicit host opt-in. Guarded at project scope: `tool`, `yolo`,
+`allow_all_network=true`, `allow_domains`, `pass_env`, `host_config`,
+`host_config_paths` (including under `tool_overrides.<tool>`), `base_image`,
+`bridge_ports`, `add_dirs`/`add_readonly_dirs` entries outside the project
+subtree, `project_mount="writable"`, and any `worktree_metadata` value that
+relaxes the inherited mode.
 
 ## Config Keys
 
--- a/website/docs/docs/configuration.md
+++ b/website/docs/docs/configuration.md
@@ -18,7 +18,7 @@ A setting can come from three places:
 | Project config | `~/.config/enclave/projects/<hash>/config.json` | one repository |
 | Global config | `~/.config/enclave/config.json` | your defaults everywhere |
 
-Both files are plain JSON. Every key is optional, and every key has a matching
+Both files are plain JSON. Every key is optional, and most keys have a matching
 CLI flag:
 
 ```json
@@ -87,7 +87,7 @@ enclave config --json            # the same data, machine-readable
 | Key | Flag | What it does |
 | --- | --- | --- |
 | `tool` | `--tool <name>` | Which agent runs. Defaults to `claude`; `enclave tools` lists what is installed. |
-| `yolo` | `--no-yolo` | Full autonomy, on by default. Turn it off to make the agent ask for confirmation again. |
+| `yolo` | `--no-yolo` | Full autonomy, on by default. Pass `--no-yolo` to make the agent ask for confirmation again. Every bundled CLI agent pins the mode in its own profile, so `"yolo": false` in a config file does not reach them; use the flag. |
 | `features` | `--features <list>` | Extra tooling baked into the image (see [Features](#features)). |
 | `ports` | `-p 3000` | Publish a container port to the host, so the agent's dev server is reachable from your browser. |
 | `bridge_ports` | `--bridge-port 9800` | The other direction: make a host service reachable at `localhost:9800` inside the container. |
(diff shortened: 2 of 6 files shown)


website/docs/docs/configuration.md
```diff
@@ -0,0 +1,232 @@
...
+```
+
+## Most commonly used options
+
+### Agent and session
+
+| Key | Flag | What it does |
+| --- | --- | --- |
+| `tool` | `--tool <name>` | Which agent runs. Defaults to `claude`; `enclave tools` lists what is installed. |
+| `yolo` | `--no-yolo` | Full autonomy, on by default. Turn it off to make the agent ask for confirmation again. |
```
EclipseSourceAI: `yolo` in a config file is a no-op for every bundled tool. `resolveYoloEnabled` only honors the config value when the profile leaves `yoloEnabled` unset ([code](https://github.com/eclipse-enclave/enclave/blob/8a89634b22280fa322c8037d8c46be7d32d72fbf/internal/app/command_run.go#L310-L326)), and every shipped spec sets it, e.g. Claude ([spec](https://github.com/eclipse-enclave/enclave/blob/8a89634b22280fa322c8037d8c46be7d32d72fbf/extensions/tools/claude/spec.yaml#L22-L23)). So `"yolo": false` is silently ignored and only `--no-yolo` works, which this row promises the opposite of.

website/docs/docs/configuration.md
```diff
@@ -0,0 +1,232 @@
...
+A setting can come from three places:
+
+| Layer | Where | Use it for |
+| --- | --- | --- |
+| CLI flag | `enclave --tool codex` | a single run |
+| Project config | `~/.config/enclave/projects/<hash>/config.json` | one repository |
+| Global config | `~/.config/enclave/config.json` | your defaults everywhere |
+
+Both files are plain JSON. Every key is optional, and every key has a matching
+CLI flag:
```
EclipseSourceAI: "every key has a matching CLI flag" is not true: `host_config_paths` has no CLI flag ([options_def.go](https://github.com/eclipse-enclave/enclave/blob/8a89634b22280fa322c8037d8c46be7d32d72fbf/internal/config/options_def.go#L155-L163)) and is dropped unless it sits under `tool_overrides.<tool>` ([config.go](https://github.com/eclipse-enclave/enclave/blob/8a89634b22280fa322c8037d8c46be7d32d72fbf/internal/config/config.go#L306-L316)), which the passthrough section further down depends on. Drop the claim or qualify it with "most".

website/docs/docs/configuration.md
```diff
@@ -0,0 +1,232 @@
...
+| Ignored in project config | Why |
+| --- | --- |
+| `tool` | Choosing the agent stays a user decision. Pass `--tool` or set it globally. |
+| `yolo` | A repository cannot decide how much autonomy its agent gets. |
+| `allow_all_network`, `allow_domains` | A repository cannot widen its own network allowlist. |
+| `pass_env` | A repository cannot ask for your host environment variables. |
+| `host_config`, `tool_overrides.<tool>.host_config_paths` | A repository cannot widen host-config passthrough. |
+| `add_dirs`, `add_readonly_dirs` outside the project | Project scope can only mount subdirectories of the project itself. |
+| `base_image`, `bridge_ports` | Base image and host-port bridging stay host-side decisions. |
+| `project_mount: "writable"` | Project scope can tighten a stricter global default, never loosen it. |
```
EclipseSourceAI, line 204: `worktree_metadata` relaxations are guarded at project scope too and are missing from this table, so it does not quite agree with the corrected paragraph in `docs/configuration.md`.

docs/configuration.md
```diff
@@ -23,7 +23,14 @@ substitute it for `~/.config/enclave/` in every path below.
 4. Global config (`~/.config/enclave/config.json`)
 5. Built-in defaults
 
-Security guardrail: project config cannot elevate guarded options such as `allow_all_network=true` or `allow_domains`. Those values are ignored with a warning; use global config or CLI flags for explicit host opt-in.
+Security guardrail: project config cannot elevate guarded options. Each is
+ignored with a warning naming the file; use global config or CLI flags for
+explicit host opt-in. Guarded at project scope: `tool`, `yolo`,
+`allow_all_network=true`, `allow_domains`, `pass_env`, `host_config`,
+`host_config_paths` (including under `tool_overrides.<tool>`), `base_image`,
```
EclipseSourceAI: Top-level `host_config_paths` is not a project-scope guardrail: it is dropped in any config file, global included ([config.go](https://github.com/eclipse-enclave/enclave/blob/8a89634b22280fa322c8037d8c46be7d32d72fbf/internal/config/config.go#L306-L316)). Only `tool_overrides.<tool>.host_config_paths` is guarded here, which is how the website table words it.

website/docs/src/css/custom.css
```diff
@@ -294,7 +294,12 @@ div[class^='codeBlockContainer'],
    Tables
    ========================================================================== */
 
+/* Infima sets `display: block` on tables, which makes the rows shrink to fit
+   their content while the table's own border box still spans the full width,
+   so header backgrounds and row rules stop short of the border. */
 .markdown table {
+  display: table;
```
EclipseSourceAI, line 301: The rule already sets `overflow: hidden` ([main](https://github.com/eclipse-enclave/enclave/blob/8a89634b22280fa322c8037d8c46be7d32d72fbf/website/docs/src/css/custom.css#L297-L303)), so horizontal scrolling was gone before this change. With `display: table` the table can no longer shrink below its min-content width, so a table that is too wide now spills out of (or gets clipped by) its container on narrow viewports instead of just wrapping. Keeping the radius clipping on a wrapper and letting the table scroll would avoid the trade-off.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Adds a Configuration page to the website (three layers, precedence and `enclave config --view`, the commonly used options grouped by area, project-scope guardrails, features, per-tool overrides, host config reuse), links `docs/cli-reference.md` from the CLI page, shifts Support to `sidebar_position: 5`, corrects the guardrail paragraph in `docs/configuration.md`, and changes `.markdown table` to `display: table; width: 100%` site-wide.

I checked the page's claims against the code. Paths, hash length, flag names, feature defaults, passthrough allow-list and backstop, guardrail set, and precedence all hold up. Three things a maintainer should look at:

- The `yolo` row promises something config cannot do: `resolveYoloEnabled` ignores the config value whenever the tool profile sets `yoloEnabled`, which every shipped spec does.
- "every key has a matching CLI flag" is contradicted by `host_config_paths`, which has no flag and is only accepted under `tool_overrides.<tool>`.
- The CSS change is site-wide and rides along in a docs PR. The rule already had `overflow: hidden`, so the trade-off is not scroll versus wrap but clipping versus wrap on narrow viewports. Worth checking the preview on mobile before merging.

website/docs/docs/configuration.md
planger: Correct, and worth fixing beyond the wording. `resolveYoloEnabled` only consults the config value when `profile.YoloEnabled == nil`, and claude, codex, opencode, pi and mistral-vibe all set `yoloEnabled: true` (only theia/theia-next leave it unset), so `"yolo": false` is dropped for every CLI agent — and silently, since the warning only fires on the branch that honours the value.

Reworded the row to point at `--no-yolo` and say the config key does not reach the bundled agents. Also fixed `docs/configuration.md`, where the `yolo` key row and the `"yolo": false` example had the same problem.

The silent drop looks like a code bug rather than a docs one; I left the behaviour alone in this docs PR and will file it separately.

website/docs/docs/configuration.md
planger: Correct — `host_config_paths` has no flag. Changed to "most keys have a matching CLI flag".

website/docs/docs/configuration.md
planger: Correct, added a row: `worktree_metadata` values that relax the inherited mode.

docs/configuration.md
planger: Correct — `validateTopLevelDefaults` runs from `readDefaults`, so top-level `host_config_paths` is dropped in global config too. Moved it to `tool_overrides.<tool>.host_config_paths` in the guarded list and added a sentence explaining that the top-level key is rejected everywhere, for a different reason.

website/docs/src/css/custom.css
planger: Right that `overflow: hidden` was already there on `main`, so the trade-off is clipping vs. wrapping, not scrolling vs. wrapping.

I kept `display: table` — it is what actually fixes the bug (with `display: block` the table border box spans the full width while the anonymous inner table shrink-to-fits, so the header background stops short of the border), and there is no way to size that anonymous box from CSS. A wrapper element would need an MDX `table` override, which is more than this docs PR should carry.

Instead I removed the trade-off: `overflow-wrap: anywhere` on `th`/`td` lowers the cells min-content width, so `width: 100%` is always satisfiable and a wide table wraps rather than overflowing the clipped border box. For these prose-heavy option tables wrapping beats horizontal scrolling on mobile anyway.

planger:
As we are starting with some promotion today, I'm merging this for now. Please still review and we can make follow-up changes later on.

xai:
> As we are starting with some promotion today, I'm merging this for now. Please still review and we can make follow-up changes later on.

Thanks! I just checked and the changes look correct to me!

PR #66, opened 2026-09-03
feat: adopt the Eclipse Foundation logo on the website and README

planger:
Replaces the placeholder app icon with the official Eclipse Enclave logo across the marketing site, the Docusaurus docs, and the repository README.

## Credits

The logo was designed and contributed by the **Eclipse Foundation**. Many thanks to **Elspeth Tory and her team** for creating it and making it available to the project.

## What changed

Only the variants that something actually uses are checked in. Two of them are derived from the supplied horizontal artwork:

- **Icon-only mark** — the wordmark paths are dropped and the `viewBox` re-framed to a square around the mark, using measurements taken from the reference PNG so the framing is exact.
- **Dark-background SVGs** — only PNG was supplied for `darkbg`, so the SVG variants flip the two fill colours (`#231f20`/`#221f20` → `#ffffff`). The result was checked against the supplied dark-background PNGs.

| Location | Files | Used by |
| --- | --- | --- |
| `docs/assets/` | `enclave-logo-horizontal-{lightbg,darkbg}.png` | README, via `<picture>` for light/dark |
| `website/assets/` | `enclave-mark-lightbg.svg`, `enclave-logo-horizontal-lightbg.svg`, `favicon.png`, `social-card.png` | site navbar, new footer logo, favicon, OG card |
| `website/docs/static/img/` | `enclave-mark-{lightbg,darkbg}.svg`, `enclave-logo-horizontal-darkbg.svg`, `favicon.png`, `social-card.png` | docs navbar (`src`/`srcDark`), docs footer, favicon, OG card |

Also in scope:

- The README logo is wrapped in the `<h1>`, so the file keeps a single top-level heading without repeating the wordmark as text.
- `docs/assets/` is added to the `.dockerignore` "repo-only content" block. `docs/` feeds the runtime image build context and its hash, so README-only branding would otherwise rebuild the image for every user whenever the logo changes.
- `twitter:card` is now `summary_large_image`; the old card pointed at a square app icon.
- The superseded placeholders are removed (`docs/assets/appicon.png`, `website/assets/appicon.png`, `website/docs/static/img/logo.png`).

## Notes for review

- **The docs footer logo is deliberately unlinked.** `themeConfig.footer.logo` validates against a schema that rejects `autoAddBaseUrl`, and Docusaurus' `Link` would rewrite a `pathname://` href back under `/docs/`. Documented in `website/docs/README.md` next to the existing linking-out gotcha.
- **The vertical lockups are not included**, since nothing uses them yet.
- **Palette mismatch, not addressed here.** A comment in the docs theme claimed its navy / beam-blue / amber palette was "drawn straight from the product mark" — that referred to the old placeholder icon. The real mark is a teal-to-violet gradient. The comment is corrected to stop asserting a derivation that no longer holds, but re-theming both sites to the brand gradient is a design decision left out of this PR.
- **Logo provenance in `NOTICE.md` is intentionally untouched.** Credit for this contribution lives in the commit message and here. Whether the checked-in artwork warrants its own `NOTICE.md` entry recording origin and trademark terms is a separate IP question that applies to the logo regardless of this PR.

## Verification

`make build`, `make lint`, and `make check-license-headers` pass, and the Docusaurus site builds with the new assets resolving in the generated HTML. The `.dockerignore` exclusion was checked against the same `patternmatcher`/`ignorefile` pair the build hash uses: `docs/assets/**` is ignored while `docs/README.md` and `docs/ARCHITECTURE.md` are not.

`make test` has one failure, `TestDockerImportsConfinedToBackendAndCarveOuts`. It is pre-existing and unrelated — it also fails on a clean tree, caused by a stale checkout under `.claude/worktrees/` that the arch test scans.


website/css/styles.css
```diff
@@ -115,7 +115,8 @@ h2 + .section-lead { margin-top: 0; }
   gap: 20px;
 }
 .nav-brand { display: flex; align-items: center; gap: 10px; font-weight: 800; color: var(--heading); font-size: 1.05rem; }
-.nav-brand img { width: 28px; height: 28px; }
+/* The mark is slightly taller than wide, so pin the height and let width follow. */
+.nav-brand img { height: 30px; width: auto; }
```
EclipseSourceAI: `enclave-mark-lightbg.svg` has `viewBox="19.258 7.57 71.021 71.021"`, so it is exactly square and `width: auto` resolves to 30px. The comment's premise does not hold, and the only real change here is 28px to 30px ([old rule](https://github.com/eclipse-enclave/enclave/blob/8a89634b22280fa322c8037d8c46be7d32d72fbf/website/css/styles.css#L118)).

website/docs/docusaurus.config.js
```diff
@@ -123,6 +124,15 @@ const config = {
...
       footer: {
         style: 'dark',
+        // The footer is always dark, so it uses the dark-background lockup
+        // regardless of the active color mode. It stays unlinked: the footer
+        // logo cannot opt out of `autoAddBaseUrl`, so a `pathname://` href
+        // would be rewritten back under the docs base path.
+        logo: {
+          alt: 'Eclipse Enclave',
+          src: 'img/enclave-logo-horizontal-darkbg.svg',
+          width: 168,
```
EclipseSourceAI, line 134: Infima caps `.footer__logo` at `max-width: 10rem` (160px) and `custom.css` does not override it, so 168 renders clamped. Either drop to 160 or add the override next to the other `.footer` rules.

README.md
```diff
@@ -1,9 +1,10 @@
 <p align="center">
-  <img src="docs/assets/appicon.png" alt="Enclave" width="128" height="128">
+  <picture>
+    <source media="(prefers-color-scheme: dark)" srcset="docs/assets/enclave-logo-horizontal-darkbg.png">
+    <img src="docs/assets/enclave-logo-horizontal-lightbg.png" alt="Eclipse Enclave" width="400">
```
EclipseSourceAI, line 4: `docs/` is copied into the sandbox image and feeds the image build hash ([Dockerfile path](https://github.com/eclipse-enclave/enclave/blob/8a89634b22280fa322c8037d8c46be7d32d72fbf/docs/ARCHITECTURE.md#L162), [hash input](https://github.com/eclipse-enclave/enclave/blob/8a89634b22280fa322c8037d8c46be7d32d72fbf/internal/app/build_hash.go#L109)), so these README-only PNGs force an image rebuild for every user whenever the branding is touched. `.dockerignore` already has a "repo-only content" block ([here](https://github.com/eclipse-enclave/enclave/blob/8a89634b22280fa322c8037d8c46be7d32d72fbf/.dockerignore#L47-L50)) that `docs/assets` would fit into.

README.md
```diff
@@ -1,9 +1,10 @@
...
+  </picture>
 </p>
 
-<h1 align="center">Eclipse Enclave</h1>
```
EclipseSourceAI, line 5: Dropping the `<h1>` leaves the README with no top-level heading, so GitHub's outline and any renderer that ignores `<picture>` lose the project name. The PR description does not mention this one; a maintainer should confirm it is intended.

website/docs/README.md
```diff
@@ -42,12 +42,38 @@ relative `../` is handled client-side, matches no docs route, and renders the
...
 deploy action replaces the published tree.
 
 Both website workflows push with the organization secret `DEPLOY_TOKEN`.
 
+## Brand assets
+
+The Eclipse Foundation logo comes in a horizontal lockup (mark plus wordmark)
+and an icon-only mark, each with a variant for light and for dark backgrounds.
+The `-darkbg` variants are the same artwork with the cube faces, inner disc, and
+wordmark in white; do not mix them up, or the wordmark disappears.
```
EclipseSourceAI, line 60: The logo artwork is Eclipse Foundation branding, but `NOTICE.md` only covers the name ([Trademarks section](https://github.com/eclipse-enclave/enclave/blob/8a89634b22280fa322c8037d8c46be7d32d72fbf/NOTICE.md#L7-L9)), so the checked-in files fall under the blanket MIT declaration by default. Adding a line there recording the origin and that trademark terms apply to the marks would be cheap insurance.

website/docs/src/css/custom.css
```diff
@@ -11,8 +11,9 @@
  * Eclipse Enclave — documentation theme.
  *
  * Design direction: "The Enclave Perimeter."
- * The palette is drawn straight from the product mark — a deep navy field, an
- * electric beam-blue core, and a gateway amber. The page reads as a bounded,
+ * The palette predates the Eclipse Foundation brand logo and does not derive
+ * from it: a deep navy field, an electric beam-blue core, and a gateway amber.
+ * The logo's teal-to-violet gradient is left to the logo. The page reads as a bounded,
```
EclipseSourceAI: This line runs to 87 columns while the rest of the block wraps at ~79. Reflow it.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Swaps the placeholder app icon for the Eclipse Foundation logo across the marketing site, the Docusaurus docs, and the README, and deletes the three superseded placeholders.

I verified the asset derivation mechanically: the `-darkbg` SVGs are byte-identical to the light ones modulo `#231f20`/`#221f20` → `#ffffff`, the mark SVGs are identical between `website/assets/` and `website/docs/static/img/`, and neither SVG carries orphaned `cls-*` rules or unreferenced `clippath`/gradient ids after the wordmark was stripped. Social card is 1200x630 (matching the `summary_large_image` switch), favicon 128x128, README PNGs 800x312. No references to the deleted `appicon.png` / `img/logo.png` remain anywhere in the tree, and the duplication between `website/assets/` and `website/docs/static/img/` is required by `website/build.sh`, which copies the two trees separately.

- The `.nav-brand img` comment justifies `width: auto` with the mark being taller than wide, but the mark's `viewBox` is exactly square, so the rule is equivalent to the 28x28 it replaces.
- The docs footer `width: 168` runs into infima's `.footer__logo` cap.
- `docs/assets/` is inside the sandbox image build context and its content hash, so README branding churn invalidates the image for every user.
- The logo is Eclipse Foundation artwork now checked in under the repo's blanket MIT declaration, and `NOTICE.md` only names the trademark.

The footer-logo-stays-unlinked reasoning holds up: with `autoAddBaseUrl` on, `pathname:///` gets `withBaseUrl` applied and lands back under `/docs/`.

website/docs/docusaurus.config.js
planger: Checked this and the premise does not hold: `--ifm-footer-logo-max-width` is `min(30rem, 90vw)` in the bundled Infima, not `10rem`.

```
infima/dist/css/default/default.css:317   --ifm-footer-logo-max-width: min(30rem, 90vw);
infima/dist/css/default/default.css:2219  .footer__logo { max-width: var(--ifm-footer-logo-max-width); }
```

The built site carries the same value (`--ifm-footer-logo-max-width:min(30rem,90vw)`), so the cap is 480px and `width: 168` renders unclamped. Leaving it at 168 to match the marketing footer, which uses the same width. No change made.

website/docs/README.md
planger: Not doing this one, on maintainer instruction rather than on the merits. The decision for this PR was to keep the logo credit in the commit message and the PR description only, and to leave `NOTICE.md` and the site itself untouched.

Flagging that the underlying point is separate from attribution: whether the checked-in artwork needs its own `NOTICE.md` entry recording origin and trademark terms is an IP question for a maintainer, and it applies to the logo regardless of this PR. Worth its own issue if we want it settled.

README.md
planger: Fixed, though not quite as suggested. Rather than adding back a text `<h1>` that repeats the wordmark already in the logo, the logo now *is* the `<h1>`:

```html
<h1 align="center">
  <picture>
    ...
    <img src="..." alt="Eclipse Enclave" width="400">
  </picture>
</h1>
```

That restores the single top-level heading you flagged, and the accessible name comes from the `alt`. It also keeps the heading rule GitHub draws under an `<h1>`, which the README already had before this PR.

PR #71, opened 2026-09-10
feat: add shared skill validation modes

tortmayr:
#### What it does
Shared skills that are otherwise portable are currently skipped when they contain agent-specific metadata such
  as Claude’s `disable-model-invocation`, even though the selected agent can handle them. This adds a native
  validation mode that passes such skills through for the agent to interpret, while retaining strict validation
  as the default.

- add strict and native validation across CLI and config override layers
- keep the portable subset as the strict default; native mode delegates metadata interpretation to the selected agent
- apply validation to standalone mounts and composed tool config

#### How to test
Add a skill 

 1. Add a shared skill under `~/.config/enclave/skills/` containing `disable-model-invocation`.
  2. Start Enclave with the default strict validation and verify that the skill is skipped.
  3. Set `"skills_validation": "native"` in Enclave’s configuration.
  4. Start Enclave again and verify that the skill is loaded unchanged.
#### Breaking changes

#### Review checklist


internal/app/validation.go
```diff
@@ -71,6 +71,14 @@ func ValidateOptions(opts model.Options, sources model.OptionSources, ctx Valida
...
 		return opts, sources, warnings, fmt.Errorf("devcontainer mode is mutually exclusive with --base-image")
 	}
+	if opts.SkillsValidation == "" {
+		opts.SkillsValidation = model.SkillsValidationStrict
+	}
+	switch opts.SkillsValidation {
+	case model.SkillsValidationStrict, model.SkillsValidationNative:
+	default:
+		return opts, sources, warnings, fmt.Errorf("--skills-validation must be strict or native")
+	}
```
EclipseSourceAI: The default and the enum check sit inline here, while the other string enum run options normalize in `normalizeOptions` ([main](https://github.com/eclipse-enclave/enclave/blob/1fcda88a958ac8dcbb1e223580b8812ef8676d8c/internal/app/validation.go#L363-L374)). Result: `--project-mount READONLY` is accepted but `--skills-validation NATIVE` errors out. Trim/lower/default there instead, the default is already in `DefaultOptions()`.

internal/runtime/skills_mount.go
```diff
@@ -141,8 +143,24 @@ func overlaySkillSource(targetDir string, sourceDir string) error {
...
+	return overlayValidatedSkillSource(targetDir, sourceDir, sharedSkillValidator(validation))
+}
+
+func sharedSkillValidator(validation string) func(string, string) error {
+	return func(skillDir string, directoryName string) error {
+		content, err := readSkillFile(skillDir)
+		if err != nil {
+			return err
+		}
+		if validation == model.SkillsValidationNative {
```
EclipseSourceAI: Raw comparison makes this the third place that encodes "empty means strict" (here, `ValidateOptions`, `DefaultOptions`). `model.WorktreeMetadataMode` ([main](https://github.com/eclipse-enclave/enclave/blob/1fcda88a958ac8dcbb1e223580b8812ef8676d8c/internal/model/worktree_metadata.go#L12-L24)) is the existing pattern for this and the runtime consumes it that way ([main](https://github.com/eclipse-enclave/enclave/blob/1fcda88a958ac8dcbb1e223580b8812ef8676d8c/internal/runtime/runtime.go#L1052)), so a `model.SkillsValidationMode` resolver would handle casing, empty, and unknown values in one spot.

internal/runtime/skills_mount.go
```diff
@@ -141,8 +143,24 @@ func overlaySkillSource(targetDir string, sourceDir string) error {
...
+			return nil
+		}
+		if err := validatePortableSkillContent(content, directoryName); err != nil {
+			return fmt.Errorf("%w; use skills_validation=native to delegate metadata validation to the agent", err)
+		}
```
EclipseSourceAI: The hint is appended to every strict failure, so a skill that just misses a `description` or has a typo in `name` gets told to turn validation off globally. Scope it to the unknown-field case, which is what native mode is actually for. The missing-SKILL.md path already returns before the wrap, so the precedent is there.

internal/runtime/skills_mount.go
```diff
@@ -180,30 +198,34 @@ type portableSkillMetadata struct {
...
+
+func validatePortableSkillContent(content []byte, directoryName string) error {
 	frontmatter, err := extractSkillFrontmatter(content)
 	if err != nil {
 		return err
 	}
 	var metadata portableSkillMetadata
 	if err := yaml.UnmarshalStrict(frontmatter, &metadata); err != nil {
-		return fmt.Errorf("frontmatter must use only portable fields (name, description, license, compatibility, metadata): %w", err)
+		return fmt.Errorf("invalid portable skill frontmatter: %w", err)
```
EclipseSourceAI, line 231: This drops the field list from the message users hit on the default path. The warning is now `invalid portable skill frontmatter: error unmarshaling JSON: while decoding JSON: json: unknown field "allowed-tools"`, where the old one named the portable fields ([main](https://github.com/eclipse-enclave/enclave/blob/1fcda88a958ac8dcbb1e223580b8812ef8676d8c/internal/runtime/skills_mount.go#L206)). Keep the list and use a separate message for unparseable YAML.

docs/configuration.md
```diff
@@ -225,9 +226,20 @@ With `host_config=passthrough`, every built-in skill-capable tool passes its nat
...
+`skills_validation` controls validation of global and project shared skills:
 
-Built-in skills remain tool-specific so extensions can carry harness-specific metadata. Skills shipped by enabled features overlay as trusted extension content and skip the portable-skill validation applied to shared skills. Tools without `sandbox.skillsDir` ignore all shared skill sources.
+- `strict` (default): `SKILL.md` must be a regular, readable file with YAML frontmatter containing required `name` and `description` fields and optional `license`, `compatibility`, and `metadata` fields. The name must match the directory and use lowercase letters, numbers, and hyphens. Other fields, including `allowed-tools` and `disable-model-invocation`, cause the skill to be skipped.
+- `native`: require a regular, readable `SKILL.md`, then copy its bytes and supporting files unchanged. Enclave does not parse the frontmatter, enforce name matching, strip fields, or translate metadata between agents. The selected agent validates and interprets the skill; metadata accepted by one agent may be ignored or rejected by another.
```
EclipseSourceAI: `native` overloads the meaning this doc already gives the word, the tool's own on-disk layout ("native host skills directory", "native config path", [main](https://github.com/eclipse-enclave/enclave/blob/1fcda88a958ac8dcbb1e223580b8812ef8676d8c/docs/configuration.md#L238)). `agent` or `passthrough` avoids the clash, and this is a config value that gets expensive to rename later. Maintainer's call.

docs/configuration.md
```diff
@@ -225,9 +226,20 @@ With `host_config=passthrough`, every built-in skill-capable tool passes its nat
...
+
+Enclave warns and skips invalid shared skills rather than failing the session. A skipped higher-precedence skill leaves the lower-precedence skill intact. Tool-specific overrides, host passthrough, and extension skills are unaffected by this option.
```
EclipseSourceAI, line 260: Add that overrides still key on the directory name: in native mode two shared skill dirs can declare the same frontmatter `name` and both end up in the mount, which strict mode ruled out via the name/directory match.

docs/configuration.md
```diff
@@ -225,9 +226,20 @@ With `host_config=passthrough`, every built-in skill-capable tool passes its nat
...
+
+Set `"skills_validation": "native"` in global or project `config.json`, or under `tool_overrides.<tool>`. The normal option precedence applies; `--skills-validation strict|native` takes priority over config files. Project configuration may select `native` because it is stored in Enclave's host-managed configuration root and does not expose additional host paths.
+
+In both modes, symlinked skill directories and symlinks inside skills are ignored; `SKILL.md` itself must be a regular file.
+
+Shared skills do not require host passthrough. If passthrough remains enabled for other settings, exclude its native skills directory with the per-tool `host_config_paths` directives above.
```
EclipseSourceAI: This paragraph is about host passthrough, not `skills_validation`, and "its native skills directory" now reads as the new mode. Drop it or fold it into the passthrough section above.

internal/config/options_def.go
```diff
@@ -152,6 +152,29 @@ func OptionDefs() []OptionDef {
 				},
 			},
 		},
+		{
+			Name:          "skills_validation",
```
EclipseSourceAI, line 165: Nit: this splits `host_config` from `host_config_paths`, and `OptionDefs` order is what `enclave config` prints. Moving it after `host_config_paths` keeps the pair adjacent.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Adds a `skills_validation` run option (`strict` default, `native`) so shared skills carrying agent-specific frontmatter such as `disable-model-invocation` are passed through instead of skipped. Plumbing follows the existing `host_config` shape (option def, generated CLI/registry code, `Defaults`, sources) and the mode is threaded into both skill composition paths, the standalone mount and the composed tool config source. `make generate` produces no drift, and build plus the touched package tests pass locally.

Main points for a maintainer:

- The mode is resolved in three places (`DefaultOptions`, `ValidateOptions`, and a raw string compare in `sharedSkillValidator`) and, unlike the sibling enum options, is not case-normalized. A `model.SkillsValidationMode` resolver like `model.WorktreeMetadataMode` would collapse that.
- The `use skills_validation=native` hint is appended to every strict failure, including plain mistakes like a missing description, which pushes users to disable validation for the wrong reason.
- Naming: `native` clashes with the meaning `native` already has in these docs (the tool's own on-disk config layout). Worth settling before the config value ships.
- The mode is a session-wide switch, so one non-portable skill takes all shared skills out of validation. Fine if that is the intended trade-off, but it is the main behavioral decision in the PR.
- Unlike `host_config`, `skills_validation` is deliberately allowed in project configs. That looks defensible since project shared skills already live in the host-managed config root and project tool-specific skills bypass validation entirely today, but it is a policy call worth a second look.
- The branch does not merge into main: the only conflict is `internal/model/option_sources_gen.go`, which is regenerated content.

tortmayr:
@EclipseSourceAI

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Follow-up review. All eight points from the earlier review are addressed in the latest commit:

- `skills_validation` is now normalized (trim/lowercase, empty falls back to strict) in `normalizeOptions` before the enum check in `ValidateOptions`, matching the other string-enum options.
- `model.SkillsValidationMode` was added as the single resolver, used by both `sharedSkillValidator` and the option layers, mirroring the `WorktreeMetadataMode` pattern.
- The "delegate to agent mode" hint is now scoped to the unsupported-frontmatter-fields error only, not appended to every strict failure (missing description, name mismatch, etc. no longer carry it).
- The unsupported-fields error again lists the offending field names and the allowed set, instead of falling through to the raw YAML decode error.
- The value was renamed from `native` to `agent`, avoiding the clash with the existing "native host config" terminology in the docs.
- `docs/configuration.md` now states that overrides key on directory name in both modes and that `agent` mode can let two directories share a frontmatter `name` without deduplication.
- The stray passthrough paragraph that read ambiguously next to the new option was cleaned up.
- `skills_validation` was moved next to `host_config_paths` in `OptionDefs`, matching `enclave config` ordering.

Test coverage was extended accordingly (case-insensitivity, unknown values, hint scoping, missing-file vs invalid-frontmatter diagnostics). Build and the touched package tests pass locally. No new issues found in the diff since the last review; nothing further to flag.

- [normalize skills_validation like the other enum options](https://github.com/eclipse-enclave/enclave/pull/71#discussion_r3980004439)
- [use a model.SkillsValidationMode resolver instead of raw string comparisons](https://github.com/eclipse-enclave/enclave/pull/71#discussion_r3980004651)
- [scope the "use native mode" hint to the unsupported-field case](https://github.com/eclipse-enclave/enclave/pull/71#discussion_r3980004877)
- [keep the field list in the unsupported-frontmatter-fields error message](https://github.com/eclipse-enclave/enclave/pull/71#discussion_r3980005073)
- [rename "native" to avoid clashing with the existing native-config-layout meaning](https://github.com/eclipse-enclave/enclave/pull/71#discussion_r3980005253)
- [document that overrides still key on directory name, including the agent-mode duplicate-name case](https://github.com/eclipse-enclave/enclave/pull/71#discussion_r3980005453)
- [drop or move the passthrough paragraph that now reads ambiguously](https://github.com/eclipse-enclave/enclave/pull/71#discussion_r3980005630)
- [reorder skills_validation next to host_config_paths in OptionDefs](https://github.com/eclipse-enclave/enclave/pull/71#discussion_r3980005799)

xai (approved):
Works fine, code looks good!

PR #73, opened 2026-09-10
build: install npm tools with --ignore-scripts where upstream supports it

### LINKED ISSUE #72: Install npm tools with --ignore-scripts where upstream supports it
#### Problem

`enclave-install-npm-tool` ends at `npm install -g --no-audit --no-fund
--progress=false <package>` (`runtime-assets/build-scripts/bin/enclave-agent-npm-install:75`).
Dependency lifecycle scripts run at image build time. For most of our tools
they do nothing we need.

#### What each tool needs

Checked against upstream sources and verified by installing each package with
`--ignore-scripts` into a temp prefix and running the binary.

| Tool | Package | `--ignore-scripts` |
| --- | --- | --- |
| pi | `@earendil-works/pi-coding-agent` | yes |
| codex | `@openai/codex` | yes |
| claude | `@anthropic-ai/sandbox-runtime` | yes |
| opencode | `opencode-ai` | **no, breaks the install** |

**pi.** No `preinstall`/`install`/`postinstall`. Clipboard
(`@mariozechner/clipboard-linux-x64-gnu`) and TUI native modules ship as
prebuilt `.node` files. Upstream documents the flag in the package README and
passes it themselves in `pi update --self`. `pi --version` works.

**codex.** Neither `@openai/codex` nor its platform packages declare a
lifecycle script; `bin/codex.js` resolves the vendored binary from the optional
dependency at runtime. The installed tree is two packages with zero scripts, so
this is preventive rather than a fix. `codex --version` works.

**sandbox-runtime.** No install-time scripts in the package or its dependency
tree. `prepare: husky` only runs for git and link installs, not registry
tarballs. `srt --help` works.

**opencode.** `opencode-ai` declares `postinstall: node ./postinstall.mjs`, and
that script is what puts the binary in place. Its `bin/opencode.exe` is a stub
whose only job is to print "opencode-ai's postinstall script was not run" and
exit 1. Postinstall picks among twelve platform packages based on musl and AVX2
detection. Bypassing it means reimplementing that selection.

#### Change

Let `enclave-install-npm-tool` accept `-`-prefixed arguments before `<package>`
and forward them. `enclave-agent-npm-install` already handles this: it forwards
`"$@"` to npm (lines 75 and 84) and its shebang-rewrite loop skips arguments
beginning with `-`.

Then:

```sh
# extensions/tools/pi/install.sh
enclave-install-npm-tool --ignore-scripts @earendil-works/pi-coding-agent@latest pi Pi

# extensions/tools/codex/install.sh
enclave-install-npm-tool --ignore-scripts @openai/codex@latest codex Codex

# extensions/tools/claude/install.sh
enclave-agent-npm-install --ignore-scripts @anthropic-ai/sandbox-runtime
```

`extensions/tools/opencode/install.sh` stays as it is.

Keep this opt-in per call site. A repo-wide default with per-tool opt-outs would
break opencode, and opencode is the only npm tool we install that needs scripts.

#### Acceptance criteria

- `enclave-install-npm-tool` forwards npm flags and keeps its current behavior
  for callers that pass none.
- pi, codex, and claude install with `--ignore-scripts`; opencode installs
  unchanged.
- Images build for all four tools and each binary runs. pi clipboard still
  works in a session.
- The usage string and `docs/extensions/adding-a-tool.md` cover the
  pass-through.

xai:
#### What it does

Fixes #72.

Dependency lifecycle scripts ran at image build time for no functional gain. enclave-install-npm-tool now forwards leading npm flags, and pi, codex, and claude's sandbox-runtime pass --ignore-scripts.

opencode is excluded: its postinstall is what places the platform binary, and the published bin stub fails with an explicit error without it.

#### How to test

Run enclave using this branch with tools installed via npm, i.e., pi, codex, and opencode. Verify that they still install and start.

#### Breaking changes

#### Review checklist

### DIFF
--- a/docs/extensions/adding-a-tool.md
+++ b/docs/extensions/adding-a-tool.md
@@ -37,6 +37,22 @@ sandbox:
   settingsTarget: .<tool>/settings.json
 ```
 
+### Installing from npm
+
+For a tool that is a plain npm package, `install.sh` can delegate to the shared
+helper, which installs through the private agent Node runtime and verifies the
+binary landed on `PATH`:
+
+```sh
+enclave-install-npm-tool [npm-flag ...] <package> <binary> [label]
+```
+
+Leading `-`-prefixed arguments are forwarded to `npm install`. Pass
+`--ignore-scripts` unless the package needs its lifecycle scripts, so
+dependency-authored code does not run at image build time. Check upstream
+first: some packages resolve a platform binary in a `postinstall` and break
+without it.
+
 ## 2) Fill in the `sandbox` block
 
 `sandbox.*` is enclave-native tool metadata. Common fields (see the README for
--- a/docs/extensions/adding-a-tool.md
+++ b/docs/extensions/adding-a-tool.md
@@ -47,8 +47,11 @@ binary landed on `PATH`:
 enclave-install-npm-tool [npm-flag ...] <package> <binary> [label]
 ```
 
-Leading `-`-prefixed arguments are forwarded to `npm install`. Pass
-`--ignore-scripts` unless the package needs its lifecycle scripts, so
+Leading `-`-prefixed arguments are forwarded to `npm install`, and must be
+self-contained — `--flag` or `--flag=value`. A flag that takes a separate value
+(`--omit dev`) would consume the package argument.
+
+Pass `--ignore-scripts` unless the package needs its lifecycle scripts, so
 dependency-authored code does not run at image build time. Check upstream
 first: some packages resolve a platform binary in a `postinstall` and break
 without it.
--- a/extensions/tools/claude/install.sh
+++ b/extensions/tools/claude/install.sh
@@ -24,26 +24,12 @@ claude --version
 enclave-agent-npm-install --ignore-scripts @anthropic-ai/sandbox-runtime
 
 # Nothing else in the tree references sandbox-runtime, so an incomplete install
-# would only surface at runtime as degraded sandbox support. Assert the package
-# landed and that every bin it declares exists, which is what would break if
-# upstream started placing those in a postinstall.
+# would only surface at runtime as degraded sandbox support. enclave-agent-npm-install
+# already fails on a declared bin that did not land; assert the package itself
+# did, which it only warns about.
 sandbox_runtime_dir="${npm_config_prefix:-$HOME/.local}/lib/node_modules/@anthropic-ai/sandbox-runtime"
 if [ ! -f "$sandbox_runtime_dir/package.json" ]; then
     echo "sandbox-runtime install failed: $sandbox_runtime_dir/package.json not found" >&2
-    exit 1
-fi
-
-missing_bins=0
-while IFS= read -r bin_path; do
-    [ -n "$bin_path" ] || continue
-    if [ ! -e "$sandbox_runtime_dir/$bin_path" ]; then
-        echo "sandbox-runtime install incomplete: declared bin $bin_path is missing" >&2
-        missing_bins=1
-    fi
-done < <(jq -r '(.bin // empty) | if type == "string" then . else .[] end' "$sandbox_runtime_dir/package.json")
-
-if [ "$missing_bins" -ne 0 ]; then
-    echo "sandbox-runtime may require lifecycle scripts; re-check the --ignore-scripts above" >&2
     exit 1
 fi
 
--- a/extensions/tools/claude/install.sh
+++ b/extensions/tools/claude/install.sh
@@ -22,3 +22,29 @@ claude --version
 # Install sandbox runtime for Claude Code sandbox support.
 # It ships its vendored artifacts prebuilt and declares no lifecycle scripts.
 enclave-agent-npm-install --ignore-scripts @anthropic-ai/sandbox-runtime
+
+# Nothing else in the tree references sandbox-runtime, so an incomplete install
+# would only surface at runtime as degraded sandbox support. Assert the package
+# landed and that every bin it declares exists, which is what would break if
+# upstream started placing those in a postinstall.
+sandbox_runtime_dir="${npm_config_prefix:-$HOME/.local}/lib/node_modules/@anthropic-ai/sandbox-runtime"
+if [ ! -f "$sandbox_runtime_dir/package.json" ]; then
+    echo "sandbox-runtime install failed: $sandbox_runtime_dir/package.json not found" >&2
+    exit 1
+fi
+
+missing_bins=0
+while IFS= read -r bin_path; do
+    [ -n "$bin_path" ] || continue
+    if [ ! -e "$sandbox_runtime_dir/$bin_path" ]; then
+        echo "sandbox-runtime install incomplete: declared bin $bin_path is missing" >&2
+        missing_bins=1
+    fi
+done < <(jq -r '(.bin // empty) | if type == "string" then . else .[] end' "$sandbox_runtime_dir/package.json")
+
+if [ "$missing_bins" -ne 0 ]; then
+    echo "sandbox-runtime may require lifecycle scripts; re-check the --ignore-scripts above" >&2
+    exit 1
+fi
+
+echo "sandbox-runtime installed at: $sandbox_runtime_dir"
--- a/extensions/tools/claude/install.sh
+++ b/extensions/tools/claude/install.sh
@@ -19,5 +19,6 @@ fi
 echo "Claude installed at: $(which claude)"
 claude --version
 
-# Install sandbox runtime for Claude Code sandbox support
-enclave-agent-npm-install @anthropic-ai/sandbox-runtime
+# Install sandbox runtime for Claude Code sandbox support.
+# It ships its vendored artifacts prebuilt and declares no lifecycle scripts.
+enclave-agent-npm-install --ignore-scripts @anthropic-ai/sandbox-runtime
--- a/runtime-assets/build-scripts/bin/enclave-install-npm-tool
+++ b/runtime-assets/build-scripts/bin/enclave-install-npm-tool
@@ -29,8 +29,7 @@ package="$1"
 binary="$2"
 label="${3:-$binary}"
 
-# `--` ends npm option parsing so the package spec can never be read as a flag.
-enclave-agent-npm-install "${npm_flags[@]}" -- "$package"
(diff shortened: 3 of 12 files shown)


runtime-assets/build-scripts/bin/enclave-install-npm-tool
```diff
@@ -8,16 +8,30 @@
...
 
-if [ "$#" -lt 2 ] || [ "$#" -gt 3 ]; then
-    echo "usage: enclave-install-npm-tool <package> <binary> [label]" >&2
+usage() {
+    echo "usage: enclave-install-npm-tool [npm-flag ...] <package> <binary> [label]" >&2
     exit 2
+}
+
+npm_flags=()
+while [ "$#" -gt 0 ] && [[ "$1" == -* ]]; do
```
EclipseSourceAI, line 19: This makes script-blocking opt-in per tool, so the next npm-based tool added silently reintroduces the problem #72 is about. `enclave-install-tool` already exports npm config for exactly this kind of build-wide default ([link](https://github.com/eclipse-enclave/enclave/blob/1fcda88a958ac8dcbb1e223580b8812ef8676d8c/runtime-assets/build-scripts/bin/enclave-install-tool#L49-L61)); exporting `npm_config_ignore_scripts=true` there and letting `opencode/install.sh` opt out with `npm_config_ignore_scripts=false` would be default-safe and needs no new arg parsing in two scripts.

runtime-assets/build-scripts/bin/enclave-install-npm-tool
```diff
@@ -8,16 +8,30 @@
...
+    if [ "$1" = "--" ]; then
+        shift
+        break
+    fi
```
EclipseSourceAI: The `--` branch is dead: no caller uses it and neither README documents it. The parser also silently misreads flags that take a separate value (`--omit dev` would make `dev` the package), so either restrict the contract to `--flag=value` form in the docs or drop the `--` handling.

extensions/tools/claude/install.sh
```diff
@@ -19,5 +19,6 @@ fi
 echo "Claude installed at: $(which claude)"
 claude --version
 
-# Install sandbox runtime for Claude Code sandbox support
-enclave-agent-npm-install @anthropic-ai/sandbox-runtime
+# Install sandbox runtime for Claude Code sandbox support.
+# It ships its vendored artifacts prebuilt and declares no lifecycle scripts.
+enclave-agent-npm-install --ignore-scripts @anthropic-ai/sandbox-runtime
```
EclipseSourceAI, line 24: Nothing verifies this install, unlike codex and pi which get a `command -v` check via `enclave-install-npm-tool` ([link](https://github.com/eclipse-enclave/enclave/blob/1fcda88a958ac8dcbb1e223580b8812ef8676d8c/runtime-assets/build-scripts/bin/enclave-install-npm-tool#L22-L25)). If upstream later adds a postinstall that places something, the build still succeeds and Claude's sandbox support degrades at runtime with no signal. A cheap assertion on the installed bin or package dir would turn that into a build failure.

docs/extensions/adding-a-tool.md
```diff
@@ -37,6 +37,22 @@ sandbox:
...
+helper, which installs through the private agent Node runtime and verifies the
+binary landed on `PATH`:
+
+```sh
+enclave-install-npm-tool [npm-flag ...] <package> <binary> [label]
+```
+
+Leading `-`-prefixed arguments are forwarded to `npm install`. Pass
+`--ignore-scripts` unless the package needs its lifecycle scripts, so
+dependency-authored code does not run at image build time. Check upstream
```
EclipseSourceAI, line 55: Guidance only covers tools. Feature installers still run lifecycle scripts at build time ([node-dev](https://github.com/eclipse-enclave/enclave/blob/1fcda88a958ac8dcbb1e223580b8812ef8676d8c/extensions/features/node-dev/install.sh#L68-L69), [playwright](https://github.com/eclipse-enclave/enclave/blob/1fcda88a958ac8dcbb1e223580b8812ef8676d8c/extensions/features/playwright/install.sh#L19)), and the node-dev set (typescript, eslint, prettier, ...) has no postinstall requirement. Fine to leave out of this PR, but it should be a tracked follow-up rather than an unmentioned gap.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Supply-chain hardening for image builds: `enclave-install-npm-tool` gains leading npm-flag forwarding, and pi, codex, and Claude's `@anthropic-ai/sandbox-runtime` now install with `--ignore-scripts`. opencode is deliberately excluded because its `postinstall` places the platform binary. Docs and the tool READMEs are updated accordingly.

The change is small, correctly scoped to the stated goal, and shellcheck-clean. The per-tool justification comments in each `install.sh` are the right call here, since the reasoning is upstream-specific and not derivable from the code.

Main points for a maintainer:

- The design is opt-in per tool. `enclave-install-tool` already exports `npm_config_*` for build-wide npm behavior, so a default-on `npm_config_ignore_scripts` with an explicit opt-out in opencode would be safer and would avoid the new arg-parsing layer. That is the one design call worth deciding before merge.
- Claude's sandbox-runtime install has no verification step, so a future upstream postinstall requirement would degrade sandbox support silently instead of failing the build.
- Feature installers (`node-dev`, `playwright`) still run lifecycle scripts at build time. Reasonable to defer, but worth a tracked follow-up.

The correctness of the per-package `--ignore-scripts` claims rests on upstream packaging, which I could not verify offline. The author's manual test of pi, codex, and opencode covers the visible failure modes; sandbox-runtime is not covered by that test.

xai:
@EclipseSourceAI

extensions/tools/claude/install.sh
```diff
@@ -19,5 +19,32 @@ fi
...
+fi
+
+missing_bins=0
+while IFS= read -r bin_path; do
+    [ -n "$bin_path" ] || continue
+    if [ ! -e "$sandbox_runtime_dir/$bin_path" ]; then
+        echo "sandbox-runtime install incomplete: declared bin $bin_path is missing" >&2
+        missing_bins=1
+    fi
+done < <(jq -r '(.bin // empty) | if type == "string" then . else .[] end' "$sandbox_runtime_dir/package.json")
```
EclipseSourceAI: This reimplements bin-existence checking that already lives in `enclave-agent-npm-install`'s bin loop, which parses the same package.json `.bin` field for every npm-installed package ([link](https://github.com/eclipse-enclave/enclave/blob/1fcda88a958ac8dcbb1e223580b8812ef8676d8c/runtime-assets/build-scripts/bin/enclave-agent-npm-install#L105-L127)). That loop already resolves `package_dir/bin_path` per declared bin, so asserting the file exists there would cover pi, codex, and opencode too instead of only sandbox-runtime through a second jq expression here.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Follow-up on the five commits pushed since my last review. They address the two concrete bugs I raised: the dead `--` handling in `enclave-install-npm-tool` was removed and later replaced with a real flag/package split that `enclave-agent-npm-install` now enforces for every caller (including claude's direct call), and the sandbox-runtime install now asserts its package landed and its declared bins exist. A further commit also hardens `rewrite_bin_shebang` against a declared bin path escaping the package directory, which is a good catch given `--ignore-scripts` now means package.json content is less trusted.

One new finding: the sandbox-runtime verification in `extensions/tools/claude/install.sh` duplicates bin-parsing logic that already exists in `enclave-agent-npm-install`. Left as an inline comment.

The design-level point about opt-in vs. default-on `--ignore-scripts` and the feature-installer follow-up are both still open and unaddressed; not repeating those here since the existing threads track them.

- [dead `--` branch in enclave-install-npm-tool, removed and replaced with a real flag/package boundary](https://github.com/eclipse-enclave/enclave/pull/73#discussion_r3980895399)
- [claude's sandbox-runtime install now verifies the package and its declared bins landed](https://github.com/eclipse-enclave/enclave/pull/73#discussion_r3980895517)

runtime-assets/build-scripts/bin/enclave-install-npm-tool
xai: Keeping this opt-in per tool. An exported npm_config_ignore_scripts=true wouldn't be the blanket default it looks like: feature installers run their install.sh straight from the Dockerfile, not through enclave-install-tool, so node-dev and playwright would stay uncovered either way. That gap needs its own fix regardless, but I'd prefer to do this in a follow-up, because the fix we have in this PR is worth merging on its own, as the agentic tools have a very high release volume.

docs/extensions/adding-a-tool.md
xai: Agreed, and out of scope here. Filing a follow-up for node-dev and playwright; both need upstream verification per package before flipping the flag, and node-dev's install is warn-only today so a regression there would be silent.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Only one commit landed since my last review (`3f331a4`). It moves the declared-bin check out of `extensions/tools/claude/install.sh` and into `enclave-agent-npm-install`'s existing bin loop, so it now runs for every npm-installed tool (pi, codex, opencode) instead of just sandbox-runtime, and drops the duplicated jq parsing in claude's installer. Checked the loop in `enclave-agent-npm-install`: the new hard-fail on a missing declared bin sits after the existing out-of-package-dir check and before the shebang rewrite, and it lines up with why pi and codex don't need lifecycle scripts (their declared bins ship in the package itself) while opencode stays excluded from `--ignore-scripts` for the same reason as before. No new issues found in the diff.

The two design-level points from earlier (opt-in vs. default-on `--ignore-scripts`, and feature installers like node-dev/playwright still running lifecycle scripts) are still open per the existing threads; the author already responded to both with reasoning to defer them, so leaving those to a maintainer's call rather than repeating.

- [sandbox-runtime bin verification no longer duplicates enclave-agent-npm-install's bin-parsing, moved into the shared loop](https://github.com/eclipse-enclave/enclave/pull/73#discussion_r3981495780)

tortmayr (approved):
Code looks good to me. AI review is also happy now.
I tested it with all mentioned tools. Everything works as expected. 
👍🏼

PR #74, opened 2026-09-10
feat: run sessions on rootless podman

### LINKED ISSUE #68: Rootless support (Podman or rootless Docker)
Enclave currently requires rootfull Docker. As discussed in [#67](https://github.com/eclipse-enclave/enclave/discussions/67#discussioncomment-18272078), if would be safer to support rootless model. 

Following the discussion it would be possible to create a new backend based on **Podman** or **rootless docker**. 

> It should probably try to do it with podman first because it sounds like we need less overhead than with docker.

This issue aim to track the idea. 

(Full discussion at : https://github.com/eclipse-enclave/enclave/discussions/67)

xai:
#### What it does

Closes #68 by adding a rootless podman backend and making the engine choice automatic.

- `--backend podman` drives the existing Docker backend through podman's Docker-compatible CLI. Engine differences are gated on `docker.IsPodman()`: `--userns=keep-id` for session and auth-reconcile containers so the host UID survives rootless podman's user namespace, no `--cache-from` and no home-directory cache mounts on build (buildah commits their parent directories as root-owned), and gateway teardown after the session container is removed. Devcontainer mode stays Docker-only.
- The backend now defaults to `auto`, resolved once at startup: docker when its CLI is on `PATH`, otherwise podman, with the `podman-docker` shim recognized as podman. With both installed and a terminal on all three streams, enclave asks once and saves the answer as `backend` in the global `config.json`. `--json`, `--yes`, non-terminal invocations, and commands that never touch an engine (`tools`, `features`, `config`, `review-target`) never prompt.
- Interrupt handling was reworked so an aborted start cleans up instead of orphaning the gateway: the foreground start runs under a SIGINT/SIGTERM context and exits 130, a stale gateway of the same session is removed before the OAuth port check, and termtint no longer re-raises signals the caller handles. Once attached, SIGINT/SIGTERM sent to the enclave process is forwarded to the engine child so a supervisor can stop a session.
- Rootless podman copies the whole image into a keep-id mapped layer on the first run after every rebuild, blocking other podman commands. The copy is triggered in a throwaway create before the gateway starts and announced when it runs long.

#### How to test

On a host with rootless podman (Fedora is one of the easiest here, it ships the short-name aliases):

```
enclave --backend podman
```

The first start after a build pauses on the keep-id image copy — it should say so rather than look stuck. Ctrl-C during that phase should exit 130 and leave no `enclave-gateway-*` container behind (`podman ps -a`). After attaching, `kill -TERM <enclave pid>` from another shell should stop the session. Then check detection: with both engines installed, a bare `enclave` asks once and writes `"backend"` to `~/.config/enclave/config.json`, while `enclave tools list --json` neither asks nor writes.

#### Follow-ups

- Devcontainer mode under podman is unverified and rejected.
- Cold rebuilds of feature and tool installs are slower under podman because the home-directory build caches are dropped.
- `cleanup --build-cache` has nothing to reclaim under podman.

#### Breaking changes

#### Review checklist


internal/app/backend.go
```diff
@@ -50,11 +58,102 @@ func selectBackend(opts model.Options, dockerOpts backenddocker.Options) (backen
...
+// option resolution, because image builds and engine checks use the wrapper
+// before any backend instance exists. Explicit backends are left untouched.
+// interactive permits the one-time engine question when both are installed.
+func resolveBackend(opts *model.Options, interactive bool) {
+	if strings.TrimSpace(opts.Backend) == backend.NameAuto {
+		opts.Backend = detectBackend(interactive)
+	}
+	switch opts.Backend {
+	case backend.NameDocker, backend.NamePodman:
+		backenddocker.UseCLI(opts.Backend)
```
EclipseSourceAI: `DetectCLIs` maps a `docker` that is really the podman-docker shim onto podman, but an explicit `--backend docker` (or `backend: docker` in config.json) skips that check and drives the shim as Docker: no `--userns=keep-id`, so the bind-mounted stores end up unwritable under rootless podman. Hosts with the shim installed are exactly the ones where users type `--backend docker`, so the explicit name should go through the same shim check.

internal/app/backend.go
```diff
@@ -50,11 +58,102 @@ func selectBackend(opts model.Options, dockerOpts backenddocker.Options) (backen
...
+		return term.IsTerminal(int(os.Stdin.Fd())) && term.IsTerminal(int(os.Stdout.Fd())) && term.IsTerminal(int(os.Stderr.Fd())) // #nosec G115 -- file descriptors fit in int on all supported platforms.
+	}
+	saveBackendChoice = config.WriteGlobalDefault
+)
+
+// backendFreeActions never touch a container engine, so the "auto" backend is
+// left unresolved for them: detection would be wasted work, and a command that
+// lists extensions or prints configuration must not ask which engine to use.
+// Validation of the backend name only runs for the run-like commands.
+var backendFreeActions = map[string]bool{
```
EclipseSourceAI: A few engine-free verbs are missing here: `ssh-init`, `validate-extensions`, `network-print` and `network-diff` never touch an engine ([runNetworkPrint](https://github.com/eclipse-enclave/enclave/blob/1fcda88a958ac8dcbb1e223580b8812ef8676d8c/internal/app/network_cmd.go#L197-L215) only renders policy), so on a dual-engine host `enclave network print` detects and can pop the engine question. Action predicates otherwise live in [actions.go](https://github.com/eclipse-enclave/enclave/blob/1fcda88a958ac8dcbb1e223580b8812ef8676d8c/internal/app/actions.go#L12-L14).

internal/docker/image.go
```diff
@@ -340,8 +344,12 @@ func buildxCommandArgs(req BuildRequest) ([]string, *bytes.Reader) {
 	}
 	args = append(args, sortedMapFlags("--build-arg", req.BuildArgs)...)
 	args = append(args, sortedMapFlags("--label", req.Labels)...)
-	for _, image := range cleanBuildxSpecs(req.CacheFrom) {
-		args = append(args, "--cache-from", image)
+	// podman's --cache-from is a registry-only source that rejects tagged
+	// references; its local layer cache already covers the previous build.
+	if !IsPodman() {
```
EclipseSourceAI, line 349: `BuildxCacheFrom`/`BuildxCacheTo` right below are still forwarded under podman, and `shouldUseBuildx` routes those builds into `podman buildx build --load` with `--cache-to type=local,...`, which podman does not accept. Either gate them the same way or reject `--buildx-cache-*` for `--backend podman`.

internal/docker/image.go
```diff
@@ -376,6 +384,11 @@ func cleanBuildxSpecs(values []string) []string {
 
 // BuildkitEnabled reports whether BuildKit is enabled via DOCKER_BUILDKIT.
 func BuildkitEnabled() bool {
+	// podman builds with buildah, which honors the BuildKit Dockerfile syntax
+	// but takes neither BuildKit progress flags nor inline-cache build args.
+	if IsPodman() {
```
EclipseSourceAI, line 389: Every image build still runs [checkRuntimeImageBuildPreflight](https://github.com/eclipse-enclave/enclave/blob/1fcda88a958ac8dcbb1e223580b8812ef8676d8c/internal/app/build_preflight.go#L59-L65), which hard-fails when `docker buildx` is unavailable and tells the user to install a Docker plugin. buildah handles `RUN --mount` natively, so that gate needs an `IsPodman()` exemption too, otherwise a podman host without a working `podman buildx version` can never build the image.

internal/backend/docker/docker.go
```diff
@@ -49,10 +49,30 @@ func New(opts Options) *Backend {
...
+	dockercmd.SetBinary(name)
+}
+
+// DetectCLIs reports which supported container CLIs the host has on PATH,
+// docker first; the podman-docker shim counts as podman.
+func DetectCLIs() []string {
+	return dockercmd.DetectCLIs()
+}
+
+func (b *Backend) Name() string {
```
EclipseSourceAI, line 56: `Name()` now derives the engine from a package-level global instead of the instance, so a `Backend` built by `New(opts)` has no record of which engine it drives and its identity can change under it. Putting the engine in `Options` would keep it on the instance. `UseCLI`/`DetectCLIs` are also plain forwards to `dockercmd`, which `internal/app` already imports directly ([build.go](https://github.com/eclipse-enclave/enclave/blob/1fcda88a958ac8dcbb1e223580b8812ef8676d8c/internal/app/build.go#L72-L76)).

internal/gateway/gateway.go
```diff
@@ -297,6 +297,11 @@ func gatewayContainerName(containerName string) string {
 	return containerName + model.GatewayContainerSuffix
 }
 
+// ContainerName returns the name of the gateway sidecar for a session container.
+func ContainerName(sessionContainer string) string {
```
EclipseSourceAI: Third spelling of the same name: [gatewayContainerName](https://github.com/eclipse-enclave/enclave/blob/1fcda88a958ac8dcbb1e223580b8812ef8676d8c/internal/gateway/gateway.go#L296-L298) right above and the inline `Ref.Name + model.GatewayContainerSuffix` in the backend ([docker.go](https://github.com/eclipse-enclave/enclave/blob/1fcda88a958ac8dcbb1e223580b8812ef8676d8c/internal/backend/docker/docker.go#L522)). Renaming `gatewayContainerName` to `ContainerName` and using it in both places avoids the wrapper.

docs/security/rootless.md
```diff
@@ -21,7 +21,11 @@ steering through the gateway, not tool execution or persistence.
 
 Use rootful Docker for the supported backend. Docker user namespace remapping
 can reduce host UID exposure while retaining the gateway; see
-[Host hardening](host-hardening.md). The experimental QEMU backend is another
+[Host hardening](host-hardening.md). Rootless podman is different: with
```
EclipseSourceAI: The page still opens with "Status: unsupported" and "Enclave currently requires a rootful Docker daemon", which this paragraph now contradicts. The header needs to be reworded to say rootless podman is supported and rootless Docker is not.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

## What this PR does

Adds a rootless podman backend by driving the existing Docker backend through podman's Docker-compatible CLI: `internal/docker` gets a switchable binary (`SetBinary`/`IsPodman`), and the engine differences are gated on `IsPodman()` (`--userns=keep-id` for session and auth-reconcile containers, no `--cache-from`, no BuildKit flags, home-directory cache mounts stripped from the rendered Dockerfile, gateway teardown after the session container). The backend default becomes `auto`, resolved once in `app.Run` from `DetectCLIs()`, with a one-time prompt persisted to the global `config.json` when both engines are installed. It also reworks interrupt handling: the foreground start runs under a SIGINT/SIGTERM context, exits 130, removes a gateway it created or one orphaned by an earlier interrupted start, and forwards signals to the engine child once attached.

## Where I would focus

- **Engine selection edges.** Shim detection only applies to `auto`; an explicit `--backend docker` on a `podman-docker` host silently loses `keep-id`. The engine-free action allowlist misses a few host-only verbs, which can pop the engine question where nothing needs an engine.
- **Build path adaptation.** `--cache-from` and BuildKit flags are gated, but the buildx cache flags and the `docker buildx` build preflight are not, so `--buildx-cache-*` and the preflight message stay Docker-shaped under podman.
- **Global engine state.** The engine now lives in a package-level variable in `internal/docker`, and `Backend.Name()` reads it instead of the instance, which is worth a deliberate call from a maintainer since it shapes how any future engine is added.
- Interrupt/cleanup handling looked carefully done (post-run auth sync stays detached from the cancelled context, gateway removal ordering matches podman's netns rules, termtint no longer re-raises caller-handled signals). The keep-id warm-up create is a pragmatic answer to the image copy stall.
- `make build` and the touched package tests pass locally. The podman-specific behavior is covered only by CLI stubs, so the real-engine claims in the description (keep-id copy, netns join, `podman version`/`info` output shapes) rest on the author's manual testing.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

All six threads from my earlier review are addressed in the commits since then (682e6a7, 063659c, 2d4c9c1, f3c62f5, 3609d62, 2cb14ee, 3f93ff3), each matching the fix I asked for and backed by new unit tests. I checked each against the current code and they hold up: the shim check now applies to explicit backends too, the engine-free allowlist covers the remaining host-only verbs, buildx cache specs and the buildx preflight are gated on `IsPodman()`, the engine now lives on the `Backend` instance via `Options.Engine`, and the gateway container name helper is deduplicated.

One more fix landed that I hadn't flagged (15a5d03, sharing the user namespace between the podman gateway and session so the session's sysfs mount doesn't fail under runc). I reviewed it against the current code: `joinGatewayNamespaces`/`gatewayUserNS`/`gatewayUser` are consistent, correctly scoped to podman only, and covered by new tests in both `internal/backend/docker` and `internal/gateway`. Nothing to flag there.

`go build ./...` and the touched package tests pass locally. No new inline comments this round.

- [explicit `--backend docker` behind the podman-docker shim now gets keep-id too](https://github.com/eclipse-enclave/enclave/pull/74#discussion_r3981112948)
- [remaining host-only verbs no longer trigger backend detection](https://github.com/eclipse-enclave/enclave/pull/74#discussion_r3981113172)
- [buildx cache specs are dropped with a warning under podman instead of failing the build](https://github.com/eclipse-enclave/enclave/pull/74#discussion_r3981113445)
- [build preflight no longer requires docker buildx under podman](https://github.com/eclipse-enclave/enclave/pull/74#discussion_r3981113758)
- [`Backend.Name()` now reads the engine off the instance instead of a package global](https://github.com/eclipse-enclave/enclave/pull/74#discussion_r3981114105)
- [gateway container name is now derived from one helper everywhere](https://github.com/eclipse-enclave/enclave/pull/74#discussion_r3981114516)
- [rootless.md header now states podman is supported and rootless Docker is not](https://github.com/eclipse-enclave/enclave/pull/74#discussion_r3981114813)

xai:
(I tested with Ubuntu 26.04 and with Fedora 44)

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Three commits landed since my last review, all bug fixes with tests, no scope creep:

- `244773f`: gateway build failures now report the last lines of engine output instead of just the exit status. Reasonable, covered by a new test.
- `2a4da10`: gateway base images are fully qualified (`docker.io/library/...`) and the golang stage is digest-pinned, so podman hosts without unqualified-search-registries can still build the gateway. Docs updated to match.
- `afcaba5`: the container inspect decoder now accepts podman 4.x's single-string `Entrypoint` (podman 5 matches Docker's array shape) via a new `StrSlice` type, and a genuine decode failure is now a hard error instead of silently dropping the container, which previously made a healthy gateway look "exited during startup". Well covered by new tests, and I confirmed the fix against the actual bug path in `waitForGatewayReady`.

`go build ./...` and `go test ./internal/docker/... ./internal/gateway/...` pass locally. No inline comments this round.

All seven threads from my earlier reviews are already marked resolved on the PR, so there's nothing left for me to flag as newly addressed.

tortmayr (approved):
Built-in support for podman is a very nice new feature!
Tested with Ubuntu 24.04. After some initial issues which have all been addressed with  the last commit everything now works as expected.
LGTM! 👍🏼

PR #76, opened 2026-09-10
fix: make qemu guest sessions usable for all bundled tools

### LINKED ISSUE #75: qemu backend: sessions for tools other than claude and pi are unusable
### Bug Description:

`--backend qemu` only works for claude and pi. Three independent faults break the
other tools, and they stack: fixing one exposes the next.

**1. Bundles larger than about 1 GB panic the guest.**

Guest memory is fixed at 4096 MiB (`DefaultMemoryMiB`, raised only by a profile's
`qemuMinMemoryMiB`) and is unrelated to the bundle size. The initramfs is
uncompressed, and the kernel holds the initrd and the ramfs it unpacks into in
memory at the same time, so a bundle costs its own size twice at boot. opencode's
bundle carries a 1.46 GB initramfs and never gets an init:

```
[    2.567863] Initramfs unpacking failed: write error
[    2.746930] Failed to execute /init (error -2)
[    2.747729] Kernel panic - not syncing: No working init found.
```

The same bundle boots at 5120 MiB. 441 MB of it is a baked-in npm cache
(`~/.npm/_cacache`): the Dockerfile keeps that in a BuildKit cache mount, the
microVM builder has no equivalent, and every rootfs byte costs a guest RAM byte
for the whole session. `/boot` adds another 16 MB the guest never reads.

**2. The guest console has no window size.**

`-serial stdio` gives the guest an emulated 16550A, which carries no window size,
and nothing seeds it. In the guest:

```
$ stty size
0 0
$ node -e 'console.log(process.stdout.columns)'
0
```

codex and opencode take that literally and wrap every character onto its own
line. claude and pi fall back to 80x24, so they render but ignore the real
terminal size.

**3. opencode's SQLite store fails on 9p.**

opencode keeps its state in `opencode.db` (SQLite, WAL) inside its config store,
which the backend mounts over virtio-9p. WAL shared-memory files need mmap
semantics there, so the queries fail:

```
$ opencode serve
Error: Unexpected error
Failed query: PRAGMA wal_checkpoint(PASSIVE)
```

The storage layer dies before the app writes a log line, and the TUI paints a
correctly sized frame of spaces, which reads as a blank screen.
`sandbox.qemuStoreCacheMmap` exists for this case; codex declares it and opencode
does not.

Two smaller findings from the same debugging:

- `/etc/hosts` is empty in the guest. Docker bind-mounts it into the provisioning
  container, so `docker export` carries an empty placeholder and `localhost`
  resolves through musl's built-ins, answering `::1` first.
- Resizing the host terminal never reaches the guest. The console is not a
  controlling terminal there, so no `SIGWINCH` is delivered.

### Steps to Reproduce:

1. `enclave --backend qemu --tool opencode` — the guest panics with
   `No working init found` (fault 1).
2. Give the guest more memory, or pick a smaller bundle such as codex, and start
   again. The tool starts but renders one character per line; `stty size` in the
   guest reports `0 0` (fault 2).
3. With the console sized, `enclave --backend qemu --tool opencode` shows a blank
   screen, and `opencode serve` in the same guest fails with the
   `PRAGMA wal_checkpoint` error (fault 3).

### Additional Information

- Operating System: Ubuntu 26.04.1 LTS, kernel 7.2.3, QEMU 10.2.1, x86-64 with
  KVM. Guest kernel is Alpine `linux-virt` 6.18.50.
- Enclave Version: 0.1.0 (0896075)

xai:
#### What it does

Fixes #75.

Three defects kept `--backend qemu` sessions from working with the bundled tools.

- **Bundles too large to boot.** The guest has no disk: the initramfs is unpacked
  into a ramfs that becomes the rootfs, and during unpacking the kernel holds both
  the image and its contents. opencode's 1.4 GB image did not fit the fixed
  4096 MiB guest and panicked without an init. The builder now strips build-time
  caches (npm, pip, uv) and `/boot`, compresses the image with zstd, and records
  packed/unpacked sizes so the host derives guest memory from the actual bundle
  (never below `qemuMinMemoryMiB`). An undersized bundle is rejected on the host
  instead of dying inside the kernel. The compressed image is padded to a 4-byte
  boundary so the kernel still recognizes the per-run overlay archive appended to it.
- **Console reported 0x0.** The emulated serial console carries no window size, so
  codex and opencode rendered one column wide and claude and pi fell back to a fixed
  80x24 box. The generated run script now seeds the console from the host terminal.
- **opencode's SQLite store failed on 9p.** `opencode.db` runs in WAL mode inside the
  config store, and WAL shared-memory files need mmap semantics there; every query
  failed and the TUI painted an empty frame. opencode now declares
  `qemuStoreCacheMmap`, as codex already did. The bundle also gets a populated
  `/etc/hosts` — Docker bind-mounts it into the provisioning container, so the export
  left an empty placeholder and the guest answered `::1` for `localhost` first.

#### How to test

Requires Docker (bundle build) and `qemu-system-x86_64`. Resize your terminal to
something clearly wider than 80 columns first.

1. `enclave --backend qemu --tool opencode` — the bundle builds, the guest boots, and
   the TUI comes up filling the terminal with no `Failed query: PRAGMA wal_checkpoint`
   errors. Send a prompt, exit, restart: the session history is still there.
2. Same for `--tool codex`, `--tool claude`, `--tool pi` — each TUI uses the full
   terminal width, not one column and not 80x24.
3. In a guest shell: `stty size` matches your host terminal, and
   `getent hosts localhost` answers `127.0.0.1`.

#### Follow-ups

The console size is applied once at startup. Resizing the host terminal mid-session
does not reach the guest, because the console is not a controlling terminal there and
no `SIGWINCH` is delivered.

#### Breaking changes

#### Review checklist

### DIFF
--- a/docs/extensions/README.md
+++ b/docs/extensions/README.md
@@ -328,8 +328,10 @@ listed paths are eligible for copy, and a hard-coded deny backstop still
 blocks auth/history/session/runtime-state paths.
 
 Generated QEMU microVM bundles use max(default 4096 MiB,
-`sandbox.qemuMinMemoryMiB`). Use it for tools that need more memory than the
-default to start reliably. Set `sandbox.qemuStoreCacheMmap` when the tool's
+`sandbox.qemuMinMemoryMiB`, what the built initramfs needs). Use it for tools
+that need more memory than the default to start reliably; it never has to
+account for the bundle's own size, because the guest rootfs is the unpacked
+initramfs and the builder already sizes memory to it. Set `sandbox.qemuStoreCacheMmap` when the tool's
 config store needs 9p `cache=mmap` (for example SQLite WAL shared-memory files).
 
 `check-update.sh` runs in a controlled containerized probe environment, never directly on the host. Contract:
--- a/internal/app/build_microvm.go
+++ b/internal/app/build_microvm.go
@@ -8,9 +8,9 @@
 package app
 
 import (
-	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"os/exec"
@@ -67,8 +67,8 @@ func ensureQEMUBundle(input *CommandInput, opts model.Options, buildCfg buildCon
 		logx.Errorf("%v", err)
 		return buildConfig{}, 1
 	}
-	bundleConfig := qemuBundleConfigForProfile(profile)
-	configHash, err := qemuBundleConfigHash(bundleConfig)
+	declaredConfig := qemuBundleConfigForProfile(profile)
+	configHash, err := qemuBundleConfigHash(declaredConfig)
 	if err != nil {
 		logx.Errorf("%v", err)
 		return buildConfig{}, 1
@@ -76,6 +76,7 @@ func ensureQEMUBundle(input *CommandInput, opts model.Options, buildCfg buildCon
 	bundleHash := plan.CombinedHash + "-" + assetHash + "-" + configHash
 	bundleDir := qemuBundleDir(host.Home, opts.Tool, bundleHash)
 	buildCfg.ImageName = bundleDir
+	logx.Debugf("qemu bundle hash: %s", bundleHash)
 	logx.Infof("Using qemu bundle: %s", bundleDir)
 	if opts.NoRebuild {
 		logx.Warnf("Skipping qemu bundle build due to --no-rebuild.")
@@ -85,7 +86,7 @@ func ensureQEMUBundle(input *CommandInput, opts model.Options, buildCfg buildCon
 		}
 		return buildCfg, 0
 	}
-	current, err := qemuBundleCurrent(bundleDir, opts.Tool, bundleHash, bundleConfig)
+	current, err := qemuBundleCurrent(bundleDir, opts.Tool, bundleHash)
 	if err != nil {
 		logx.Debugf("qemu bundle stamp check failed: %v", err)
 	}
@@ -97,7 +98,7 @@ func ensureQEMUBundle(input *CommandInput, opts model.Options, buildCfg buildCon
 		} else {
 			logx.Infof("Agent update interval elapsed, rebuilding qemu bundle automatically.")
 		}
-		if err := buildQEMUBundle(context.Background(), input.Ctx.Paths, host, bundleDir, opts.Tool, bundleHash, bundleConfig); err != nil {
+		if err := buildQEMUBundle(context.Background(), input.Ctx.Paths, host, bundleDir, opts.Tool, bundleHash, declaredConfig.MemoryMiB); err != nil {
 			logx.Errorf("%v", err)
 			return buildConfig{}, 1
 		}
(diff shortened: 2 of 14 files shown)


xai:
@EclipseSourceAI

internal/backend/qemu/bundle.go
```diff
@@ -94,23 +129,113 @@ func resolveBundle(image string) (bundle, error) {
...
+		if errors.Is(err, os.ErrNotExist) {
+			return nil
+		}
+		return err
+	}
+	required := RequiredMemoryMiB(meta.UncompressedBytes, meta.CompressedBytes)
+	if memoryMiB >= required {
+		return nil
+	}
+	return fmt.Errorf("qemu backend: bundle %s needs at least %d MiB to unpack its %d MiB initramfs but %s declares %d MiB; rebuild the bundle with --rebuild",
```
EclipseSourceAI, line 235: The `--rebuild` advice is a dead end for the case that actually trips this. A prebuilt bundle passed via `--image-name` typically has no `enclave-vm-bundle.json`, so `resolveBundleMemoryMiB` falls back to `DefaultMemoryMiB` and fails here, while `--rebuild` is explicitly ignored for `--image-name` ([link](https://github.com/eclipse-enclave/enclave/blob/5c35daf63dda42924fabf2cc7ce539433a14afd4/internal/app/build_microvm.go#L49-L59)). Falling back to `RequiredMemoryMiB` when the config is absent but the metadata is present would make that bundle just work.

internal/app/build_microvm.go
```diff
@@ -259,7 +283,14 @@ func buildQEMUBundle(ctx context.Context, paths model.Paths, host model.Host, bu
 	if err := ensureExistingQEMUBundle(bundleDir); err != nil {
 		return err
 	}
-	if err := writeQEMUBundleConfig(bundleDir, cfg); err != nil {
+	memoryMiB, err := qemuBundleMemoryMiB(bundleDir, minMemoryMiB)
+	if err != nil {
+		return err
+	}
+	if memoryMiB > minMemoryMiB {
+		logx.Infof("Sizing qemu guest memory to %d MiB for the built initramfs.", memoryMiB)
```
EclipseSourceAI, line 291: The derived size is unbounded and never checked against host RAM. With a big enough bundle QEMU gets `-m` larger than the host has, starts fine, and the session dies to the OOM killer mid-run instead of failing up front. The old fixed 4096 MiB could not do that, so a maintainer should decide whether a cap or an upfront warning belongs here.

internal/backend/qemu/console.go
```diff
@@ -0,0 +1,54 @@
...
+		if file, ok := value.(*os.File); ok && file != nil {
+			candidates = append(candidates, file)
+		}
+	}
+	// The session's own stdio first, then the process stdio the backend falls
+	// back to when the caller leaves AttachIO empty.
+	appendFile(attach.Out)
+	appendFile(attach.Err)
+	appendFile(attach.In)
+	candidates = append(candidates, os.Stdout, os.Stderr, os.Stdin)
```
EclipseSourceAI, line 45: This re-implements the nil-stdio fallback that `readerOrDefault`/`writerOrDefault`/`stderrOrDefault` already do ([link](https://github.com/eclipse-enclave/enclave/blob/5c35daf63dda42924fabf2cc7ce539433a14afd4/internal/backend/qemu/qemu.go#L167-L186)). Running `attach` through those instead keeps the probed descriptors identical to the ones QEMU is actually handed.

runtime-assets/microvm/alpine/build-bundle.sh
```diff
@@ -210,9 +232,38 @@ kernel_version=$(find "$root/lib/modules" -mindepth 1 -maxdepth 1 -type d | sed
...
+rm -rf "$root/boot"
+mkdir -m 755 "$root/boot"
+
+# Compress the image: the kernel holds the initrd and the fully unpacked ramfs
+# in memory at the same time, so an uncompressed archive costs its own size
+# twice at boot. zstd is both smaller and faster to unpack than gzip here.
+initramfs=/out/initramfs.cpio.zst
+( cd "$root" && find . -print | cpio -o -H newc --quiet ) | zstd -q -f -T0 -9 -o "$initramfs"
+zstd -q -t "$initramfs"
+uncompressed=$(zstd -q -d -c "$initramfs" | wc -c)
```
EclipseSourceAI, line 247: `sh` has no `pipefail` under `set -eu`, so a zstd failure here is swallowed and `wc` reports a short count that lands in the metadata and under-sizes the guest, which is exactly the boot failure this PR fixes. The `-t` pass above catches a corrupt file but not a failure in this pass; `set -o pipefail` (busybox ash supports it) or an explicit status check would.

runtime-assets/microvm/alpine/build-bundle.sh
```diff
@@ -210,9 +232,38 @@ kernel_version=$(find "$root/lib/modules" -mindepth 1 -maxdepth 1 -type d | sed
...
+if [ "$uncompressed" -le 0 ]; then
+    echo "empty guest initramfs" >&2
+    exit 1
+fi
+
+# unpack_to_rootfs() only takes an appended plain cpio segment - the host-side
+# run overlay - when it starts on a 4-byte boundary, so pad the compressed
+# image out to one. The kernel skips the padding bytes before the segment.
+compressed=$(wc -c < "$initramfs")
+pad=$(( (4 - compressed % 4) % 4 ))
```
EclipseSourceAI, line 257: `concatenateFiles` now aligns every segment before appending, so this padding is redundant for any bundle the host boots. Two copies of the same non-obvious kernel invariant in two languages will drift; worth picking one as authoritative.

runtime-assets/microvm/alpine/README.md
```diff
@@ -5,9 +5,41 @@ This directory contains the experimental builder for the `qemu` backend's minima
...
 - `enclave-vm-bundle.json` with guest sizing metadata
 
+## Initramfs sizing
+
+The guest has no disk: the initramfs is unpacked into a ramfs that becomes the
+rootfs and is never reclaimed, and while unpacking the kernel holds the image
+and its unpacked contents at the same time. Guest memory therefore scales with
+the bundle, and the host derives it from `enclave-vm-initramfs.json`
+(`backendqemu.RequiredMemoryMiB`), never below the profile's
+`qemuMinMemoryMiB`. Two consequences for the builder:
```
EclipseSourceAI, line 19: Guest memory scaling with the bundle is user-visible (opencode now gets 5 GiB instead of 4), but the QEMU section in `docs/cli-reference.md` still says nothing about memory. A sentence there would help, this README is a builder asset doc.

docs/extensions/README.md
```diff
@@ -328,8 +328,10 @@ listed paths are eligible for copy, and a hard-coded deny backstop still
 blocks auth/history/session/runtime-state paths.
 
 Generated QEMU microVM bundles use max(default 4096 MiB,
-`sandbox.qemuMinMemoryMiB`). Use it for tools that need more memory than the
-default to start reliably. Set `sandbox.qemuStoreCacheMmap` when the tool's
+`sandbox.qemuMinMemoryMiB`, what the built initramfs needs). Use it for tools
+that need more memory than the default to start reliably; it never has to
+account for the bundle's own size, because the guest rootfs is the unpacked
+initramfs and the builder already sizes memory to it. Set `sandbox.qemuStoreCacheMmap` when the tool's
```
EclipseSourceAI, line 334: Line is 102 chars while the rest of the paragraph wraps at ~76. Rewrap.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Three independent qemu-guest fixes in one PR: bundle sizing (zstd-compress the initramfs, strip build caches and `/boot`, record packed/unpacked sizes and derive guest memory from them), console sizing (seed `stty rows/cols` from the host terminal in the generated run script), and opencode's SQLite store (`qemuStoreCacheMmap` plus a populated `/etc/hosts` in the bundle).

The approach looks sound. `qemuMinMemoryMiB` correctly becomes a floor that feeds the bundle hash while the effective size is a build output, and dropping the config-content comparison from `qemuBundleCurrent` follows from that. The 4-byte alignment reasoning for the appended overlay matches `unpack_to_rootfs()`, and `concatenateFiles` now enforces it generically, which also covers the legacy uncompressed bundles. `make test` and `shellcheck` pass locally.

- Derived guest memory is unbounded and never compared against host RAM. A large bundle now hands QEMU more than the host has and the session dies to the OOM killer mid-run rather than failing up front.
- The undersized-bundle error advises `--rebuild`, which is exactly the flag that is ignored for the `--image-name` bundles most likely to hit it.
- The builder counts the unpacked size through an unchecked pipeline, so a zstd failure there silently under-reports the size and reintroduces the boot failure this PR fixes.
- The alignment invariant now lives in both the builder and `concatenateFiles`.

Scope is tight, no unrelated refactors, and the new comments explain non-obvious kernel and 9p behavior rather than restating the code.

PR #77, opened 2026-09-12
fix: resolve host.docker.internal to IPv4 for bridge ports

danieldietrich:
#### What it does

Fixes `--bridge-port` (and the automatic IDE bridge) on macOS with Docker Desktop, where forwarding silently failed and host services stayed unreachable from inside the container.

Reported in [discussion #40](https://github.com/eclipse-enclave/enclave/discussions/40#discussion-10589406).

`setup_ide_bridge` in `gateway-entrypoint.sh` resolved `host.docker.internal` with `getent hosts`. On Docker Desktop that returns the IPv6 entry (and may return multiple lines). The address was then passed to `iptables` DNAT/MASQUERADE/ACCEPT rules, which only accept IPv4, so every rule was rejected as a bad address: the DNAT for `localhost:<port>` was never installed (connection refused) and the egress ACCEPT for the host address was missing (`host.docker.internal:<port>` hangs) — which also explains why `--allow-all-network` made it work.

Resolution is now restricted to IPv4 and to a single address:

- `getent ahostsv4` instead of `getent hosts`, taking the first line only.
- The `/etc/hosts` fallback matches only IPv4 addresses.

No flags, options, or output contracts change.

#### How to test

Reproduces the setup from [discussion #40](https://github.com/eclipse-enclave/enclave/discussions/40#discussion-10589406) on macOS with Docker Desktop:

1. Serve something on the host on `0.0.0.0:8000` (the report used oMLX; `python3 -m http.server 8000 --bind 0.0.0.0` works as well).
2. Confirm the host side: `curl http://localhost:8000/health`.
3. Confirm plain Docker reaches it: `docker run --rm curlimages/curl http://host.docker.internal:8000/health`.
4. Run `enclave shell --bridge-port 8000` and inside the container:
   - `curl http://localhost:8000/health` — on `main` this fails immediately with connection refused; with the fix it answers.
   - `curl http://host.docker.internal:8000/health` — on `main` this hangs; with the fix it answers.
5. Neither call should need `--allow-all-network` anymore.
6. Gateway log check: `docker logs <container>-gateway | grep "IDE bridge"` should show `IDE bridge: port 8000 -> <IPv4>:8000` instead of `IDE bridge: failed DNAT for port 8000`.

Same for the automatic IDE bridge: with VS Code attached (a `~/.claude/ide/<port>.lock` file present), the discovered port should now be forwarded successfully.

Linux is unaffected — `getent ahostsv4` returns the same IPv4 address that `getent hosts` returned there.

#### Follow-ups

- The gateway entrypoint has no automated test coverage; this path is verified manually only. A shell-level test for the host resolution in `gateway-entrypoint.sh` would be worth adding.
- Host resolution is Docker-specific (`host.docker.internal`). Other backends (microVMs/QEMU) will need a transparent way to resolve the host address, as noted in the discussion.
- If IPv6-only hosts ever need support, the rules would need an `ip6tables` counterpart. Not required for any currently supported setup.

#### Breaking changes

#### Review checklist

### DIFF
--- a/gateway-entrypoint.sh
+++ b/gateway-entrypoint.sh
@@ -326,7 +326,12 @@ setup_ide_bridge() {
         return
     fi
 
-    ide_host_ip=$(awk '$1 ~ /^[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+$/ && /host\.docker\.internal/ {print $1; exit}' /etc/hosts)
+    ide_host_ip=$(awk '{ sub(/#.*/, "") }
+        $1 ~ /^[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+$/ {
+            for (i = 2; i <= NF; i++) {
+                if ($i == "host.docker.internal") { print $1; exit }
+            }
+        }' /etc/hosts)
     if [ -z "$ide_host_ip" ]; then
         ide_host_ip=$(getent ahostsv4 host.docker.internal 2>/dev/null | awk '{print $1; exit}')
     fi
--- a/gateway-entrypoint.sh
+++ b/gateway-entrypoint.sh
@@ -326,15 +326,12 @@ setup_ide_bridge() {
         return
     fi
 
-    ide_host_ip=$(awk '{ sub(/#.*/, "") }
-        $1 ~ /^[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+$/ {
-            for (i = 2; i <= NF; i++) {
-                if ($i == "host.docker.internal") { print $1; exit }
-            }
-        }' /etc/hosts)
-    if [ -z "$ide_host_ip" ]; then
-        ide_host_ip=$(getent ahostsv4 host.docker.internal 2>/dev/null | awk '{print $1; exit}')
+    if ! command -v enclave_resolve_docker_host_ipv4 >/dev/null 2>&1; then
+        log "IDE bridge: network helper unavailable; disabled"
+        return
     fi
+
+    ide_host_ip=$(enclave_resolve_docker_host_ipv4 || true)
     if [ -z "$ide_host_ip" ]; then
         log "IDE bridge: cannot resolve host.docker.internal; disabled"
         return
--- a/gateway-entrypoint.sh
+++ b/gateway-entrypoint.sh
@@ -326,10 +326,6 @@ setup_ide_bridge() {
         return
     fi
 
-    # Only IPv4: these rules go into iptables (not ip6tables), and on Docker
-    # Desktop `getent hosts` returns the IPv6 entry for host.docker.internal,
-    # which iptables rejects as a bad address. Take the first address only --
-    # getent prints one line per address.
     ide_host_ip=$(getent ahostsv4 host.docker.internal 2>/dev/null | awk '{print $1; exit}')
     if [ -z "$ide_host_ip" ]; then
         ide_host_ip=$(awk '$1 ~ /^[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+$/ && /host\.docker\.internal/ {print $1; exit}' /etc/hosts)
--- a/gateway-entrypoint.sh
+++ b/gateway-entrypoint.sh
@@ -326,9 +326,13 @@ setup_ide_bridge() {
         return
     fi
 
-    ide_host_ip=$(getent hosts host.docker.internal 2>/dev/null | awk '{print $1}')
+    # Only IPv4: these rules go into iptables (not ip6tables), and on Docker
+    # Desktop `getent hosts` returns the IPv6 entry for host.docker.internal,
+    # which iptables rejects as a bad address. Take the first address only --
+    # getent prints one line per address.
+    ide_host_ip=$(getent ahostsv4 host.docker.internal 2>/dev/null | awk '{print $1; exit}')
     if [ -z "$ide_host_ip" ]; then
-        ide_host_ip=$(awk '/host\.docker\.internal/ {print $1; exit}' /etc/hosts)
+        ide_host_ip=$(awk '$1 ~ /^[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+$/ && /host\.docker\.internal/ {print $1; exit}' /etc/hosts)
     fi
     if [ -z "$ide_host_ip" ]; then
         log "IDE bridge: cannot resolve host.docker.internal; disabled"
--- a/gateway-entrypoint.sh
+++ b/gateway-entrypoint.sh
@@ -326,9 +326,9 @@ setup_ide_bridge() {
         return
     fi
 
-    ide_host_ip=$(getent ahostsv4 host.docker.internal 2>/dev/null | awk '{print $1; exit}')
+    ide_host_ip=$(awk '$1 ~ /^[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+$/ && /host\.docker\.internal/ {print $1; exit}' /etc/hosts)
     if [ -z "$ide_host_ip" ]; then
-        ide_host_ip=$(awk '$1 ~ /^[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+$/ && /host\.docker\.internal/ {print $1; exit}' /etc/hosts)
+        ide_host_ip=$(getent ahostsv4 host.docker.internal 2>/dev/null | awk '{print $1; exit}')
     fi
     if [ -z "$ide_host_ip" ]; then
         log "IDE bridge: cannot resolve host.docker.internal; disabled"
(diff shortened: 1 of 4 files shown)


tortmayr:
@EclipseSourceAI

gateway-entrypoint.sh
```diff
@@ -326,9 +326,13 @@ setup_ide_bridge() {
...
-    ide_host_ip=$(getent hosts host.docker.internal 2>/dev/null | awk '{print $1}')
+    # Only IPv4: these rules go into iptables (not ip6tables), and on Docker
+    # Desktop `getent hosts` returns the IPv6 entry for host.docker.internal,
+    # which iptables rejects as a bad address. Take the first address only --
+    # getent prints one line per address.
+    ide_host_ip=$(getent ahostsv4 host.docker.internal 2>/dev/null | awk '{print $1; exit}')
     if [ -z "$ide_host_ip" ]; then
-        ide_host_ip=$(awk '/host\.docker\.internal/ {print $1; exit}' /etc/hosts)
+        ide_host_ip=$(awk '$1 ~ /^[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+$/ && /host\.docker\.internal/ {print $1; exit}' /etc/hosts)
     fi
```
EclipseSourceAI, line 332: `getent` here can now block on DNS where the old call did not. `main` rewrites `/etc/resolv.conf` to `nameserver 127.0.0.1` before `setup_firewall` and only starts dnsmasq afterwards ([link](https://github.com/eclipse-enclave/enclave/blob/d547563df7eed5b9f3658ff786ea50a38390c0d0/gateway-entrypoint.sh#L695-L728)), so any hosts-file miss (IPv6-only entry, exactly the Docker Desktop case this fixes) turns into a query against a dead resolver and burns the musl resolver timeout inside the 15s [`gatewayReadyTimeout`](https://github.com/eclipse-enclave/enclave/blob/d547563df7eed5b9f3658ff786ea50a38390c0d0/internal/gateway/gateway.go#L37). Swapping the order (parse `/etc/hosts` first, `getent` only as fallback) avoids DNS entirely, since Docker always writes the `--add-host host.docker.internal:host-gateway` entry into `/etc/hosts`.

gateway-entrypoint.sh
```diff
@@ -326,9 +326,13 @@ setup_ide_bridge() {
...
     if [ -z "$ide_host_ip" ]; then
         log "IDE bridge: cannot resolve host.docker.internal; disabled"
```
EclipseSourceAI, line 336: The `/etc/hosts` regex matches the whole line, so `10.1.2.3 build.host.docker.internal.example` wins over the real entry. `net.sh` already matches the hostname field exactly ([link](https://github.com/eclipse-enclave/enclave/blob/d547563df7eed5b9f3658ff786ea50a38390c0d0/runtime-assets/net.sh#L37)), same thing works here: `$1 ~ /^[0-9.]+$/ && $2 == "host.docker.internal"`.

gateway-entrypoint.sh
EclipseSourceAI, line 332: Both entrypoints resolve host addresses and `runtime-assets/net.sh` is the shared helper file for that, already parsing `/etc/hosts` with a `host.docker.internal` special case ([link](https://github.com/eclipse-enclave/enclave/blob/d547563df7eed5b9f3658ff786ea50a38390c0d0/runtime-assets/net.sh#L28-L43)). Putting this in `enclave_resolve_docker_host_ipv4` there also makes it testable: `net.sh` is sourceable, `gateway-entrypoint.sh` runs `main` unconditionally, which is why the follow-up about missing test coverage is currently blocked.

gateway-entrypoint.sh
```diff
@@ -326,9 +326,13 @@ setup_ide_bridge() {
...
 
-    ide_host_ip=$(getent hosts host.docker.internal 2>/dev/null | awk '{print $1}')
+    # Only IPv4: these rules go into iptables (not ip6tables), and on Docker
+    # Desktop `getent hosts` returns the IPv6 entry for host.docker.internal,
+    # which iptables rejects as a bad address. Take the first address only --
+    # getent prints one line per address.
+    ide_host_ip=$(getent ahostsv4 host.docker.internal 2>/dev/null | awk '{print $1; exit}')
     if [ -z "$ide_host_ip" ]; then
-        ide_host_ip=$(awk '/host\.docker\.internal/ {print $1; exit}' /etc/hosts)
+        ide_host_ip=$(awk '$1 ~ /^[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+$/ && /host\.docker\.internal/ {print $1; exit}' /etc/hosts)
```
EclipseSourceAI: Four lines of comment for two lines of code, and the `--` is a dash used as sentence structure. Something like `# iptables needs IPv4; getent hosts returns the IPv6 entry on Docker Desktop.` carries the same information.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Two-line fix in `setup_ide_bridge` (`gateway-entrypoint.sh`): resolve `host.docker.internal` with `getent ahostsv4` and take only the first line, plus an IPv4-only filter on the `/etc/hosts` fallback. The diagnosis matches the code: the result is fed straight into `iptables` DNAT/MASQUERADE/ACCEPT rules, so an IPv6 address or a multi-address string makes every rule fail and the bridge silently dies. Scope is tight, no flags or output contracts change, and shellcheck is clean.

- Ordering. The lookup runs during `setup_firewall`, after `/etc/resolv.conf` has been pointed at `127.0.0.1` and before dnsmasq starts, so a hosts-file miss now becomes a DNS query against a dead resolver. Doing the `/etc/hosts` parse first would remove that dependency.
- The `/etc/hosts` fallback matches the whole line rather than the hostname field.
- Placement: `runtime-assets/net.sh` is the shared helper file both entrypoints source, and the current shape of `gateway-entrypoint.sh` is exactly what blocks the "no automated test coverage" follow-up the PR notes.

One thing I could not verify: `entrypoint.sh` has a parallel IDE bridge path using `socat ... TCP:host.docker.internal:${port}` for the non-gateway network mode, which carries the same address-family assumption. The discussion report suggests that path already works on Docker Desktop, so this may be a non-issue, but pinning it to `TCP4:` would make the two paths consistent.

gateway-entrypoint.sh
danieldietrich: Addressed in 3a3cb78382a1d7888bad240207f9454db511aee9

gateway-entrypoint.sh
danieldietrich: Addressed in 705d0121db203289e709f28f75406c36738f397a

gateway-entrypoint.sh
danieldietrich: Addressed in cc7a8d9dc748806f317e1ca7b3685e8076d32cef

gateway-entrypoint.sh
danieldietrich: Addressed in f9097e27dc66e238de922e5c9315593f6c47a355

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Follow-up review. Since my last pass, four commits landed: removing the oversized comment, reordering `/etc/hosts` parsing before the `getent` fallback, matching the hostname field instead of the whole line (with `#` comment stripping, which is a nice extra), and moving the whole thing into `runtime-assets/net.sh` as `enclave_resolve_docker_host_ipv4`, guarded the same way the other net.sh helpers already are in both entrypoints.

The refactor also added `internal/gateway/net_asset_test.go`, which sources the real `net.sh` and exercises the hosts-file parsing and the `getent` fallback with a stubbed binary. I ran it, it passes, and it closes the "no test coverage" follow-up called out in the PR description. Ran `gofmt`, `shellcheck`, and `go build ./...` on the touched files, all clean.

No new findings on the commits pushed since my last review.

- [DNS lookup ordering: parse /etc/hosts before falling back to getent, avoiding a query against the dead resolver](https://github.com/eclipse-enclave/enclave/pull/77#discussion_r3994562579)
- [match host.docker.internal by hostname field instead of matching the whole /etc/hosts line](https://github.com/eclipse-enclave/enclave/pull/77#discussion_r3994562622)
- [move the resolution logic into the shared runtime-assets/net.sh helper so it's testable](https://github.com/eclipse-enclave/enclave/pull/77#discussion_r3994562651)
- [trim the four-line comment down, it was disproportionate to the two lines of code](https://github.com/eclipse-enclave/enclave/pull/77#discussion_r3994563445)

tortmayr (approved):
@danieldietrich Thanks for the PR. Changes look good to me.
Unfortunately I don't have a MacOS system to test this with. However, the changes are sound and should fix the issue.
Can confirm that Linux is unaffected.

danieldietrich:
Thank you so much & looking forward to the next PR...
(e.g. `enclave tools add <source>` might not work for local resources on main).

kalyi:
Hi,
I tested it with macOS Tahoe and it works as expected (except that `/health` doesn't necessarily exist with an arbitrary service). Thanks for the PR!

xai (approved):

PR #79, opened 2026-09-12
feat: enable per-project Codex memories

### LINKED ISSUE #78: Support Codex memories and generalize the agent-memory system
Every enclave session starts from a fresh container, so Codex relearns the project on each run. Codex ships a native memories feature that carries project context across sessions. Enable it in the Codex settings template.

The existing memory system was built for Claude, where memory is self-contained markdown in the mounted directory. Codex breaks three of its assumptions, so enabling it has implications beyond the template:

1. Memory is split between the mount and the config store. Codex keeps memory files in ~/.codex/memories/ but the SQLite index and writer locks at the config-store root. Since enclave can create several config stores per project (concurrent sessions, worktrees), a shared memory directory would desync from the per-store indexes. Memory must be keyed by config-store key (sandbox.memoryScope: session), and cleanup must treat a session-scoped memory directory and its config store as one unit: --keep memory retains the store, --keep history retains memory, and the ephemeral sweep removes memory/<key> together with config-store/<key>.
2. Config overlays wipe store-resident state. With a config patch or host override active, the config-source overlay rebuilds the store on every run and would delete the memory index. Tools need a way to pin config-relative runtime state across overlays and out of host passthrough (sandbox.statePaths).
3. --no-memory only omits the mount. With memories = true baked into the template, Codex would keep generating memories into the persistent store on --no-memory and --ephemeral runs. Tools need native disable arguments (sandbox.noMemoryArgs). Claude has the same gap today: autoMemoryEnabled: true stays active without the mount and its writes land in the config store, so Claude should declare disable arguments as well.

Two further consequences: the new spec fields grant persistent host writes and argv injection, so tools add/update must show them in the capability summary and update diff; and background memory consolidation uses plan quota, so opting out must work per run (--no-memory) and permanently (config patch).

Memory stays per project and per tool under ~/.local/state/enclave/projects/<hash>/<tool>/memory/, never shared between projects or agents.

tortmayr:
#### What it does

Turns on Codex memories (`[features] memories = true` in the settings template) so sandboxed sessions stop relearning the project from scratch. Background consolidation uses plan quota; opt out per run with `--no-memory` or permanently with a config patch.

The existing memory system assumed Claude's model (self-contained files in one mounted directory). Codex splits memory between the mount and its config store, which drives the new spec-schema controls (available to all tool extensions, including installed third-party ones):

- `sandbox.memoryScope: session` keys the memory mount by config-store key, so concurrent stores never share a memory database
- `sandbox.noMemoryArgs` natively disables memory use and generation for `--no-memory` and `--ephemeral` runs. Claude declares it too (`--settings '{"autoMemoryEnabled": false}'`), so those flags now stop generation instead of only omitting the mount
- `sandbox.statePaths` pins config-relative runtime state across config-source overlays and keeps it out of host config passthrough. Codex pins its memory files, their index database, and its thread databases; the thread databases are rebuildable rollout projections, pinned to avoid re-deriving them on every overlaid run
- `cleanup` treats session-scoped memory and its config store as one unit (`--keep memory` / `--keep history`) and tolerates tools whose extension has been removed
- `tools add`/`update` report the new grants (agent memory, no-memory args, preserved state paths) in the capability summary and update diff

Memory scope is only resolved for tools that declare a memory dir, so memory-less tools don't grow a meaningless field in their golden surface.

Fixes #78

#### How to test

- [x] Verified locally: build, tests, lint, and license-header checks pass

All probes run inside `enclave --tool codex shell` sessions; no host paths needed. Planted markers stand in for real memory generation, which the model triggers at its own discretion.

1. **Persistence:** in a shell session, `echo probe > ~/.codex/memories/PROBE.md`, exit. Start a new shell session: the probe is still there.
2. **Mount-backed, not store-backed:** restart with `--no-memory`: the probe is gone, because it lives in the omitted memory mount rather than the config store. A regular restart shows it again; remove it afterwards.
3. **Overlay survival:** with a config patch active (`~/.config/enclave/patches/codex/config.toml`), in a shell session `touch ~/.codex/state_999.sqlite ~/.codex/memories_999.sqlite`, exit, restart: both markers survive the store overlay. On main the overlay deletes them.
4. **Ephemeral:** in an `--ephemeral` shell session write a probe into `~/.codex/memories/`, exit, start a regular session: the probe is gone, discarded with the session store.
5. **Cleanup coupling:** `enclave cleanup --tool codex --dry-run --keep memory` retains the config store together with memory; `--keep history` retains both as well; a plain `cleanup --dry-run` lists both for removal.
6. **Disable argv:** a `--no-memory` Codex session reports memories disabled via `/memories`. The injected argv (`-c features.memories=false`, Claude's `--settings '{"autoMemoryEnabled": false}'`) is pinned by unit tests.

The same seams are covered by automated tests (`memory_mount_test.go`, `overlay_test.go`, `cleanup_test.go`, `host_config_passthrough_test.go`); the probes are the end-to-end variant of those assertions.

#### Follow-ups

- The overlay stages preserved paths by copying through `$TMPDIR` and walks into directories it already staged wholesale. With `statePaths` now pinning large trees (`memories/` can hold a git repository), staging via rename on the same filesystem and skipping staged directories would cut startup cost on overlaid runs. Worth a ticket; not done here to keep `overlay.go` out of this PR.

#### Breaking changes

#### Review checklist


internal/config/profile.go
```diff
@@ -73,18 +73,43 @@ func validateAndNormalizeProfile(profile *model.Profile) error {
...
+	case model.MemoryScopeSession:
+		if profile.MemoryDir == "" || profile.ConfigDir == "" {
+			return fmt.Errorf("memory_scope session requires memory_dir and config_dir")
+		}
+	default:
+		return fmt.Errorf("memory_scope must be project or session")
+	}
+	if profile.MemoryDir != "" {
+		profile.MemoryScope = model.ResolveMemoryScope(profile.MemoryScope)
+	}
```
EclipseSourceAI, line 113: A spec with `memoryScope: project` but no `memoryDir` passes validation and keeps the scope on the profile, so it still shows up in the serialized surface while doing nothing (`describeMemory` hides it too). Rejecting it would match the `session` branch above and how the other dependent fields are handled ([skills_dir requires config_dir](https://github.com/eclipse-enclave/enclave/blob/5c35daf63dda42924fabf2cc7ce539433a14afd4/internal/config/profile.go#L117-L124)).

internal/app/cleanup.go
```diff
@@ -57,7 +131,11 @@ func runCleanup(run model.RunOptions, cleanup model.CleanupOptions) int {
 		}
 		// Ephemeral config stores are host directories keyed by a session or
 		// worktree suffix.
-		storeDirs := resolveEphemeralStoreDirs(run, cleanup, home, project)
+		storeDirs, err := resolveEphemeralStoreDirs(run, cleanup, home, project, scopes)
+		if err != nil {
+			logx.Errorf("Failed to resolve memory cleanup policy: %v", err)
+			return 1
```
EclipseSourceAI: For `--ephemeral --all` one unloadable third-party spec aborts the sweep for every project and tool. Under the fallback scope nothing gets over-deleted (worst case some session memory dirs are left behind), so warning and continuing per tool would fit the "cleanup is what a user reaches for when the extension tree is broken" rationale above better than a hard abort.

extensions/tools/codex/spec.yaml
```diff
@@ -17,6 +17,18 @@ sandbox:
...
   qemuStoreCacheMmap: true
   skillsDir: .codex/skills
+  memoryDir: .codex/memories
+  memoryScope: session
+  noMemoryArgs: [-c, features.memories=false]
+  statePaths:
+    # Memory files, their SQLite index, and codex-managed writer locks.
+    - memories/
+    - memories_*.sqlite*
+    - thread-writer-locks/
```
EclipseSourceAI, line 29: Pinning the lock directory means a lock left behind by a container killed mid-write survives the overlay instead of being cleared with the rest of the store. Is that safe for codex's writer locks, or should this one stay out so an overlaid run resets them?

extensions/tools/codex/spec.yaml
```diff
@@ -17,6 +17,18 @@ sandbox:
   qemuMinMemoryMiB: 4096
   qemuStoreCacheMmap: true
   skillsDir: .codex/skills
+  memoryDir: .codex/memories
+  memoryScope: session
+  noMemoryArgs: [-c, features.memories=false]
+  statePaths:
```
EclipseSourceAI, line 23: `memoryDir` is home-relative, `statePaths` is config-relative, so the same directory reads as `.codex/memories` on line 20 and `memories/` here. It's documented, but a short note on the base would save third-party spec authors the trip to the docs.

extensions/tools/codex/templates/config.toml
```diff
@@ -31,6 +31,12 @@ metrics_exporter = "none"
 log_user_prompt = false
 
 [features]
+# Opted in, unlike fast_mode below. A container starts with no accumulated
+# context, so without memories every session relearns the project from scratch,
+# and no other setting recovers that. Background consolidation does use quota.
+# Opt out per run with enclave --no-memory or permanently with a config patch;
+# see this tool's README.
+memories = true
```
EclipseSourceAI, line 39: This turns on background consolidation and its quota use for every existing codex user on upgrade. The reasoning is sound and there are opt-outs, but a maintainer should sign off on the default-on and make sure it lands in the release notes.

extensions/tools/codex/README.md
```diff
@@ -56,6 +57,36 @@ To show branch and usage limits in Codex's TUI status line, add a TOML patch:
...
+Enclave's default template enables [Codex memories](https://developers.openai.com/codex/memories)
+with `features.memories = true`. This lets Codex use eligible conversation content
+for memory generation and consumes additional model quota. A full override such
+as `~/.config/enclave/tools/codex/config.toml` replaces the template, so those
+users get the feature setting from their own config (Codex defaults to off).
+
+Enclave isolates memory by config-store key. See [Agent Memory](../../../docs/runtime/stores.md#agent-memory)
+for session reuse, `--no-memory`, ephemeral runs, and cleanup semantics.
+
+To disable memories by default, add a Codex config patch:
```
EclipseSourceAI: Missing from the opt-out list: enclave's own [`no_memory`](https://github.com/eclipse-enclave/enclave/blob/5c35daf63dda42924fabf2cc7ce539433a14afd4/docs/configuration.md#L66) config option, which is the first thing most users would reach for and also drops the mount.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Enables Codex memories (`features.memories = true` in the settings template) and generalizes enclave's memory model to fit a tool that splits memory between the bind mount and its config store. Three new spec-schema fields carry that: `sandbox.memoryScope: session` keys the memory mount by config-store key (reusing the existing `sessionGeneratedKey`, same as the skills mount), `sandbox.noMemoryArgs` gives `--no-memory`/`--ephemeral` a native disable argv (Claude declares one too, so those flags now stop generation instead of only dropping the mount), and `sandbox.statePaths` pins config-relative runtime state, routed into the existing `HostConfigPassthroughDeniedPaths` so both the overlay preserve list and passthrough denial pick it up for free. Cleanup learns to treat session memory and its config store as one unit and to tolerate tools whose extension is gone.

The integration looks clean: existing seams get reused rather than duplicated, `compactProfileArgs` moves to load-time normalization where it also covers the new argv, and the docs (`stores.md`, extension docs, cli-reference) are updated properly. Tests are thorough, including the overlay running against the real preserve-path policy. Build, vet, and the touched package tests pass locally.

- Turning `memories = true` on by default consumes plan quota for every existing Codex user on upgrade. A product call, not a code issue.
- Cleanup's abort-vs-warn split on an unloadable spec is fine for the single-tool path but hard-fails the whole `--ephemeral --all` sweep.
- `statePaths` pins `thread-writer-locks/`, which is the one entry where preserving state across an overlay could carry a stale lock forward.
- The permanent-opt-out documentation omits enclave's own `no_memory` config option.

Remaining inline comments are small validation and documentation nits.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Follow-up review. The new commit (a76a6d7) addresses all four inline points from my earlier review: `memory_scope: project` without `memory_dir` is now rejected, the ephemeral sweep no longer aborts entirely on one unloadable spec (falls back per tool, skipping only under `--keep memory`), the codex spec comment now explains the config-relative base and why pinning the lock dir is safe, and the README lists `no_memory` as an opt-out. Tests were updated to match (`cleanup_test.go`, `spec_loader_test.go`) and pass locally, along with `go build ./...`.

The default-on `features.memories = true` in `config.toml` (quota impact for existing users) is unchanged, but that was flagged as a product/maintainer sign-off question, not a code defect, so nothing to add there.

No new findings in the delta since my last review.

- [reject `memory_scope: project` declared without `memory_dir`](https://github.com/eclipse-enclave/enclave/pull/79#discussion_r3996144893)
- [stop one broken spec from aborting the whole `--ephemeral --all` sweep](https://github.com/eclipse-enclave/enclave/pull/79#discussion_r3996144947)
- [explain why pinning codex's writer-lock dir across an overlay is safe](https://github.com/eclipse-enclave/enclave/pull/79#discussion_r3996144988)
- [clarify statePaths are config-relative unlike the home-relative memoryDir](https://github.com/eclipse-enclave/enclave/pull/79#discussion_r3996145010)
- [list enclave's own `no_memory` option as a memory opt-out](https://github.com/eclipse-enclave/enclave/pull/79#discussion_r3996145065)

xai (approved):
Code looks good and the feature works fine!

PR #81, opened 2026-09-12
fix: keep gateway leaf certificates container-private

### LINKED ISSUE #80: Gateway MITM fails when docker and rootless podman share a host
### Bug Description:

The gateway bind-mounts `~/.local/state/enclave/tls/hosts` into the container as a
cache for the per-host leaf certificates it mints for TLS interception, and
`gateway-entrypoint.sh` chowns that host directory to the in-container proxy user
(`gateway`).

That UID means different things to the two engines. Under docker the container UID
lands on the host as the same numeric UID (101 for `gateway`). Under rootless podman
it is mapped through the user's subuid range, so the same chown writes an unrelated
high UID onto the host directory.

On a host that runs sessions under both engines, whichever one starts second finds the
directory owned by a UID its proxy user does not have, cannot write leaf certificates,
and the MITM handshake fails. The failure surfaces far from its cause: secret release
stops working, and `gh` and other tools that receive placeholder credentials report an
`invalid token` error instead of a TLS problem.

Nothing under the host TLS root needs to be writable by the container in the first
place. The CA has to be shared so the agent container trusts it; the leaves are
derived data that each gateway can mint for itself.

### Steps to Reproduce:

1. On one host, start an Enclave session with the docker backend and let the gateway
   intercept at least one HTTPS host, so `~/.local/state/enclave/tls/hosts` is created
   and chowned to UID 101.
2. Start a session with the rootless podman backend on the same host, with a restricted
   network profile and a secret-backed tool such as `gh`.
3. Run a `gh` command inside the session that needs the released secret.

Observed: the request fails and `gh` reports `invalid token`. The gateway proxy log
shows the leaf certificate write to `/tls/hosts` failing with a permission error.

Expected: both backends work on the same host, independent of which one ran first.

The reverse order reproduces it as well: podman first leaves the directory owned by a
subuid, and the docker gateway then cannot write to it.

xai:
#### What it does

Fixes #80: MITM interception breaks on hosts that run sessions under both docker and
rootless podman.

The gateway bind-mounted `~/.local/state/enclave/tls/hosts` as a leaf certificate cache
and `gateway-entrypoint.sh` chowned it to the in-container proxy user. That UID lands on
the host as UID 101 under docker, but as a mapped subuid under rootless podman, so
whichever engine started second could not write leaf certificates. The handshake failed,
secret release stopped working, and `gh` and similar tools reported a misleading
`invalid token`.

Only the CA is shared with the host now. Each gateway mints its leaves into
`/tls/hosts` inside the container, owned by the proxy user and created fresh per
container, so nothing under the host TLS root is ever chowned to a container UID. Leaves
are derived data, so dropping the cross-session cache costs one extra certificate mint
per host per session and nothing else.

Removed along with the mount: `model.GatewayTLSHostsPath`, `config.HostTLSHostsDir`, and
the host-side `MkdirAll` calls in `internal/gateway/gateway.go` and
`internal/backend/docker/gateway.go`. `docs/ARCHITECTURE.md` and
`docs/runtime/network-request-flow.md` are updated to match.

Existing `~/.local/state/enclave/tls/hosts` directories are left behind and no longer
used; they can be deleted.

#### How to test

1. On a host with both engines available, start a session on the docker backend with a
   restricted network profile, and make an HTTPS request so the gateway mints a leaf.
2. Start a session on the rootless podman backend on the same host and run a
   secret-backed `gh` command. On `main` this fails with `invalid token`; here it
   succeeds.
3. Repeat in the reverse order (podman first, then docker).
4. `make build`, `make test`, `make lint`.

#### Follow-ups

None.

#### Breaking changes

#### Review checklist

### DIFF
--- a/gateway-entrypoint.sh
+++ b/gateway-entrypoint.sh
@@ -483,12 +483,10 @@ prepare_proxy_runtime_data() {
         mkdir -p "$proxy_tls"
         [ -f "${tls_root}/ca.crt" ] && cp "${tls_root}/ca.crt" "$proxy_tls/"
         [ -f "${tls_root}/ca.key" ] && cp "${tls_root}/ca.key" "$proxy_tls/"
-        if [ -d "${tls_root}/hosts" ]; then
-            chown "$PROXY_USER:$PROXY_USER" "${tls_root}/hosts" 2>/dev/null || true
-            ln -sfn "${tls_root}/hosts" "$proxy_tls/hosts"
-        else
-            mkdir -p "$proxy_tls/hosts"
-        fi
+        # Leaf certificates stay container-private; nothing under the host
+        # TLS root is ever chowned to a container UID.
+        mkdir -p "$proxy_tls/hosts"
+        chmod 700 "$proxy_tls/hosts"
         chown "$PROXY_USER:$PROXY_USER" "$proxy_data" "$proxy_tls" 2>/dev/null || true
         chown "$PROXY_USER:$PROXY_USER" "$proxy_tls/ca.crt" "$proxy_tls/ca.key" "$proxy_tls/hosts" 2>/dev/null || true
         chmod 600 "$proxy_tls/ca.key" 2>/dev/null || true
--- a/internal/gateway/gateway.go
+++ b/internal/gateway/gateway.go
@@ -439,19 +439,15 @@ func Start(ctx context.Context, cfg StartConfig) (StartResult, error) {
 		mounts = appendReadOnlyMount(mounts, cfg.SecretReleaseFile, model.GatewaySecretReleasePath)
 	}
 	if cfg.TLSRootDir != "" {
+		// Only the CA is shared with the host. Leaf certificates are minted
+		// per gateway inside the container: a host-side leaf cache had to be
+		// chowned to the proxy user, and that UID differs between docker
+		// (host UID) and rootless podman (subuid), so a host that runs both
+		// engines left the cache unreadable for the other one.
 		caCertPath := filepath.Join(cfg.TLSRootDir, "ca.crt")
 		caKeyPath := filepath.Join(cfg.TLSRootDir, "ca.key")
-		hostsPath := filepath.Join(cfg.TLSRootDir, "hosts")
-		if err := os.MkdirAll(hostsPath, 0o700); err != nil {
-			return empty, fmt.Errorf("failed to create gateway TLS hosts dir: %w", err)
-		}
 		mounts = appendReadOnlyMount(mounts, caCertPath, model.GatewayTLSCACertPath)
 		mounts = appendReadOnlyMount(mounts, caKeyPath, model.GatewayTLSCAKeyPath)
-		mounts = append(mounts, docker.Mount{
-			Type:   docker.MountTypeBind,
-			Source: hostsPath,
-			Target: model.GatewayTLSHostsPath,
-		})
 	}
 
 	env := []string{}
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -127,7 +127,7 @@ The restricted network request flow has a separate
 - [`internal/gateway/bundle/bundle.go`](../internal/gateway/bundle/bundle.go) writes host-managed gateway bundles (`dnsmasq.conf`, `domains.txt`, `meta.json`) used for runtime apply/reload.
 - [`internal/gateway/mitm/`](../internal/gateway/mitm/) implements TLS MITM request forwarding, allowlist checks, and secret placeholder header rewriting/blocking. It writes audit events through `internal/netlog` so writer and reader cannot drift apart.
 - [`internal/gateway/dnsaudit/`](../internal/gateway/dnsaudit/) tails dnsmasq's log and translates denied or failed lookups into `dns` audit events. It runs as its own process (`enclave-gateway-proxy -dns-audit <log>`) so DNS denials are recorded even when the proxy is disabled.
-- [`internal/gateway/tlsstore/`](../internal/gateway/tlsstore/) manages CA generation and per-host leaf certificate cache files for gateway TLS interception.
+- [`internal/gateway/tlsstore/`](../internal/gateway/tlsstore/) manages CA generation and the container-private per-host leaf certificate cache for gateway TLS interception; only the CA is shared with the host.
 - [`internal/docker/`](../internal/docker/) wraps the Docker CLI for Docker-specific backend/build/cleanup implementation details.
 
 ### Tool Handlers
--- a/docs/runtime/network-request-flow.md
+++ b/docs/runtime/network-request-flow.md
@@ -27,7 +27,7 @@ sequenceDiagram
 
     Note over Host,API: Session startup
     Host->>Entry: Start gateway with config bundle<br/>(dnsmasq.conf, domains.txt, meta.json)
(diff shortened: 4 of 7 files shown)


gateway-entrypoint.sh
```diff
@@ -483,12 +483,10 @@ prepare_proxy_runtime_data() {
...
-        if [ -d "${tls_root}/hosts" ]; then
-            chown "$PROXY_USER:$PROXY_USER" "${tls_root}/hosts" 2>/dev/null || true
-            ln -sfn "${tls_root}/hosts" "$proxy_tls/hosts"
-        else
-            mkdir -p "$proxy_tls/hosts"
-        fi
+        # Leaf certificates stay container-private; nothing under the host
+        # TLS root is ever chowned to a container UID.
+        mkdir -p "$proxy_tls/hosts"
+        chmod 700 "$proxy_tls/hosts"
```
EclipseSourceAI, line 489: `mkdir -p` plus `chmod 700` duplicates what the store already does: `EnsureLeaf` runs `MkdirAll(HostsDir, 0o700)` itself ([store.go#L79-L81](https://github.com/eclipse-enclave/enclave/blob/b1f56f533135de5953227c8f8161d9d44385ea77/internal/gateway/tlsstore/store.go#L79-L81)), and `$proxy_tls` is chowned to the proxy user on the next line, so the proxy can create it. The new `chmod` is also the only command in this block without `2>/dev/null || true`.

internal/gateway/gateway.go
```diff
@@ -439,19 +439,15 @@ func Start(ctx context.Context, cfg StartConfig) (StartResult, error) {
 		mounts = appendReadOnlyMount(mounts, cfg.SecretReleaseFile, model.GatewaySecretReleasePath)
 	}
 	if cfg.TLSRootDir != "" {
+		// Only the CA is shared with the host. Leaf certificates are minted
+		// per gateway inside the container: a host-side leaf cache had to be
+		// chowned to the proxy user, and that UID differs between docker
+		// (host UID) and rootless podman (subuid), so a host that runs both
+		// engines left the cache unreadable for the other one.
```
EclipseSourceAI, line 446: The PR body says the leftover `~/.local/state/enclave/tls/hosts` can be deleted, but the old entrypoint chowned it to the proxy UID at mode 0700, so a plain `rm -rf` fails with EPERM (needs `sudo`, or `podman unshare` for rootless). Worth spelling that out in the release notes, since the leftover holds leaf private keys for every host that was ever MITM'd.

internal/gateway/gateway.go
```diff
@@ -439,19 +439,15 @@ func Start(ctx context.Context, cfg StartConfig) (StartResult, error) {
 		mounts = appendReadOnlyMount(mounts, cfg.SecretReleaseFile, model.GatewaySecretReleasePath)
 	}
 	if cfg.TLSRootDir != "" {
+		// Only the CA is shared with the host. Leaf certificates are minted
```
EclipseSourceAI, line 442: The cost is a bit more than "one extra mint per host per session": `EnsureLeaf` holds one store-wide mutex across the RSA-2048 keygen ([store.go#L68-L108](https://github.com/eclipse-enclave/enclave/blob/b1f56f533135de5953227c8f8161d9d44385ea77/internal/gateway/tlsstore/store.go#L68-L108)), so under `--network-log requests` the first handshake to each allowlisted host serializes behind every other cold one. Now that leaves never leave the container, making `writeLeaf` mint ECDSA P-256 instead of RSA ([store.go#L206-L210](https://github.com/eclipse-enclave/enclave/blob/b1f56f533135de5953227c8f8161d9d44385ea77/internal/gateway/tlsstore/store.go#L206-L210)) would drop that to near zero.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Fixes #80 by dropping the host bind mount for the gateway's per-host leaf certificate cache. Only `ca.crt`/`ca.key` are still shared with the host (read-only), and `gateway-entrypoint.sh` now creates `/var/lib/enclave/proxy/tls/hosts` inside the container instead of symlinking to the mounted host dir and chowning it to the in-container proxy user. `model.GatewayTLSHostsPath`, `config.HostTLSHostsDir` and the two host-side `MkdirAll` calls go away with it.

The diagnosis holds up: the old entrypoint chowned a host path to a UID that means something different under docker vs rootless podman, so whichever engine ran second could not write leaves. No `VOLUME` in `Dockerfile.gateway` and `AutoRemove: true` on the sidecar, so the new cache really is per container. Old images degrade gracefully (the removed `hosts` mount just takes the `else` branch), and `make build` plus the gateway/config/backend tests pass here.

- The stale `~/.local/state/enclave/tls/hosts` left on upgraded hosts is not removable without root, which the PR body understates.
- The cold-cache cost per session is larger than "one extra mint per host" because `EnsureLeaf` serializes RSA keygen behind one mutex.
- There is no regression test pinning the gateway mount set, so nothing stops the host leaf mount from coming back.

Note that this change also silently improves behaviour the PR does not claim: `EnsureLeaf` never checks leaf expiry, so the old shared host cache could serve leaves past their 3-month `NotAfter`. An ephemeral cache sidesteps that.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Follow-up review. Since my last pass, the only change is 3e5065c ("tighten gateway leaf cache handling"), which addresses two of my three earlier points:

- The entrypoint no longer creates/chmods `$proxy_tls/hosts` itself; that's left entirely to `EnsureLeaf`, and the leftover `chown` line no longer references the removed dir.
- `docs/networking.md` now documents that the legacy `~/.local/state/enclave/tls/hosts` cache needs `sudo` or `podman unshare` to remove.

The `internal/gateway/gateway.go` mount-building change in that commit is a clean extraction (`appendTLSCAMounts`) with a matching unit test in the existing `reflect.DeepEqual` style used elsewhere in the package. No new issues found in the delta.

My mutex/RSA-keygen serialization point (on `EnsureLeaf`) is still open and not addressed by this commit; leaving that to the existing thread rather than repeating it here.

- [dropped the redundant `mkdir -p`/`chmod 700` for the leaf dir in the entrypoint](https://github.com/eclipse-enclave/enclave/pull/81#discussion_r3996516879)
- [documented that the leftover host TLS cache needs sudo/podman unshare to remove](https://github.com/eclipse-enclave/enclave/pull/81#discussion_r3996516912)

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Only change since my last pass is fed3534 ("mint gateway leaf certificates with ECDSA P-256"), plus a merge of unrelated origin/main commits (skill validation, docs) that don't touch this PR's code (confirmed via `git diff origin/main...HEAD`, still just the 11 files from the original fix).

fed3534 switches `writeLeaf` from RSA-2048 to ECDSA P-256 for leaf certs, keeping the CA on RSA. This is exactly the fix for my mutex/keygen-serialization point: `writeLeaf` still runs under `Store.mu` inside `EnsureLeaf`, but ECDSA P-256 keygen is microseconds instead of ~130ms, so the serialization no longer matters in practice. `KeyUsage` was also correctly trimmed to drop `KeyEncipherment` (meaningless for ECDSA), and the test was updated to check for its absence. No new issues in this delta.

- [mutex/RSA keygen serialization cost on cold leaf minting](https://github.com/eclipse-enclave/enclave/pull/81#discussion_r3996516943)

tortmayr (approved):
Code looks good to me. Executed testplan and everything seems to work as expected!

PR #83, opened 2026-09-12
Allow extensions to declare project-scoped cache mounts

### LINKED ISSUE #82: Allow extensions to declare project-scoped cache mounts
The persistent per-project package caches (npm, pip, go, cargo, pnpm, uv, yarn, bun, nvm) are hard-coded in addCacheMounts (internal/runtime/runtime.go). Several entries exist only for specific features (nvm for node-dev, uv/pip for python-dev), and extensions have no way to add their own — a java-dev feature, for example, cannot persist the Maven local repository (~/.m2/repository), so every fresh container re-downloads all dependencies.

Proposal

Add a declarative caches field to extension specs (mixin and sandbox):

caches:
  - name: m2               # subdirectory under the enclave-managed project cache dir
    target: .m2/repository # container path inside $HOME

At session start, enabled extensions' entries join the built-in list in addCacheMounts: bind mount HostCacheToolProjectDir(...)/<name> → <home>/<target>, respecting --no-cache.

Key constraints:

- The host side stays enclave-managed — specs only name a cache subdirectory and container target, never a host path.
- Validate targets at load time: cleaned, no .., inside the container home, no collision with reserved paths or other features' caches (same name + same target dedupes; same name with different targets is an error).
- Surface declared caches in the capability summary (spec_summary.go) so features add shows them for third-party extensions.
- Document that an empty cache mount shadows image-baked content at the target (seed from feature-entrypoint.d, as node-dev does for nvm).

Follow-up: migrate the feature-specific built-in entries (nvm, uv, pip, pnpm, yarn) into their features' specs, keeping the same name values so existing host cache directories remain valid.

tortmayr:
#### What it does

The persistent per-project package caches were hard-coded in `addCacheMounts`, so extensions had no way to persist their own directories — a `java-dev` feature, for example, re-downloads the whole Maven repository in every fresh container.

- Add a declarative `caches` field to extension specs (`kind: sandbox` and `kind: mixin`): each entry names a subdirectory under the enclave-managed project cache dir and a container target inside `$HOME` — a spec can never name a host path
- At session start, entries from the tool spec and the enabled features join the built-in list in `addCacheMounts`, respecting `--no-cache`
- Validate at load time: targets are cleaned, traversal-free, inside the container home, and must not collide with built-in cache names/targets or reserved mounts (auth stores, shell history)
- Validate across extensions at session start: identical `name`+`target` declarations dedupe; the same name with a different target — or the same target under a different name — fails with an error naming both claimants
- Surface declared caches in the capability summary and update diff shown by `enclave features|tools add`
- Document the `caches` field, including that an empty cache mount shadows image-baked content at the target (seed from `feature-entrypoint.d`, as node-dev does for nvm)

Fixes #82

#### How to test

```bash
```

- Declare a cache in a feature spec, e.g. in `extensions/features/devtools/spec.yaml`:
  ```yaml
  caches:
    - name: m2
      target: .m2/repository
  ```
- Start a session and verify `~/.cache/enclave/<tool>/<project-hash>/m2` exists on the host and is mounted at `~/.m2/repository` in the container; files written there survive a fresh container
- `--no-cache` starts the session without the mount
- Two features declaring the same name with different targets fail session start with an error naming both features
- `enclave features add` against an extension repo with a `caches` entry lists it in the capability summary

- [x] Verified locally: `make build`, `make test`, `make lint`, `make check-license-headers`

#### Follow-ups

Migrate the feature-specific built-in entries (nvm, uv, pip, pnpm, yarn) into their features' specs, keeping the same `name` values so existing host cache directories stay valid.

#### Breaking changes

#### Review checklist

### DIFF
--- a/docs/extensions/README.md
+++ b/docs/extensions/README.md
@@ -107,6 +107,7 @@ environment: {...}
 credentials: {...}
 providers: [...]          # enclave-native, see "Tool Extensions" below
 ports: [...]              # enclave-native; honored for tools and enabled features
+caches: [...]             # enclave-native; persistent per-project cache mounts, see below
 
 # mixin-only fields (kind: mixin)
 priority: <int, default 100>
@@ -183,6 +184,42 @@ Honored for both sandbox and mixin extensions:
   entry must match a declared `credentials.sources.<id>.env` alias; a typo is
   a load error. Entries are unioned across the tool spec and all enabled
   mixins.
+
+### Project caches (`caches`)
+
+Honored for both sandbox and mixin extensions (enclave-native, not part of the
+sbx format). Each entry asks the runtime to persist one directory of the
+container home across sessions of the same project — a `java-dev` feature can
+keep the Maven local repository, for example:
+
+```yaml
+caches:
+  - name: m2               # subdirectory under the enclave-managed project cache dir
+    target: .m2/repository # container path inside $HOME
+```
+
+At session start the declared entries join the built-in package caches (npm,
+pip, go, go-build, cargo, pnpm, uv, yarn, bun, nvm):
+`~/.cache/enclave/<tool>/<project-hash>/<name>` on the host is bind-mounted at
+`$HOME/<target>` in the container. `--no-cache` disables extension caches along
+with the built-in ones, and `enclave cleanup` removes the whole cache dir
+unless `--keep cache` is set.
+
+The host side is always enclave-managed: a spec names a cache subdirectory and
+a container target, never a host path. Load-time validation rejects a `name`
+outside the extension-name charset, a `target` that is absolute, contains
+traversal, or resolves outside the container home, and any collision with the
+built-in cache names/targets or the runtime's reserved mounts (auth stores,
+shell history). Across the tool spec and all enabled features, an identical
+`name` + `target` pair declared twice dedupes; the same name with a different
+target — or the same target under a different name — fails session start with
+an error naming both claimants.
+
+A cache mount starts empty and **shadows image-baked content** at its target:
+anything the image installed there is invisible until the cache is seeded.
+Seed it from a `feature-entrypoint.d/` script on first start, the way
+`node-dev` copies the image's default Node version into the persistent `nvm`
+cache.
 
 ### Examples
 
@@ -631,6 +668,12 @@ the same container port. `hostAllocation` selects the host port: `fixed` (the
 default) mirrors the container port, while `auto` publishes with an OS-assigned
 host port (the `-p 0:<port>` form) so concurrent sessions do not contend — the
 resolved port appears in the printed `openUrl` and in `enclave ps`.
+
+### Declared caches
+
+A feature can declare top-level `caches` exactly like a tool (see
+[Project caches](#project-caches-caches)). Entries apply only to sessions that
+enable the feature.
 
 ### Available Features
 
--- a/internal/config/caches.go
+++ b/internal/config/caches.go
@@ -1,0 +1,85 @@
+// Copyright (C) 2026 EclipseSource GmbH and others.
+//
+// This program and the accompanying materials are made available under the
+// terms of the MIT License, which is available in the project root.
+//
+// SPDX-License-Identifier: MIT
+
+package config
+
+import (
+	"fmt"
+
+	"enclave/internal/model"
+)
+
+// reservedCacheTargets maps each container-home-relative path the runtime
+// already claims to a label naming its owner, so a cache colliding with one is
+// rejected with a message that says what it collides with. Besides the
+// built-in caches these are the auth-store mounts and the shell-history mount
+// (see runtime.addHistoryMounts).
+func reservedCacheTargets() map[string]string {
+	reserved := map[string]string{
+		model.ContainerAuthDir:        "the auth store",
+		model.ContainerFeatureAuthDir: "the feature auth store",
+		model.ContainerHistoryDir:     "the shell history store",
+	}
+	for _, cache := range model.BuiltinProjectCaches {
+		reserved[cache.Target] = fmt.Sprintf("the built-in %q cache", cache.Name)
+	}
+	return reserved
+}
+
+// validateAndNormalizeCaches checks a spec's declared cache mounts. Each name
+// becomes a host directory under the enclave-managed project cache dir, so it
+// is held to the extension-name charset; each target is a container path that
+// must stay inside the container home. Collisions with built-in cache names
+// and reserved targets are load errors. Within the list, a repeated name with
+// the same target dedupes, while a name or target redeclared with a different
+// counterpart is an error. Cross-extension collisions can only be checked at
+// session start, where the tool spec and the enabled features meet.
+func validateAndNormalizeCaches(caches []model.CacheConfig) ([]model.CacheConfig, error) {
+	if len(caches) == 0 {
+		return nil, nil
+	}
+	builtinNames := map[string]struct{}{}
+	for _, cache := range model.BuiltinProjectCaches {
+		builtinNames[cache.Name] = struct{}{}
+	}
+	reservedTargets := reservedCacheTargets()
+
+	targetByName := map[string]string{}
+	nameByTarget := map[string]string{}
+	out := make([]model.CacheConfig, 0, len(caches))
+	for i, cache := range caches {
+		if cache.Name == "" {
+			return nil, fmt.Errorf("caches[%d]: name must not be empty", i)
+		}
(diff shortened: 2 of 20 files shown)


internal/config/caches.go
```diff
@@ -0,0 +1,85 @@
...
+// already claims to a label naming its owner, so a cache colliding with one is
+// rejected with a message that says what it collides with. Besides the
+// built-in caches these are the auth-store mounts and the shell-history mount
+// (see runtime.addHistoryMounts).
+func reservedCacheTargets() map[string]string {
+	reserved := map[string]string{
+		model.ContainerAuthDir:        "the auth store",
+		model.ContainerFeatureAuthDir: "the feature auth store",
+		model.ContainerHistoryDir:     "the shell history store",
+	}
```
EclipseSourceAI, line 26: The SSH mount and the tool config file mounts are missing here, so `target: .ssh` or `target: .npmrc` passes load-time validation and then produces two mounts at the same container path, which docker rejects with a duplicate mount point error at session start. Both are regular home mounts: [.ssh](https://github.com/eclipse-enclave/enclave/blob/b1f56f533135de5953227c8f8161d9d44385ea77/internal/runtime/runtime.go#L1148) and [npmrc/yarnrc/...](https://github.com/eclipse-enclave/enclave/blob/b1f56f533135de5953227c8f8161d9d44385ea77/internal/runtime/runtime.go#L1295-L1299).

internal/config/caches.go
```diff
@@ -0,0 +1,85 @@
...
+			return nil, fmt.Errorf("caches[%d]: invalid name %q: %s", i, cache.Name, ExtensionNameCharset)
+		}
+		target, err := cleanHomeRelativePath(cache.Target)
+		if err != nil {
+			return nil, fmt.Errorf("caches[%d] (%q): target: %w", i, cache.Name, err)
+		}
+		if _, ok := builtinNames[cache.Name]; ok {
+			return nil, fmt.Errorf("caches[%d]: name %q is reserved by a built-in cache", i, cache.Name)
+		}
+		if owner, ok := reservedTargets[target]; ok {
```
EclipseSourceAI, line 68: The reservation is exact-match only, so `.enclave-auth/tokens` or `.cache/pip/wheels` gets through and lands nested inside an existing mount. A prefix check in both directions would close it, and the same gap exists in the cross-extension check in `projectCaches`.

internal/runtime/runtime.go
```diff
@@ -1200,42 +1202,74 @@ func (r *Runtime) addSessionMonitorEnv(mounts *mountAccumulator) {
...
+		ownerByTarget[cache.Target] = owner
+		caches = append(caches, cache)
+		return nil
+	}
+	for _, cache := range model.BuiltinProjectCaches {
+		ownerByName[cache.Name] = "the built-in cache list"
+		targetByName[cache.Name] = cache.Target
+		ownerByTarget[cache.Target] = "the built-in cache list"
+	}
+	for _, cache := range r.profile.Caches {
```
EclipseSourceAI, line 1260: `profile.ConfigDir` and `profile.MemoryDir` are home mounts too and are not in the claim set, so a tool declaring a cache at its own `memory_dir` fails at docker run rather than with the error built here. See [config dir](https://github.com/eclipse-enclave/enclave/blob/b1f56f533135de5953227c8f8161d9d44385ea77/internal/runtime/runtime.go#L1513) and [memory dir](https://github.com/eclipse-enclave/enclave/blob/b1f56f533135de5953227c8f8161d9d44385ea77/internal/runtime/runtime.go#L1268).

internal/runtime/runtime.go
```diff
@@ -1200,42 +1202,74 @@ func (r *Runtime) addSessionMonitorEnv(mounts *mountAccumulator) {
...
 
 	if r.run.NoCache {
-		return
+		return nil
+	}
+	caches, err := r.projectCaches()
+	if err != nil {
+		return err
 	}
 	cacheDir := config.HostCacheToolProjectDir(r.host.Home, r.profile.Name, r.project.Hash)
```
EclipseSourceAI, line 1216: The cache dir is keyed by tool ([HostCacheToolProjectDir](https://github.com/eclipse-enclave/enclave/blob/b1f56f533135de5953227c8f8161d9d44385ea77/internal/config/host_paths.go#L285-L287)), so a feature-declared `m2` cache is re-downloaded per tool in the same project. That is fine for the built-ins (they are tool-agnostic by accident), but it works against the java-dev motivation in the description. A maintainer should decide whether feature caches belong under a project-only key.

docs/extensions/README.md
```diff
@@ -184,6 +185,42 @@ Honored for both sandbox and mixin extensions:
...
+built-in cache names/targets or the runtime's reserved mounts (auth stores,
+shell history). Across the tool spec and all enabled features, an identical
+`name` + `target` pair declared twice dedupes; the same name with a different
+target — or the same target under a different name — fails session start with
+an error naming both claimants.
+
+A cache mount starts empty and **shadows image-baked content** at its target:
+anything the image installed there is invisible until the cache is seeded.
+Seed it from a `feature-entrypoint.d/` script on first start, the way
+`node-dev` copies the image's default Node version into the persistent `nvm`
```
EclipseSourceAI, line 221: The other half of this gotcha is the parent directory: the built-in targets are pre-created for the container user in the [Dockerfile](https://github.com/eclipse-enclave/enclave/blob/b1f56f533135de5953227c8f8161d9d44385ea77/Dockerfile#L190-L192), but for an arbitrary target the container runtime creates the missing parent as root, so the agent cannot write siblings (e.g. `~/.m2/settings.xml` next to a `.m2/repository` cache). Worth spelling out that the extension has to create the parent in its install step.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Adds a declarative `caches` field to tool and feature specs so extensions can persist directories of the container home per project, replacing the hard-coded list in `addCacheMounts` with `model.BuiltinProjectCaches` plus spec-declared entries. Validation is split in two: per-spec at load time (`validateAndNormalizeCaches`, reusing the renamed `cleanHomeRelativePath`) and cross-extension at session start (`projectCaches`), which mirrors the existing `activeSecrets` merge pattern. The capability summary and install diff pick up the new field, and the `.shell_history` literal is pulled into a `model.ContainerHistoryDir` constant.

The approach fits the codebase well and the refactor of `addCacheMounts` is a clean net win. Build and tests pass locally.

The main thing to look at is the completeness of the reserved-target set. It covers the auth stores, shell history, and built-in caches, but not the SSH mount, the npmrc/yarnrc file mounts, or the profile's own `config_dir`/`memory_dir`, and it matches targets exactly rather than by prefix. Those cases turn what should be a clear load-time error into a raw docker duplicate-mount failure or a silently nested mount. Beyond that, the per-tool cache key and the root-owned parent directory for arbitrary targets are both worth a maintainer decision.

internal/config/caches.go
tortmayr: Fixed in 97f881f: `.ssh` and the persistent home-config file mounts are now in the reserved set. Their paths moved to shared model constants (`ContainerSSHDir`, `ContainerHomeConfigFiles`) that the runtime mounts from too, so the two lists cannot drift.

internal/config/caches.go
tortmayr: Fixed in 97f881f: reservation and within-spec checks now treat nesting in either direction as a collision (`model.CacheTargetsOverlap`), and `projectCaches` applies the same overlap rule to the cross-extension claims.

internal/runtime/runtime.go
tortmayr: Fixed in 97f881f: `projectCaches` now claims the tool's resolved `config_dir` and `memory_dir` (when they resolve under the container home), so a colliding cache fails session start with an error naming the mount instead of a docker duplicate-mount failure.

internal/runtime/runtime.go
tortmayr: Maintainer decision: keeping the per-tool key. It matches the existing built-in cache layout (npm/go caches are already duplicated per tool today) and the per-tool cleanup contract. Moving to a project-scoped key is a separate layout decision that should cover built-ins and extension caches together, with migration of existing dirs — out of scope for this PR.

docs/extensions/README.md
tortmayr: Added in 97f881f: the caches section now notes that a missing target parent is created root-owned by the container runtime, so an extension should create and chown the parent in its install.sh when the agent needs to write siblings next to the mount.

PR #84, opened 2026-09-12
feat: allow extensions to declare project-scoped cache mounts

### LINKED ISSUE #82: Allow extensions to declare project-scoped cache mounts

tortmayr:
#### What it does

The persistent per-project package caches were hard-coded in `addCacheMounts`, so extensions had no way to persist their own directories — a `java-dev` feature, for example, re-downloads the whole Maven repository in every fresh container.

- Add a declarative `caches` field to extension specs (`kind: sandbox` and `kind: mixin`): each entry names a subdirectory under the enclave-managed project cache dir and a container target inside `$HOME` — a spec can never name a host path
- At session start, entries from the tool spec and the enabled features join the built-in list in `addCacheMounts`, respecting `--no-cache`
- Validate at load time: targets are cleaned, traversal-free, inside the container home, and must not collide with built-in cache names/targets or reserved mounts (auth stores, shell history)
- Validate across extensions at session start: identical `name`+`target` declarations dedupe; the same name with a different target — or the same target under a different name — fails with an error naming both claimants
- Surface declared caches in the capability summary and update diff shown by `enclave features|tools add`
- Document the `caches` field, including that an empty cache mount shadows image-baked content at the target (seed from `feature-entrypoint.d`, as node-dev does for nvm)

Fixes #82

#### How to test

```bash
```

- Declare a cache in a feature spec, e.g. in `extensions/features/devtools/spec.yaml`:
  ```yaml
  caches:
    - name: m2
      target: .m2/repository
  ```
- Start a session and verify `~/.cache/enclave/<tool>/<project-hash>/m2` exists on the host and is mounted at `~/.m2/repository` in the container; files written there survive a fresh container
- `--no-cache` starts the session without the mount
- Two features declaring the same name with different targets fail session start with an error naming both features
- `enclave features add` against an extension repo with a `caches` entry lists it in the capability summary

- [x] Verified locally: `make build`, `make test`, `make lint`, `make check-license-headers`

#### Follow-ups

Migrate the feature-specific built-in entries (nvm, uv, pip, pnpm, yarn) into their features' specs, keeping the same `name` values so existing host cache directories stay valid.

#### Breaking changes

#### Review checklist


internal/model/types.go
```diff
@@ -242,6 +246,45 @@ type Profile struct {
...
+// never name a host path.
+type CacheConfig struct {
+	Name   string `json:"name"`
+	Target string `json:"target"`
+}
+
+// CacheTargetsOverlap reports whether two home-relative mount targets collide:
+// equal or nested either way. Nested bind mounts depend on mount order and
+// silently shadow each other, so overlap counts as a collision.
+func CacheTargetsOverlap(a string, b string) bool {
```
EclipseSourceAI: `util.PathWithin` already does equal-or-inside ([link](https://github.com/eclipse-enclave/enclave/blob/b1f56f533135de5953227c8f8161d9d44385ea77/internal/util/util.go#L76-L92)), so this is a hand-rolled version of it applied twice. It cannot be reused from `model` (model -> util -> logx -> model), which is an argument for putting `CacheTargetsOverlap` and `BuiltinProjectCaches` in `config`, where `util` is already imported and `runtime` still reaches them.

internal/runtime/runtime.go
```diff
@@ -1200,42 +1202,97 @@ func (r *Runtime) addSessionMonitorEnv(mounts *mountAccumulator) {
...
 	cacheDir := config.HostCacheToolProjectDir(r.host.Home, r.profile.Name, r.project.Hash)
+	for _, cache := range caches {
+		hostDir := filepath.Join(cacheDir, cache.Name)
+		_ = os.MkdirAll(hostDir, 0o700)
+		mounts.AddMount(bindMount(hostDir, r.containerHome+"/"+cache.Target, false))
+	}
+	return nil
+}
+
+// cacheClaim records one home-relative target already taken, either by a cache
```
EclipseSourceAI: The cache-claim logic lands in `runtime.go` while its tests go into a new `cache_mounts_test.go`. The package already splits this kind of thing into its own file ([skills_mount.go](https://github.com/eclipse-enclave/enclave/blob/b1f56f533135de5953227c8f8161d9d44385ea77/internal/runtime/skills_mount.go#L36-L40)), so a `cache_mounts.go` would fit better and keeps runtime.go from growing again.

internal/runtime/runtime.go
```diff
@@ -1200,42 +1202,97 @@ func (r *Runtime) addSessionMonitorEnv(mounts *mountAccumulator) {
...
 
 	if r.run.NoCache {
-		return
+		return nil
+	}
+	caches, err := r.projectCaches()
+	if err != nil {
+		return err
 	}
 	cacheDir := config.HostCacheToolProjectDir(r.host.Home, r.profile.Name, r.project.Hash)
```
EclipseSourceAI: Feature-declared caches land under `HostCacheToolProjectDir(..., r.profile.Name, ...)`, so a `java-dev` m2 cache is a separate host directory per tool in the same project and Maven re-downloads after switching from claude to codex. That is fine for the built-ins (they are per-tool by design), less obvious for features, which are tool-independent. A maintainer should confirm that is the intended keying.

docs/persistence.md
```diff
@@ -45,7 +45,7 @@ Per-project data is stored on the host and reused across sessions:
 
 | Data | Location |
 |------|----------|
-| Package caches | `~/.cache/enclave/<tool>/<project-hash>/` |
+| Package caches | `~/.cache/enclave/<tool>/<project-hash>/` (built-in plus [extension-declared](extensions/README.md#project-caches-caches) caches) |
```
EclipseSourceAI, line 48: `docs/ARCHITECTURE.md` carries the detailed cache layout with a hard-coded subdirectory list ([link](https://github.com/eclipse-enclave/enclave/blob/b1f56f533135de5953227c8f8161d9d44385ea77/docs/ARCHITECTURE.md#L213-L222)) and is not touched here. It should point at extension-declared entries the same way this table now does (it is also still missing `nvm/`).

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Adds a declarative `caches` field to tool and feature specs so extensions can persist directories of the container home per project, replacing the hard-coded list in `addCacheMounts` with `model.BuiltinProjectCaches` plus spec-declared entries.

The approach looks sound: specs only name a cache subdirectory and a `$HOME`-relative target (never a host path), per-spec validation happens at load (`validateAndNormalizeCaches`), cross-extension conflicts are resolved at session start (`projectCaches`), and the new entries show up in the `features|tools add` capability summary and diff. Refactoring the shell-history, SSH, and home-config-file paths into `model` constants so they can serve as reserved targets is a reasonable side effect rather than an unrelated drive-by. Tests cover the load-time rules, the dedupe/conflict matrix, and `--no-cache`, and `go build` plus the tests for the touched packages pass here.

Worth a closer look:
- `CacheTargetsOverlap` reimplements `util.PathWithin` because `model` cannot import `util`, which suggests the helper and the built-in list belong in `config`.
- Feature-declared caches are keyed per tool, so a tool-independent feature cache (the motivating Maven case) is duplicated across tools in one project.
- `docs/ARCHITECTURE.md` still documents the cache directory as a fixed list.

Nothing blocking found in the validation logic itself. The reserved-target set covers the auth stores, shell history, SSH mount, home config files, and built-in caches, and the tool config/memory store targets are checked at session start, which closes the collisions that would otherwise surface as raw container-start errors.

internal/model/types.go
tortmayr: Done in 0e0f6e3: `CacheTargetsOverlap` and `BuiltinProjectCaches` moved to `config`, with the overlap check now implemented as `util.PathWithin` in both directions. The `model.CacheConfig` type stays in model since Profile/Extension carry it.

internal/runtime/runtime.go
tortmayr: Done in 0e0f6e3: `addCacheMounts`, `projectCaches`, `profileMountClaims`, and `cacheClaim` moved to `internal/runtime/cache_mounts.go`, matching the existing test file.

docs/persistence.md
tortmayr: Done in 0e0f6e3: the ARCHITECTURE.md cache list now includes `nvm/` and a `<name>/` entry pointing at the extension-declared caches doc.

internal/runtime/runtime.go
tortmayr: Confirmed (maintainer decision): the per-tool keying is intended. It matches the existing built-in cache layout and the per-tool cleanup contract; a project-scoped key would be a separate layout decision covering built-ins and extension caches together, including migration of existing dirs.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Follow-up review. The only change since my last pass (238157f → 0e0f6e3) is the refactor addressing my earlier comments: `CacheTargetsOverlap`/`BuiltinProjectCaches` moved from `model` to `config`, the cache-mount logic moved from `runtime.go` into `cache_mounts.go`, and `docs/ARCHITECTURE.md` now lists `nvm/` and points at extension-declared caches. I checked the diff and it's a clean move, `go build` and the `runtime`/`config`/`model` tests pass. No new issues found.

- [move `CacheTargetsOverlap`/`BuiltinProjectCaches` into `config`](https://github.com/eclipse-enclave/enclave/pull/84#discussion_r3997512555)
- [split cache-mount logic into its own `cache_mounts.go` file](https://github.com/eclipse-enclave/enclave/pull/84#discussion_r3997512584)
- [document `nvm/` and extension-declared caches in ARCHITECTURE.md](https://github.com/eclipse-enclave/enclave/pull/84#discussion_r3997512668)

PR #85, opened 2026-09-14
docs: note that Enclave has no auto-update yet

planger:
Adds an "Updating" subsection to the website's Getting Started install docs.

It states that Enclave neither updates itself nor notifies about newer builds, and tells users to periodically re-download the rolling release assets and re-run the same install command they used initially (apt, dnf, `install`, or a source rebuild).

Docs-only change.


website/docs/docs/getting-started.md
```diff
@@ -132,6 +132,15 @@ cd enclave
 make build
 ```
 
+### Updating
+
+Enclave does not update itself yet, and it does not tell you when a newer build
```
EclipseSourceAI: `enclave update` already exists (rebuilds tool images) and agent CLIs inside the sandbox do refresh on their own 24h interval, see [docs/tools.md](https://github.com/eclipse-enclave/enclave/blob/b1f56f533135de5953227c8f8161d9d44385ea77/docs/tools.md#L74-L86). Under a bare "Updating" heading this blanket sentence reads as if nothing updates. Retitle to "Updating Enclave" and scope the sentence to the Enclave binary.

website/docs/docs/getting-started.md
```diff
@@ -132,6 +132,15 @@ cd enclave
...
+is available. Since the rolling release moves with `main`, download the current
```
EclipseSourceAI: "every so often" gives users nothing to compare against. `enclave version` prints version, commit and date ([cli-reference](https://github.com/eclipse-enclave/enclave/blob/b1f56f533135de5953227c8f8161d9d44385ea77/docs/cli-reference.md#L81-L81)), so pointing at it makes the check actionable.

website/docs/docs/getting-started.md
```diff
@@ -132,6 +132,15 @@ cd enclave
...
+assets every so often and re-run the same install step you used above: `sudo apt
```
EclipseSourceAI: Every other command on this page lives in a fenced bash block. This packs four into one prose sentence, with an inline code span broken across lines 139/140, and `sudo install ...` with a literal ellipsis is not copy-pasteable. A short list with one code block per install path matches the rest of the section.

website/docs/docs/getting-started.md
```diff
@@ -132,6 +132,15 @@ cd enclave
...
+install ./enclave_*_amd64.deb`, `sudo dnf install ./enclave-*.x86_64.rpm`,
```
EclipseSourceAI: The enumeration claims to cover "the same install step you used above" but skips two paths from this same page: the Windows launcher, where the manifest has to be re-downloaded because the pinned hashes change ([packaging/scoop/README.md](https://github.com/eclipse-enclave/enclave/blob/b1f56f533135de5953227c8f8161d9d44385ea77/packaging/scoop/README.md#L12-L14)), and macOS, where a re-downloaded binary carries the quarantine attribute again so `xattr -d` has to run before `install` ([install block](https://github.com/eclipse-enclave/enclave/blob/b1f56f533135de5953227c8f8161d9d44385ea77/website/docs/docs/getting-started.md#L74-L87)).

website/docs/docs/getting-started.md
```diff
@@ -132,6 +132,15 @@ cd enclave
...
+`sudo install ...` over the binary, or `git pull && make build` for a source
```
EclipseSourceAI: `make build` only writes `./bin/enclave`, so it leaves an installed binary stale. Source installs go through `make install` ([README](https://github.com/eclipse-enclave/enclave/blob/b1f56f533135de5953227c8f8161d9d44385ea77/README.md#L126-L139)), so `git pull && make install` is the update step for anyone who installed that way.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Docs-only change: adds an "Updating" subsection to the website Getting Started page, stating that Enclave has no self-update and no update notification, and telling users to re-download the rolling assets and re-run their original install step. Scope is tight and the core claim checks out (no version check or self-update exists in the binary).

Main points for a maintainer:

- The wording collides with the existing update surfaces. `enclave update` rebuilds tool images and agent CLIs auto-refresh on an interval, so a blanket "Enclave does not update itself yet" under a heading called "Updating" is ambiguous about what does not update.
- The enumerated update commands do not cover all install paths on the same page: the Windows Scoop launcher and the macOS quarantine step are missing, and `git pull && make build` does not refresh a binary installed with `make install`.
- Presentation deviates from the rest of the page (four commands inside one prose sentence instead of fenced blocks, one code span split across lines, a non-copy-pasteable `sudo install ...`).

The same gap exists in `README.md`, which has install instructions and no updating note. Whether that stays website-only is a maintainer call.

planger:
Addressed the review:

- Retitled to "Updating Enclave" and scoped the statement to the Enclave binary, with a note that `enclave update` (tool images) and the in-session agent CLI refresh are separate things.
- Replaced "every so often" with `enclave version` as the concrete check against the rolling release.
- Split the four inline commands into one fenced block per install path, so nothing is a broken code span or a literal ellipsis.
- Added the two missing paths: macOS (quarantine attribute returns on re-download) and the Scoop launcher (pinned hashes, so the manifest is fetched again).
- Source updates now use `git pull && make install`, noting that `make build` only refreshes `./bin/enclave`.

Left `README.md` alone for now — happy to mirror the section there if you want it in both places.

xai (approved):

PR #91, opened 2026-09-17
feat: ask which tool to use on first run

### LINKED ISSUE #89: Ask which tool to use on first run instead of defaulting to Claude
## Summary

`enclave` currently defaults to the `claude` tool profile with no prompt:
`internal/config/default_options.go:15` sets `Tool: "claude"`, and
`internal/config/options_def.go:97` documents it as `"Tool profile (default: claude)"`.

A first run should **ask which agent to use and persist the answer**, rather than picking a
single vendor's agent silently.

## Why

- **The mechanism already exists.** #74 introduced exactly this pattern for `backend`: the
  value defaults to `auto`, `app.resolveBackend` resolves it once, asks via `prompt.Choose`
  when the choice is genuinely ambiguous, and persists with `config.WriteGlobalDefault`.
  This is a second instance of a shipped pattern, not a new concept.
- A hardcoded single-vendor default is a poor fit for an Eclipse Foundation project, and a
  first official release is when that gets noticed.
- Enclave bundles five agent profiles (`claude`, `codex`, `mistral-vibe`, `opencode`, `pi`)
  next to the two IDE attach profiles. Most users never learn that from a default that
  never asks.

## Proposal

1. Introduce an `auto`-style unset state for `tool`, mirroring `backend`.
2. The first command that starts a session or builds its image (`run`, `shell`, `continue`,
   `resume`, `exec`, and `update` without explicit tool arguments) asks once which installed
   agent to use and persists the answer to `~/.config/enclave/config.json` via
   `config.WriteGlobalDefault`.
3. **Never ask** under `--json`, `--yes`, or a non-terminal run, and never from the verbs
   that do not commit to a tool (`ps`, `status`, `stop`, `cleanup`, `tools`, `features`,
   `config`, `review-target`, …). An `update` naming explicit tools rebuilds exactly those
   images and never reads the default, so it does not ask either.
4. Offer every installed agent profile. The IDE profiles (`theia`, `theia-next`) attach a
   host IDE instead of running an agent in the terminal, so the question does not offer
   them; `--tool theia` still selects them.
5. Fall back to `claude` when the question cannot be asked, so scripts and CI are
   unaffected. A host with exactly one installed agent uses that one instead, since an
   interactive run there would not ask either and `claude` may not be installed at all.
6. `--tool` continues to override, unchanged. A configured `tool` disables the question;
   `"tool": "auto"` asks again.

## Out of scope

The question lists the installed agents and nothing else. Detecting which of them the host
can already authenticate — credential environment variables, secrets-file entries, auth
stores, a tool signed in on the host — is deliberately left out. It does not serve vendor
neutrality, and the available signals mislead often enough to be worse than no signal: a
host OAuth login never reaches the container, and a credential for one of a tool's several
providers says little about whether that tool starts ready to work.

## Acceptance

- A first interactive run with no persisted `tool` asks once and writes the choice.
- A second run does not ask.
- `--json`, `--yes` and non-TTY runs never ask and behave exactly as today.
- The question does not offer the IDE profiles.
- Docs updated: `docs/configuration.md`, `docs/cli-reference.md`, `docs/ARCHITECTURE.md`,
  `README.md` quickstart, and the website getting-started page.

xai:
#### What it does

Closes #89.

The `tool` default is no longer `claude`. It is unset (`auto`), and the first
command that starts a session or builds its image asks once which agent to run,
then saves the answer as `"tool"` in the global config — the same shape #74 gave
`backend`.

- Asked only for `run`, `shell`, `continue`, `resume`, `exec`, and a bare
  `update`. An `update` with explicit tool arguments rebuilds exactly those
  images and never reads the default tool, so it does not ask.
- Never asked under `--json`, `--yes`, or when stdin/stdout/stderr are not all
  terminals, and never by the other verbs (`ps`, `status`, `stop`, `cleanup`,
  `tools`, `features`, `config`, `review-target`, …). Those keep using `claude`,
  so scripts and CI are unaffected. An unanswered question does the same for
  that run and saves nothing.
- The menu lists every installed agent. The IDE profiles (`theia`,
  `theia-next`) are never offered; they attach a host IDE rather than running an
  agent in the terminal. `--tool theia` still selects them.
- A host with exactly one installed agent has nothing to choose between, so it
  is never asked — and the non-interactive path uses that agent too, rather than
  a `claude` that may not be installed there. An empty inventory falls back to
  `claude`.
- `--tool` and a configured `tool` skip the question entirely. `"tool": "auto"`
  asks again.

#### How to test

A scratch `HOME` keeps your real config and stores untouched. From the repo
root, after `make build`:

```bash
scratch=$(mktemp -d); app="$PWD"; bin="$PWD/bin/enclave"
run() { env -u XDG_CONFIG_HOME -u XDG_STATE_HOME -u XDG_CACHE_HOME \
  HOME="$scratch" ENCLAVE_HOME="$app" "$bin" "$@"; }
cd /some/project
```

1. **It asks once and remembers.** `run` in a terminal. The question lists
   `claude/codex/mistral-vibe/opencode/pi` — no `theia` entries — and names the
   config file the answer goes to. Answer `codex`. The run then proceeds
   normally (from an empty scratch state that means an image build; Ctrl-C is
   fine, the answer is already saved).
   `cat $scratch/.config/enclave/config.json` shows `{"tool": "codex"}`. Run
   `run` again: no question.
2. **`auto` asks again.** Write `{"tool": "auto"}` to
   `$scratch/.config/enclave/config.json` and `run` in a terminal: the question
   is back. The same config resolves to `claude` without asking when it cannot
   ask — `run config --json` reports `tool` effective as `claude`.
3. **Scripts are unaffected.** `rm $scratch/.config/enclave/config.json`, then
   `run < /dev/null`, `run ps --json`, `run tools`, `run config`. None asks,
   none writes `config.json`, and `run config` shows `tool` effective as
   `claude`.
4. **Explicit update targets do not ask.** Still without a config:
   `run update codex` goes straight to the codex rebuild, and
   `$scratch/.config/enclave/` gains no `config.json`. A bare `run update` does
   ask and saves.
5. **Overrides still win.** With `{"tool": "codex"}` saved, `run --tool pi`
   uses pi for that run and leaves the saved key at `codex`.
6. **Nothing changed for you.** Run `enclave` normally, without the scratch
   `HOME`: an existing `tool` in your own config is used and no question
   appears. Clean up with `rm -rf "$scratch"`.

#### Follow-ups

`enclave tools` marks `claude` as `(selected)` while the tool is unset, because
engine-free verbs resolve to the fallback. That is what a non-interactive run
would use, but it no longer tells the whole story; printing no marker in that
state would need the unresolved value threaded through to the listing.

#### Breaking changes

- [x] This PR introduces breaking changes and has been coordinated with maintainers.

On its first invocation without an explicit `--tool <name>`, enclave now prompts
instead of implicitly starting `claude`.

#### Review checklist


internal/app/tool.go
```diff
@@ -0,0 +1,310 @@
...
+	for _, candidate := range candidates {
+		all = append(all, candidate.Name)
+		if candidate.Credentials {
+			withCredentials = append(withCredentials, candidate.Name)
+		}
+	}
+	if len(withCredentials) > 1 {
+		return withCredentials, withCredentials
+	}
+	return all, withCredentials
```
EclipseSourceAI: The narrowing hides installed tools that cannot be typed anyway: with only `ANTHROPIC_API_KEY` exported, claude/opencode/pi all count as credentialed (all three declare `anthropic-api-key`), so codex and mistral-vibe never appear, and `prompt.Choose` only accepts listed options, so typing `codex` three times falls back to claude and saves nothing. Offering every installed agent and marking which ones already have credentials would avoid the dead end.

internal/app/tool.go
```diff
@@ -0,0 +1,310 @@
...
+			return true
+		}
+	}
+	return false
+}
+
+// hostAuthFilesExist reports whether the tool is set up on the host itself,
+// outside enclave (for example ~/.claude/.credentials.json).
+func hostAuthFilesExist(home string, profile model.Profile) bool {
+	if credentials := config.HostProfileCredentialsPath(home, profile); credentials != "" && fileExists(credentials) {
```
EclipseSourceAI: A tool signed in on the host does not make the session authenticated: the auth files are explicitly denied from host-config passthrough ([link](https://github.com/eclipse-enclave/enclave/blob/eddb9fc7941d5125c30dc440aa5ecf9dcdcb204a/internal/config/host_config_passthrough.go#L42-L51)) and only reach the store through an explicit `enclave auth import` ([link](https://github.com/eclipse-enclave/enclave/blob/eddb9fc7941d5125c30dc440aa5ecf9dcdcb204a/docs/auth.md#L81-L86)). So `~/.claude/.credentials.json` makes the question promise a sign-in-free start that the container then does not deliver.

internal/app/tool.go
```diff
@@ -0,0 +1,310 @@
...
+	}
+	entries, err := os.ReadDir(config.HostStoreAuthTreeDir(home, profile.Name))
+	if err != nil {
+		return false
+	}
+	for _, entry := range entries {
+		if !entry.IsDir() {
+			continue
+		}
+		if anyFileExists(config.HostStoreAuthDir(home, profile.Name, entry.Name()), authFiles) {
```
EclipseSourceAI: Bare file existence is not what the runtime treats as a session for opencode and pi: their providers use `json_pointer_non_null` checks on `auth.json`, evaluated in `checkProviderSession` ([link](https://github.com/eclipse-enclave/enclave/blob/eddb9fc7941d5125c30dc440aa5ecf9dcdcb204a/internal/runtime/auth_manager.go#L137-L152)). An `auth.json` holding only a github-copilot entry (or `{}` after a failed login) marks the tool as credentialed here, which contradicts the doc comment above claiming this mirrors the authSession checks.

internal/app/tool.go
```diff
@@ -0,0 +1,310 @@
...
+			continue
+		}
+		if fileExists(filepath.Join(dir, filepath.FromSlash(name))) {
+			return true
+		}
+	}
+	return false
+}
+
+func fileExists(path string) bool {
```
EclipseSourceAI: Duplicates `util.FileExists`, same body ([link](https://github.com/eclipse-enclave/enclave/blob/eddb9fc7941d5125c30dc440aa5ecf9dcdcb204a/internal/util/util.go#L306-L309)). `internal/app` already imports `internal/util` elsewhere, so use that one.

internal/app/tool.go
```diff
@@ -0,0 +1,310 @@
...
+// config, so only the first run pays for the question. An unanswered question
+// falls back to claude for this run without saving anything.
+func askForTool() string {
+	candidates := listToolCandidates()
+	options, withCredentials := promptableTools(candidates)
+	switch len(options) {
+	case 0:
+		return toolFallback
+	case 1:
+		return options[0]
```
EclipseSourceAI: With exactly one installed agent the interactive run uses it while a scripted run on the same host still falls back to `claude`, which may not even be installed. Making the non-interactive path use a single candidate too would keep both consistent.

internal/app/tool_test.go
```diff
@@ -0,0 +1,413 @@
...
+		t.Skip("could not locate the repo app root; skipping")
+	}
+	t.Setenv(model.EnvHome, root)
+	t.Setenv("HOME", t.TempDir())
+
+	var names []string
+	for _, candidate := range hostToolCandidates() {
+		names = append(names, candidate.Name)
+	}
+	want := "claude,codex,mistral-vibe,opencode,pi"
```
EclipseSourceAI: Hardcoding the full bundled list makes this fail on every added or removed tool extension for reasons unrelated to the question. Asserting that the agent profiles are present and the theia ones are not would be enough.

internal/app/tool_test.go
```diff
@@ -0,0 +1,413 @@
...
+	if got := strings.Join(names, ","); got != want {
+		t.Fatalf("candidates = %q, want %q", got, want)
+	}
+}
+
+func findRepoAppRoot(t *testing.T) (string, bool) {
```
EclipseSourceAI: Tests in this package locate the repo root with `filepath.Join("..", "..")` ([link](https://github.com/eclipse-enclave/enclave/blob/eddb9fc7941d5125c30dc440aa5ecf9dcdcb204a/internal/app/dockerfile_gen_test.go#L215)), which drops this 20-line upward walk.

website/docs/docs/getting-started.md
EclipseSourceAI, line 1: "Start your first session" still says `enclave` launches the default agent, but that is exactly the run that now asks the question. The note belongs there, not only under "Pick a specific agent" further down.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

## What this PR does

The `tool` default moves from `claude` to an unset `auto`, and `app.Run` resolves it once (`resolveTool`) before anything reads the tool name, re-layering options with `config.ResolveOptionsForTool` when the answer differs. Only the session-starting verbs (`run`, `shell`, `continue`, `resume`, `exec`, and a bare `update`) may ask; everything else, plus non-terminal/`--json`/`--yes` invocations, keeps `claude`. The answer is written to the global config with the same `WriteGlobalDefault` seam #74 used for `backend`, and the backend prompt helpers are renamed to shared names (`promptUsable`, `promptAllowed`). Docs are updated consistently across README, `docs/`, and the website.

The wiring is clean and fits the existing option-resolution flow well, the seam/test structure mirrors `backend.go`, and `go test ./...` passes here.

## Where to focus

The interesting part is `toolHasCredentials` and the menu narrowing, which is where most of the new code sits:

- Narrowing to credentialed tools hides installed agents that `prompt.Choose` then refuses to accept, and the dead end falls back to `claude`. A single exported `ANTHROPIC_API_KEY` is enough to trigger it.
- The "signed in on the host" signal does not actually mean the session can authenticate: host auth files are deliberately blocked from passthrough and only reach the store via `enclave auth import`.
- The auth-store check treats any existing `auth.json` as a session, which is weaker than the `authSession` checks opencode and pi declare.

Whether the credential probing is worth its complexity at all (versus offering every installed agent) is a maintainer call. Smaller items are in the inline comments.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Follow-up review. Since my last pass, the credential-probing subsystem (`toolCandidate`, `promptableTools`, `toolHasCredentials`, `authStoreHasSession`, `hostAuthFilesExist`, `fileExists`) was removed entirely and replaced with a much simpler model: the menu just offers every installed agent, and a host with exactly one installed agent uses it without asking on both the interactive and non-interactive paths (`toolWithoutAsking`). Tests and docs were updated to match. `go build ./...` and `go test ./internal/app/...` pass, `gofmt` is clean.

This resolves every open thread from my earlier review, all of them for the reasons above. No new issues found in the changes since then.

- [credential narrowing hid installed tools with a dead-end fallback; the whole narrowing mechanism was removed](https://github.com/eclipse-enclave/enclave/pull/91#discussion_r4039256823)
- ["signed in on the host" did not mean the session could authenticate; that check was removed](https://github.com/eclipse-enclave/enclave/pull/91#discussion_r4039256919)
- [bare `auth.json` existence contradicted the documented authSession checks; that check was removed](https://github.com/eclipse-enclave/enclave/pull/91#discussion_r4039257095)
- [duplicate of `util.FileExists`; the local copy is gone along with the code that used it](https://github.com/eclipse-enclave/enclave/pull/91#discussion_r4039257266)
- [single-installed-agent case now behaves the same on interactive and scripted runs](https://github.com/eclipse-enclave/enclave/pull/91#discussion_r4039257408)
- [test no longer hardcodes the full bundled tool list, only asserts agent vs IDE membership](https://github.com/eclipse-enclave/enclave/pull/91#discussion_r4039257565)
- [test now locates the repo root with the same `filepath.Join("..", "..")` convention used elsewhere in the package](https://github.com/eclipse-enclave/enclave/pull/91#discussion_r4039257785)
- [getting-started.md now mentions the first-run question where it first tells users to just run `enclave`](https://github.com/eclipse-enclave/enclave/pull/91#discussion_r4039257936)

xai:
@EclipseSourceAI

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Second follow-up. Since my last review (95ebcc3), the only changes are: docs (README, website/docs/configuration.md) updated to spell out that non-interactive runs fall back to the sole installed agent when there's exactly one, and two template-y comments removed from `tool_test.go` above the test funcs. Both match the current `toolWithoutAsking` behavior and the second-review state, no functional changes.

All threads from my previous reviews were already resolved before this pass, nothing new to raise.

docs/cli-reference.md
```diff
@@ -286,6 +286,14 @@ Mutation commands (`add-domain`, `remove-domain`, `set-mode`) apply the new poli
 
 Persistent defaults can be set in `~/.config/enclave/config.json` (global) or `~/.config/enclave/projects/<hash>/config.json` (per-project). See [Configuration](configuration.md).
 
+## Tool selection
+
+The tool is unset by default. The first command that starts a session or builds its image (`run`, `shell`, `continue`, `resume`, `exec`, and `update` without explicit tool arguments) asks once which coding agent to use and saves the answer as `"tool"` in `~/.config/enclave/config.json`; later runs read it from there and never ask again. The question is only asked when stdin, stdout, and stderr are terminals and no `--json` or `--yes` was given; scripts, captured output, JSON consumers, and every other verb (`ps`, `status`, `stop`, `cleanup`, `tools`, `features`, `config`, `review-target`, …) fall back to `claude` silently, exactly as before — unless exactly one agent is installed, which is the tool an interactive run would use without asking too. An unanswered question falls back for that run without saving anything.
```
tortmayr: I have two concerns with the documented behavior:

First, silently falling back to Claude seems surprising now that the default tool is intentionally unset. For commands that actually require a tool, would it be clearer to fail with an actionable message asking
  the user to pass --tool or configure "tool"?  I guess this was added for backwards compatibility but I would rather see a minor breaking change here instead of this behavior.
Also as far as I can see from the code,  commands that do not need the --tool option at all still resolve it.
e.g. enclave ps or status. Why?

Second, when can the “exactly one agent is installed” case occur? As far as I can tell, tool discovery includes the built-in profiles, built-ins cannot currently be removed or disabled.
That makes this branch unreachable in a normal installation unless someone uses a custom, pruned ENCLAVE_HOME.

tortmayr (commented):
In general code looks good to me and the change seems to work as expected.
I have one inline comment/question regarding the documented behavior.

docs/cli-reference.md
xai: I agree, we should not try to be more backwards-compatible than necessary and raise an error instead of doing a weird fallback. I'll change it to fail hard in a follow-up commit.

Regarding the "exactly oone agent installed" situation, it probably only happens when someone uses a custom ENCLAVE_HOME. We probably can ignore a such specific corner case. I'll remove that specific handling.

xai:
@EclipseSourceAI

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Third follow-up review. Since my last pass (af63932), one commit landed (3c4d5ee), and it's a direct response to tortmayr's inline feedback on this PR: an unresolved tool now fails naming `--tool` and the `tool` key instead of silently falling back to `claude`, actions that never read the tool (`ps`, `status`, `tools`, `config`, `attach`, …) no longer resolve it at all via the new `actionNeedsTool`/`toolFreeActions` split in `internal/app/actions.go`, and the unreachable single-installed-agent shortcut is gone.

I checked `actionNeedsTool` against every CLI action string in the codebase (network-status/print/diff/apply/add-domain/remove-domain/set-mode, auth-import/export, devcontainer-generate, img-import, info, user-command, …) and each one's tool-need matches whether its handler actually reads `opts.Tool` for image/store keys. `go build ./...`, `go test ./internal/app/...`, and `gofmt` are all clean. No new issues found.

All threads from my earlier reviews were already reported resolved in my previous pass.

tortmayr (approved):
LGTM! 👍🏼

PR #93, opened 2026-09-17
fix: pre-create the tool skills directory in the config store

### LINKED ISSUE #92: Nested tool skillsDir leaves its parent directory root-owned in the config store
### Bug Description:

A tool's skills directory is bind-mounted read-only into the config store
(`addSkillMounts`, `internal/runtime/skills_mount.go:81`). The store layout
prepared ahead of the session only covers the parents of `settingsTarget` and
the declared auth files (`configVolumeRelativeDirs`,
`internal/runtime/volume_manager.go:149`), so when `skillsDir` sits in a
subtree none of those touch, the container runtime creates the mount point and
its missing parents itself — as root.

The tool then cannot write anything beside its skills directory. With a spec
like

```yaml
configDir: .tool
skillsDir: .tool/shared/skills
settingsTarget: .tool/cli/settings.json
```

`~/.tool/shared` is created root-owned 0755, and the tool fails on its first
write below it:

```
mkdir /home/agent/.tool/shared/projects: permission denied
```

on the host:

```
$ ls -la ~/.local/state/enclave/projects/<hash>/<tool>/config-store/default/
drwx------ user   user    cli      <- prepared by enclave
drwxr-xr-x 100000 100000  shared   <- created by the container runtime
```

This is not engine-specific: rootless podman with `--userns=keep-id` shows the
first subuid, a rootful engine shows uid 0, and in both cases the container
user cannot write the directory. A `skillsDir` one level below `configDir`
(`.tool/skills`) is unaffected, because its parent is the store root, which
enclave creates itself, and everything the tool writes beside `skills/` lands
in that host-owned root.

Fix: derive a layout dir from `skillsDir` as well, so the mount point and its
parents are created host-side with the host user's ownership.

### Steps to Reproduce:

1. Install a tool extension whose `skillsDir` is nested below `configDir` in a
   different subtree than `settingsTarget` and the auth files (see the spec
   above).
2. Start a session for that tool.
3. Have the tool write next to its skills directory — it fails with
   `permission denied`, and the directory is root-owned in the config store on
   the host.

xai:
#### What it does

When a tool configures a skills directory within its config directory but outside the settings subtree (such as `.tool/shared/skills` alongside `.tool/cli/settings.json`), the container runtime created the missing mount target and intermediate directories as root. This prevented the tool from creating sibling directories such as `.tool/shared/projects` due to permission errors.

This change adds the relative skills directory to the layout directories in `volumeManager.configVolumeRelativeDirs()`. Store preparation then creates the directory structure with host user ownership before the container starts.

Closes #92

#### How to test

Run the unit tests covering the layout configuration:

```bash
go test -v ./internal/runtime -run TestConfigVolumeRelativeDirs
```

Run test and lint targets:

```bash
make test
make lint
```

#### Follow-ups

None.

#### Breaking changes

#### Review checklist


internal/runtime/volume_manager.go
```diff
@@ -148,21 +148,32 @@ func (m volumeManager) authSyncSpec(stores storeSet) *backend.AuthSyncSpec {
...
 
 	if settingsPath, err := m.settingsRelativePath(); err == nil {
 		addDirForFile(settingsPath)
 	}
+	// The skills directory is a read-only mount target inside the config store.
+	// Creating it here keeps it, and every parent it needs, owned by the host
+	// user: left to the container runtime, the mount point and its missing
+	// parents are created as root, and a tool that writes siblings of the
+	// skills directory can no longer do so.
+	if skillsPath, err := m.relativePathWithinConfig(m.profile.SkillsDir); err == nil {
```
EclipseSourceAI: `MkdirAll` in [`prepareConfigStoreLayout`](https://github.com/eclipse-enclave/enclave/blob/eddb9fc7941d5125c30dc440aa5ecf9dcdcb204a/internal/backend/docker/prepare.go#L120-L134) is a no-op when the directory already exists, so a persistent store that already picked up a root-owned `.gemini/config` from an earlier run stays broken after this change. Anyone currently hitting #92 needs to wipe the store by hand (they cannot `rm -rf` it without sudo either), so that should at least be called out in the issue or the release notes.

internal/runtime/volume_manager.go
```diff
@@ -148,21 +148,32 @@ func (m volumeManager) authSyncSpec(stores storeSet) *backend.AuthSyncSpec {
...
+		addDir(skillsPath)
```
EclipseSourceAI: Skills is not the only mount that lands inside the config store: the memory dir ([here](https://github.com/eclipse-enclave/enclave/blob/eddb9fc7941d5125c30dc440aa5ecf9dcdcb204a/internal/runtime/runtime.go#L1268-L1269)) and the IDE bridge dir ([here](https://github.com/eclipse-enclave/enclave/blob/eddb9fc7941d5125c30dc440aa5ecf9dcdcb204a/internal/runtime/ide_bridge.go#L73-L74)) hit the same root-created-target problem for any profile that nests them deeper than one level. `prepareMounts` already runs before `prepareVolumes`, so the layout dirs could be derived from the accumulated container mount targets inside the config dir instead of hardcoding `SkillsDir`. A maintainer should decide whether that generalization belongs here or in a follow-up.

internal/runtime/volume_manager.go
```diff
@@ -148,21 +148,32 @@ func (m volumeManager) authSyncSpec(stores storeSet) *backend.AuthSyncSpec {
...
 			return
 		}
-		dirs[dir] = struct{}{}
+		addDir(filepath.Dir(relPath))
 	}
 
 	if settingsPath, err := m.settingsRelativePath(); err == nil {
 		addDirForFile(settingsPath)
 	}
+	// The skills directory is a read-only mount target inside the config store.
```
EclipseSourceAI: The skills dir is only a mount target when the config source does not handle skills ([runtime.go](https://github.com/eclipse-enclave/enclave/blob/eddb9fc7941d5125c30dc440aa5ecf9dcdcb204a/internal/runtime/runtime.go#L358-L362)); in the config-source case `addSkillMounts` never runs and this layout entry is wiped again by the overlay. The comment reads as if it always applies.

internal/runtime/volume_manager_test.go
```diff
@@ -43,6 +44,34 @@ func TestConfigVolumeRelativeDirsIncludesSettingsAndAuthParents(t *testing.T) {
...
+		t.Fatalf("configVolumeRelativeDirs() len = %d, want %d (%v)", len(got), len(want), got)
+	}
+	for i := range want {
+		if got[i] != want[i] {
+			t.Fatalf("configVolumeRelativeDirs()[%d] = %q, want %q", i, got[i], want[i])
+		}
+	}
+}
+
 func TestConfigVolumeRelativeDirsSkipsRootLevelFiles(t *testing.T) {
```
EclipseSourceAI: The shipped claude spec sets [`skillsDir: .claude/skills`](https://github.com/eclipse-enclave/enclave/blob/eddb9fc7941d5125c30dc440aa5ecf9dcdcb204a/extensions/tools/claude/spec.yaml#L18), so this fixture no longer matches the real profile and the new `skills` entry for the most-used tool stays untested. Setting `SkillsDir` here and expecting `["skills"]` would cover it.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Small, focused fix: `volumeManager.configVolumeRelativeDirs()` now also emits the config-relative skills directory, so store preparation creates it (and its intermediate parents) as the host user before the container starts, instead of letting Docker create the bind-mount target as root. The `addDir`/`addDirForFile` split is behavior preserving, and the mechanism is the right one (the layout dirs already exist for settings and auth parents, and `prepareConfigStoreLayout` creates them host-owned).

Points worth a closer look:

- Existing persistent stores that already contain a root-owned intermediate dir are not repaired, since `MkdirAll` succeeds on an existing directory. Users currently affected by #92 stay broken after upgrading.
- The skills dir is only a real mount target when the config source does not handle skills. In the config-source path the new layout entry is dead weight, and the new comment does not reflect that.
- Memory dir and IDE bridge dir are the same class of config-store-internal mount targets and are not covered. Since `prepareMounts` runs before `prepareVolumes`, deriving layout dirs from the accumulated mount targets would generalize this.
- Test coverage uses a synthetic antigravity profile. No shipped profile exercises the new entry, and the claude fixture omits the `skillsDir` the real spec sets.

`go build ./...` and `go test ./internal/runtime -run TestConfigVolumeRelativeDirs` pass locally.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Since my last review, the only change is 6020fd3, which reworded the `configVolumeRelativeDirs` comment to say the skills dir is a mount target "when the config source does not handle skills," and added a test using the real Claude profile's `skillsDir: .claude/skills` expecting `["skills"]`.

- [comment on the skills-dir explanation reading as unconditional](https://github.com/eclipse-enclave/enclave/pull/93#discussion_r4040128859)
- [claude fixture not exercising the shipped skillsDir](https://github.com/eclipse-enclave/enclave/pull/93#discussion_r4040129068)

The other two threads (root-owned dirs from earlier runs not getting repaired, and memory/IDE-bridge dirs having the same mount-target problem) are still open. No changes touched `prepare.go`, `runtime.go`, or `ide_bridge.go` in this update, so those still need a maintainer decision.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Two commits landed since my last pass: a2eff48 adds `MemoryDir` to `configVolumeRelativeDirs` (mirroring the skills handling, with an updated comment and a new test), and 9b7c9c4 adds a real fix for the "existing root-owned dirs aren't repaired" concern: `prepareConfigStoreLayout` now walks each layout dir after `MkdirAll` and reports the outermost path component not owned by the host user, with a cleanup hint (`internal/backend/docker/prepare.go`, `internal/util/owner_unix.go`/`owner_windows.go`). Docs in `docs/runtime/stores.md` were updated to match. Both changes are covered by new unit tests, `go build ./...` and the touched packages' tests pass.

The IDE bridge mount (`internal/runtime/ide_bridge.go`) is still not covered: its target is a sibling directory under the config store root, not on the path to any declared layout dir, so the new ownership-warning walk won't catch a root-owned `.claude/ide` either. That part of my earlier "memory/IDE bridge" comment is still open and needs a maintainer call on whether it's in scope here.

- [existing root-owned store dirs from #92 silently staying broken](https://github.com/eclipse-enclave/enclave/pull/93#discussion_r4040128450)

internal/runtime/volume_manager.go
xai: The IDE bridge mount is intentionally not covered. Its target is always exactly one level below the config store root (.claude/ide) and read-only, so a root-created directory there never blocks the tool from writing siblings, and the config-source overlay removes it on the next run anyway. Nothing observable changes for users.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

One commit landed since my last pass: 9edb8a8 ("fix: restore config layout after overlay").

It fixes a real gap, not just cleanup: `overlayConfigStore` wipes the config store and repopulates it from the generated source (`overlayConfigDir` → `clearDirContents` + `copyDirContents`), and skills/memory dirs aren't in `ConfigSourcePreservePaths` (only auth files, `HostCredentialsFile`, and the backstop list are), so for config-source tools those layout dirs were being wiped again on every session and never recreated host-owned. `overlayConfigStore` now calls `prepareConfigStoreLayout` again after the overlay, with a new test (`TestPrepareStoresRecreatesLayoutAfterConfigOverlay`) covering it. The warning text also changed from suggesting `sudo rm -rf` to `sudo chown -R`, which is the safer fix since it doesn't ask users to delete session data. The `volume_manager.go` `addDirForFile` simplification is behavior preserving (the removed early-return was redundant given `addDir`'s own cleanup) and matches the `slices.Equal` test tidy-up. `go test ./internal/backend/docker/... ./internal/runtime/...` passes. No new issues found in this commit.

- [memory dir now pre-created like skills, and the IDE bridge follow-up was addressed by your reply on the design intent](https://github.com/eclipse-enclave/enclave/pull/93#discussion_r4040128655)

planger (approved):
Thanks, looks good to me! :+1:

Issue #87, opened 2026-09-15
Allow "external" extensions to contribute commands


sdirix:
### Feature Description:

At the moment external extensions (e.g. from https://github.com/eclipse-enclave/enclave-extensions) can contribute tools and features. But it would also be nice if commands could also be contributed.

I stumbled over this when preparing vnc-viewer to be an "external" extension. This is currently not possible as I can't contribute an additional command.

To discuss: Host commands are executed on the host, so a "malicious" extension could easily infect any host which installs it.

sdirix:
@planger Wdyt?

planger:
Agreed that'd be pretty powerful. I think we could show a proper indication when installing it (as we already do when showing the extension capabilities) to at least make the risk explicit. @xai what do you think?

xai:
Yes, I like the idea as well. A warning with maybe a confirmation dialog on command install sounds fine to me.

PR #95, opened 2026-09-18
feat: let extensions contribute host commands

sdirix:
#### What it does

An installed extension's commands/host/ executables become `enclave <name>` verbs, so an extension can ship something that has to run outside the sandbox, like a VNC viewer. Names resolve after the user's own commands/ trees, so an install never takes over a name already in use.

These are the only thing an extension ships that executes outside a container. The capability summary leads with them, the diff an update renders repeats the risk, and the result envelope and `list` carry them under --json, where there is no narration to read.

commands/ is excluded from the build context and the image identity hash like the provenance sidecar, but the tree hash still covers it, so the rebuild hint now asks whether a change reached the image at all.

Implements #87

#### How to test

Install the `vnc` feature of https://github.com/sdirix/enclave-extensions which makes use of `hostCommand` to offer `enclave vnc-viewer` command.

```
enclave features add sdirix/enclave-extensions
```

#### Breaking changes

#### Review checklist


internal/usercmd/usercmd.go
```diff
@@ -84,16 +129,76 @@ func Discover(home string) ([]Command, []string) {
...
+		entries, err := os.ReadDir(kindDir)
+		if err != nil {
+			if !errors.Is(err, fs.ErrNotExist) {
+				warnings = append(warnings, fmt.Sprintf(
+					"cannot read extension directory %q: %v", kindDir, err))
+			}
+			continue
+		}
+		for _, entry := range entries {
+			if !entry.IsDir() {
```
EclipseSourceAI, line 160: `entry.IsDir()` alone picks up the installer's own dot-prefixed staging directories. I ran this against a `.replaced-vnc-1234-1/commands/host/vnc-viewer` and it registered as `enclave vnc-viewer` with `Extension: ".replaced-vnc-1234-1"`, no warning. An interrupted update leaves that directory around for up to `staleAge` (24h), and `sweepStale` explicitly assumes it is inert until swept ([stage.go](https://github.com/eclipse-enclave/enclave/blob/3e28d1275e6ea71e4221c6d2b8a5e3c39e2688d1/internal/extinstall/stage.go#L70-L72)), so the old version of an extension keeps executing host code while `list` and `remove` cannot see it. Use the same rule as [`config.isExtensionDir`](https://github.com/eclipse-enclave/enclave/blob/3e28d1275e6ea71e4221c6d2b8a5e3c39e2688d1/internal/config/extension.go#L357-L365) (exporting it would be better than a second copy of the dot check).

internal/extinstall/hostcommands.go
```diff
@@ -0,0 +1,51 @@
...
+	"io/fs"
+	"os"
+	"path/filepath"
+
+	"enclave/internal/model"
+)
+
+// hostCommandNames lists the enclave verbs extDir contributes: the executable
+// files directly inside its commands/host/, in os.ReadDir's sorted order. This
+// is the single definition of what counts, shared by the capability summary,
```
EclipseSourceAI, line 21: It is not the single definition. [`usercmd.scanDir`](https://github.com/eclipse-enclave/enclave/blob/3e28d1275e6ea71e4221c6d2b8a5e3c39e2688d1/internal/usercmd/usercmd.go#L89-L145) independently implements the same filter (regular file, follow symlink, any exec bit), and it is the one that decides what actually registers. Two copies with a comment claiming they are one will drift. `usercmd` imports only `config` and `model`, so an exported helper there that `extinstall` calls would be acyclic.

internal/extinstall/capabilities_render.go
```diff
@@ -23,6 +23,15 @@ func (c capabilities) render(w io.Writer, style Style, source string) {
...
 	row("source", source)
 
+	// First after the source: everything else an extension ships is bounded by
+	// the sandbox, and this is not. The row spells out where the verbs run
+	// instead of leaving the reader to infer it from the label.
+	hostCommands := strings.Join(c.HostCommands, ", ")
+	if hostCommands != "" {
+		hostCommands += fmt.Sprintf(" (added as `%s <name>`, run on your host outside the sandbox with your user's privileges)", model.AppName)
+	}
+	row("host commands", hostCommands)
```
EclipseSourceAI, line 33: The row promises `added as \`enclave <name>\``, but a name colliding with a built-in verb is dropped at registration with a warning ([parse.go](https://github.com/eclipse-enclave/enclave/blob/3e28d1275e6ea71e4221c6d2b8a5e3c39e2688d1/internal/cli/parse.go#L177-L194)), and so is one the user already owns. Same for `remove`'s "host command(s) no longer resolve" note, which names verbs that may never have resolved. Catching a built-in collision in `validateCommandsDir` as a warning would be the useful place, since that is the install-time decision point.

docs/ARCHITECTURE.md
```diff
@@ -179,6 +179,7 @@ The restricted network request flow has a separate
...
...e deterministically (tools before features, alphabetical within a kind) and every loser is warned about. A non-executable file in an extension's tree is ignored silently rather than warned about, unlike one in the user's own tree, because it belongs to the extension and the next update would undo a `chmod`. The directory is host-side alone: `overlayNamedExtensionDirs` skips it, so it reaches neither the build context nor the image identity hash and editing a command forces no rebuild. `TreeHash` still covers it, so a hand-edited command counts as a local modification, and `changesReachImage` keeps an update confined to it from promising a rebuild. `extinstall.hostCommandNames` is the single definition of which files count, shared by the capability summary, `Inventory`, and `remove`. The commands are reported as the capability summary's first row (they are the only capability not bounded by the sandbox), repeated on the diff an update renders, and carried in the `hostCommands` field of both `list --json` and the add/update/remove result envelope. `config.validateCommandsDir` warns about anything under `commands/` other than `host/`, which would otherwise be inert in every direction.
```
EclipseSourceAI, line 182: This bullet is roughly 300 words in one paragraph and repeats most of the new "Host commands" section in `docs/extensions/README.md`. The concepts list around it is 2-4 sentences per entry. Trim to the architectural facts (third scan source, precedence, excluded from build context and image hash but covered by `TreeHash`) and link the extensions doc for the rest.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Adds a third source for `enclave <verb>`: an installed extension's `commands/host/` executables. `usercmd.Discover` scans the user extension root after the user's own `host/`/`session/` trees, so an install can never take a name already in use, and every loser gets a warning. `commands/` is excluded from the build context and the image identity hash (`overlayNamedExtensionDirs`), while `TreeHash` still covers it, with `changesReachImage` keeping an update confined to that tree from promising a rebuild. The capability surface is extended consistently: a leading "host commands" row in the summary, a diff entry that repeats the risk on update, `hostCommands` in the `list --json` and add/update/remove envelopes, and a text `[host: ...]` suffix.

The approach fits the existing extension machinery well and the test coverage is solid. Points to focus on:

- `scanExtensions` filters extension directories with `entry.IsDir()` only, which pulls in the installer's dot-prefixed `.replaced-*` leftovers. I reproduced it: a leftover from an interrupted update contributes a working host verb that `list` and `remove` cannot see. This is the one I would fix before merge.
- `extinstall.hostCommandNames` duplicates the file filter that `usercmd.scanDir` already implements, and the doc comment claims it is the single definition.
- The summary and `remove` report verbs that may never register (built-in or user-command collisions).

Also a docs-size nit on the new `ARCHITECTURE.md` bullet.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Follow-up review of commit b0e18ea, which addresses all four points from my last pass.

- `.replaced-*` staging dirs are excluded from `scanExtensions` via `config.IsExtensionDir` (shared with `validate_extensions.go`), with a regression test (`TestDiscoverIgnoresInstallerStagingDirs`).
- `extinstall.hostCommandNames` is gone; `usercmd.ExtensionCommandNames` is now the single filter used by capabilities, `Inventory`, and (indirectly, via `Resolving`) remove.
- Names shadowed by a built-in or the user's own command are no longer promised: `usercmd.Shadowed` backs a `ShadowedHostCommands` map that both the capability summary/diff (`addedHostCommands()`) and the remove note (`usercmd.Resolving`) respect, each covered by new tests.
- The `ARCHITECTURE.md` bullet is trimmed to the architectural facts and links to `extensions/README.md#host-commands`.

Verified `go build ./...` and `go test ./internal/usercmd/... ./internal/extinstall/... ./internal/cli/...` all pass on the merged tree. Didn't find anything new to flag in this commit; scope is tight and matches its stated purpose.

- [installer staging dirs (`.replaced-*`) no longer register as host commands](https://github.com/eclipse-enclave/enclave/pull/95#discussion_r4048218494)
- [the host-command file filter now has a single shared definition instead of a second copy](https://github.com/eclipse-enclave/enclave/pull/95#discussion_r4048218687)
- [the summary/diff and remove note no longer promise verbs shadowed by a built-in or user command](https://github.com/eclipse-enclave/enclave/pull/95#discussion_r4048218841)
- [the ARCHITECTURE.md bullet is trimmed and links to the extensions doc instead of duplicating it](https://github.com/eclipse-enclave/enclave/pull/95#discussion_r4048218997)

PR #96, opened 2026-09-19
fix: persist OpenCode runtime state

xai:
#### What it does

Persists OpenCode's data and state directories in the project config store so runtime state, including the selected model, survives container restarts. Runtime files are excluded from host config passthrough.

#### How to test

1. Build Enclave with `make build`.
2. From a test project, run `./bin/enclave --tool opencode --rebuild` and select a non-default model in OpenCode.
3. Exit OpenCode, then run `./bin/enclave --tool opencode` from the same project.
4. Confirm that OpenCode starts with the model selected in step 2.

#### Follow-ups

None.

#### Breaking changes

#### Review checklist


extensions/tools/opencode/spec.yaml
```diff
@@ -15,6 +15,17 @@ sandbox:
...
+    - storage/
+    - plans/
+    - snapshot/
+    - worktree/
+    - repos/
+    - tool-output/
+    - log/
+    - mcp-auth.json
+    - opencode*.db*
+    - state/
```
EclipseSourceAI: Ten patterns with no explanation of where they came from. The codex spec groups its entries with a comment saying what each group is and why it is safe to pin ([codex spec.yaml L23-33](https://github.com/eclipse-enclave/enclave/blob/3e28d1275e6ea71e4221c6d2b8a5e3c39e2688d1/extensions/tools/codex/spec.yaml#L23-L33)). Same here, otherwise nobody can tell later which of these OpenCode still writes.

Also worth spelling out in that comment: directory patterns are anchored at the config-store root ([HostConfigPathMatches L242-245](https://github.com/eclipse-enclave/enclave/blob/3e28d1275e6ea71e4221c6d2b8a5e3c39e2688d1/internal/config/host_config_passthrough.go#L242-L245)), so anything OpenCode nests one level deeper (under a per-project subdir, say) is not covered by these.

extensions/tools/opencode/spec.yaml
```diff
@@ -15,6 +15,17 @@ sandbox:
...
   configDir: .config/opencode
   skillsDir: .config/opencode/skills
+  statePaths:
+    - storage/
+    - plans/
+    - snapshot/
+    - worktree/
+    - repos/
+    - tool-output/
+    - log/
```
EclipseSourceAI: `log/` and `tool-output/` are not state worth pinning, and `statePaths` also means "preserve across every overlay", so they just accumulate in the config store. The generic backstop already denies `logs/` for every tool ([host_config_passthrough.go L21-36](https://github.com/eclipse-enclave/enclave/blob/3e28d1275e6ea71e4221c6d2b8a5e3c39e2688d1/internal/config/host_config_passthrough.go#L21-L36)); `log/` looks like it belongs there instead.

extensions/tools/opencode/spec.yaml
```diff
@@ -15,6 +15,17 @@ sandbox:
...
+    - mcp-auth.json
```
EclipseSourceAI, line 30: `mcp-auth.json` holds MCP OAuth tokens. Pinning it via `statePaths` keeps it in the per-project config store, so every project re-authenticates. Should this be a provider `authFiles` entry so it lands in the shared auth store like `auth.json`?

extensions/tools/opencode/entrypoint.d/setup.sh
```diff
@@ -9,32 +9,48 @@
 # OpenCode extension setup
 config_dir="$HOME/.config/opencode"
 data_dir="$HOME/.local/share/opencode"
+state_dir="$HOME/.local/state/opencode"
+state_store_dir="$config_dir/state"
```
EclipseSourceAI: `data_dir` links to `$config_dir` itself, so `$config_dir/state` is also reachable as `$HOME/.local/share/opencode/state`. If OpenCode ever writes a `state` entry in its data dir the two namespaces silently merge. A less generic store name (`xdg-state/`) would avoid that.

extensions/tools/opencode/README.md
```diff
@@ -8,6 +8,8 @@ LLM providers (OpenAI, Anthropic, Google).
 - **Command**: `opencode`
 - **YOLO flag**: *(none — runs in autonomous mode by default)*
 - **Config directory**: `~/.config/opencode`
+- **Data directory**: `~/.local/share/opencode` (persisted through the config store)
+- **State directory**: `~/.local/state/opencode` (persisted under the config store's `state/` directory)
```
EclipseSourceAI: `docs/runtime/stores.md` is where `statePaths` is documented and it currently names codex as the only tool doing this ([stores.md L34-39](https://github.com/eclipse-enclave/enclave/blob/3e28d1275e6ea71e4221c6d2b8a5e3c39e2688d1/docs/runtime/stores.md#L34-L39)). The new synthetic `state/` directory in the config store is user-visible, so it should be mentioned there too.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

## What this PR does

Makes OpenCode's XDG state survive container restarts. `entrypoint.d/setup.sh` generalizes the existing data-dir migration into a reusable `opencode_link_into_store` helper and applies it twice: `~/.local/share/opencode` -> `~/.config/opencode` (unchanged behavior) and `~/.local/state/opencode` -> `~/.config/opencode/state` (new). The config dir is the bind-mounted config store, so both now persist. `spec.yaml` gains a `statePaths` list so the config-source overlay preserves that runtime state and host config passthrough cannot overwrite it.

The approach fits the codebase: `statePaths` already exists and feeds both `ConfigSourcePreservePaths` and the passthrough deny-list, and the symlink-into-store pattern is what OpenCode already did for its data dir. Tests cover the entrypoint end to end and mirror the existing codex passthrough test. Shellcheck is clean and the touched packages pass.

## Where to focus

The `statePaths` list is the part worth scrutiny. Ten patterns arrive with no justification, unlike the codex spec which documents each group, and they pull double duty: deny from host passthrough *and* pin forever in the config store. `log/` and `tool-output/` fall on the wrong side of that, and `mcp-auth.json` is arguably shared-auth-store material rather than per-project state. A maintainer who has run OpenCode in the container should confirm the paths are the real top-level names, since directory patterns are anchored at the store root.

Smaller points inline: the `state/` store name collides with the data-dir namespace, and `docs/runtime/stores.md` (the canonical `statePaths` doc) was not updated.

extensions/tools/opencode/spec.yaml
xai: i think this would be better done as a follow-up instead of mixing it into a mere persistence fix.

xai:
@EclipseSourceAI

extensions/tools/opencode/entrypoint.d/setup.sh
```diff
@@ -9,32 +9,48 @@
 # OpenCode extension setup
 config_dir="$HOME/.config/opencode"
 data_dir="$HOME/.local/share/opencode"
+state_dir="$HOME/.local/state/opencode"
```
EclipseSourceAI: OpenCode resolves its roots through `xdg-basedir` ([global.ts L3-L14](https://github.com/sst/opencode/blob/fee476bb90043a1012abda156dd9af9e5c71b19d/packages/core/src/global.ts#L3-L14)), so an image that exports `XDG_STATE_HOME` leaves this symlink unused and persistence silently stops. `entrypoint.sh` already assumes those vars can be set (it rewrites `XDG_*_HOME` for devcontainers, [L422-L425](https://github.com/eclipse-enclave/enclave/blob/3e28d1275e6ea71e4221c6d2b8a5e3c39e2688d1/entrypoint.sh#L422-L425), and uses `${XDG_CONFIG_HOME:-$HOME/.config}` at [L557](https://github.com/eclipse-enclave/enclave/blob/3e28d1275e6ea71e4221c6d2b8a5e3c39e2688d1/entrypoint.sh#L557)). `${XDG_STATE_HOME:-$HOME/.local/state}/opencode` would be safer.

extensions/tools/opencode/spec.yaml
```diff
@@ -15,6 +15,21 @@ sandbox:
...
+    - snapshot/
+    - worktree/
+    - repos/
+    - opencode*.db*
+    # Full truncated tool output remains addressable from session history.
+    - tool-output/
+    # MCP OAuth remains project-scoped with the rest of the tool data.
+    - mcp-auth.json
+    # The separate XDG state root is redirected here by setup.sh.
+    - xdg-state/
```
EclipseSourceAI, line 32: Pinning the whole state root drags in more than TUI state: upstream also writes `locks/`, `server.json` (a dead pid/url once the container exits) and `password`, a generated daemon secret, into that directory ([daemon.ts L39-L41](https://github.com/sst/opencode/blob/fee476bb90043a1012abda156dd9af9e5c71b19d/packages/cli/src/services/daemon.ts#L39-L41)). The model selection this PR is after is just `model.json` ([variant.shared.ts L19](https://github.com/sst/opencode/blob/fee476bb90043a1012abda156dd9af9e5c71b19d/packages/opencode/src/cli/cmd/run/variant.shared.ts#L19)), so `xdg-state/model.json` plus `xdg-state/plugin-meta.json` would keep the daemon credential out of the persisted store and let the overlay clear stale registrations.

extensions/tools/opencode/README.md
```diff
@@ -8,6 +8,8 @@ LLM providers (OpenAI, Anthropic, Google).
 - **Command**: `opencode`
 - **YOLO flag**: *(none — runs in autonomous mode by default)*
 - **Config directory**: `~/.config/opencode`
+- **Data directory**: `~/.local/share/opencode` (persisted through the config store)
+- **State directory**: `~/.local/state/opencode` (persisted under the config store's `xdg-state/` directory)
```
EclipseSourceAI: Follow-up material, not this PR: OpenCode's third root `~/.cache/opencode` holds `bin/`, where it downloads ripgrep and LSP servers (gopls, eslint, rubocop) at runtime ([global.ts L22](https://github.com/sst/opencode/blob/fee476bb90043a1012abda156dd9af9e5c71b19d/packages/core/src/global.ts#L22)). That stays container-local, so every session re-downloads them.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

## What this PR does

Makes OpenCode's XDG state survive container restarts. `entrypoint.d/setup.sh` generalizes the existing data-dir migration into a reusable `opencode_link_into_store` helper and applies it twice: `~/.local/share/opencode` -> `~/.config/opencode` (unchanged) and `~/.local/state/opencode` -> `~/.config/opencode/xdg-state` (new). Since the config dir is the bind-mounted config store, both persist. `spec.yaml` gains `statePaths` so the config-source overlay preserves that runtime state and host passthrough cannot overwrite it.

The follow-up commit addressed the earlier round well: the store name no longer collides with the data-dir namespace, `log/` is gone, the path list is annotated, and `docs/runtime/stores.md` is updated.

## Verification

I checked the `statePaths` list against OpenCode upstream (`packages/core/src/global.ts` and the call sites of `Global.Path.data`). It is accurate and complete: `storage/`, `plans/`, `snapshot/`, `worktree/`, `repos/`, `tool-output/`, `mcp-auth.json` and `opencode*.db*` are exactly the top-level entries OpenCode writes under its data root, and `auth.json` is already covered by the provider auth files. `log/` is the only omission and it is disposable.

## Where to focus

Two points inline. The state root is resolved through `xdg-basedir` upstream, so hardcoding `$HOME/.local/state` breaks silently if the image exports `XDG_STATE_HOME`. And pinning `xdg-state/` wholesale persists OpenCode's generated daemon `password` and a stale `server.json` registration in the per-project config store, where only `model.json` is actually needed for the stated goal.

- [document where the statePaths entries come from](https://github.com/eclipse-enclave/enclave/pull/96#discussion_r4053124130)
- [drop `log/` from the pinned state paths](https://github.com/eclipse-enclave/enclave/pull/96#discussion_r4053124159)
- [rename the store dir so it cannot collide with the data-dir namespace](https://github.com/eclipse-enclave/enclave/pull/96#discussion_r4053124248)
- [mention the new store directory in `docs/runtime/stores.md`](https://github.com/eclipse-enclave/enclave/pull/96#discussion_r4053124288)

tortmayr:
I executed the testplan and confirm that state persists now.
Could you please address  (or discard) the remaining AI review comments? Then I will finish my review

extensions/tools/opencode/spec.yaml
xai: The state root is symlinked into xdg-state/ by setup.sh, so pinning individual files would only change what an overlay clears, not what gets persisted. Upstream already treats password as a durable credential and health-checks server.json before reuse, so I'm keeping the whole directory rather than an allow-list that drifts with each OpenCode release.

EclipseSourceAI (commented):
> [!NOTE]
> **Autonomous AI review.**
>
>
>
>

Since my last review, the only change is `e9322f1` ("fix: honor OpenCode XDG state home"). It resolves the `XDG_STATE_HOME` hardcoding issue I flagged: `setup.sh` now resolves `state_base` from `${XDG_STATE_HOME:-$HOME/.local/state}` with a writability probe that falls back to the home-local root, and the devcontainer path-normalization call in `entrypoint.sh` was hoisted to run before extension setup scripts are sourced, so the rewritten `HOME`/`XDG_*` vars are visible to tool setup. New tests cover both the normalization ordering and the fallback path, and everything passes (`go test`, `shellcheck`, `gofmt`). I didn't find any new issues in this change.

The other still-open thread (pinning the whole `xdg-state/` directory, including the daemon `password` and `server.json`) was pushed back on with reasoning, so I'm not re-raising it. That's a maintainer call now.

- [hardcoded XDG_STATE_HOME path breaks silently under a custom XDG_STATE_HOME](https://github.com/eclipse-enclave/enclave/pull/96#discussion_r4053582331)

tortmayr (approved):
LGTM! 👍🏼

